PR #5, opened 2024-10-03
Standalone Mode (beta)

pmarkowsky:
This PR creates a new Santa operating mode -- *standalone*
    
This adds a new operating mode to Santa called standalone mode.
    
When running in standalone mode TouchID can be used to approved binaries. If a
binary is properly signed a SigningID rule is generated otherwise a SHA256 rule
is generated. Note this lacks a GUI for browsing rules in the local rule db.

Example video of behavior attached (has the wrong logo):

[Standalone Santa Operation](https://www.youtube.com/watch?v=ja5mylENUro)

This PR is marked draft until the following things are added

1. [x] Comments are cleaned up
2. [ ] Documentation is updated (Mostly done -- sync updated, mode updated)
3. [ ] Tests are updated
4. [x] We decide if this is the right place to put the auth (move to reply block)
5. [x] We decide if Events need to have a new flag on them that informs the sync service that a binary was approved via Standalone mode. (Added standaloneApproval flag to events.
6. [ ] Rebase changes for the Swift UI

---
Aside I've been running with this mode for a month+.

### DIFF
--- a/Source/common/SNTXPCUnprivilegedControlInterface.h
+++ b/Source/common/SNTXPCUnprivilegedControlInterface.h
@@ -51,6 +51,11 @@ struct RuleCounts {
 - (void)databaseEventCount:(void (^)(int64_t count))reply;
 - (void)staticRuleCount:(void (^)(int64_t count))reply;
 
+/// 
+/// Standalone rules
+///
+- (void)addStandaloneRule:(SNTRule *)rule reply:(void (^)(NSError *))reply;
+
 ///
 ///  Decision ops
 ///
--- a/Source/common/SNTXPCUnprivilegedControlInterface.h
+++ b/Source/common/SNTXPCUnprivilegedControlInterface.h
@@ -51,7 +51,7 @@ struct RuleCounts {
 - (void)databaseEventCount:(void (^)(int64_t count))reply;
 - (void)staticRuleCount:(void (^)(int64_t count))reply;
 
-/// 
+///
 /// Standalone rules
 ///
 - (void)addStandaloneRule:(SNTRule *)rule reply:(void (^)(NSError *))reply;
--- a/Source/common/SNTXPCUnprivilegedControlInterface.h
+++ b/Source/common/SNTXPCUnprivilegedControlInterface.h
@@ -52,11 +52,6 @@ struct RuleCounts {
 - (void)staticRuleCount:(void (^)(int64_t count))reply;
 
 ///
-/// Standalone rules
-///
-- (void)addStandaloneRule:(SNTRule *)rule reply:(void (^)(NSError *))reply;
-
-///
 ///  Decision ops
 ///
 
--- a/Source/gui/SNTBinaryMessageWindowController.h
+++ b/Source/gui/SNTBinaryMessageWindowController.h
@@ -26,7 +26,7 @@
 - (instancetype)initWithEvent:(SNTStoredEvent *)event
                     customMsg:(NSString *)message
                     customURL:(NSString *)url
-                    reply:(void (^)(BOOL authenticated))replyBlock;
+                        reply:(void (^)(BOOL authenticated))replyBlock;
 
 - (IBAction)showCertInfo:(id)sender;
 - (void)updateBlockNotification:(SNTStoredEvent *)event withBundleHash:(NSString *)bundleHash;
@@ -66,9 +66,8 @@
 @property(readonly) SNTStoredEvent *event;
 
 ///
-/// TODO(PLM) Fix comment
-/// The reply block to call when the user has authenticated via touch ID in standalone mode.
-///
+///  The reply block to call when the user has made a decision in standalone
+///  mode.
 @property(readonly, nonatomic) void (^replyBlock)(BOOL authenticated);
 
 ///
--- a/Source/gui/SNTBinaryMessageWindowController.h
+++ b/Source/gui/SNTBinaryMessageWindowController.h
@@ -60,12 +60,6 @@
 @property(weak) IBOutlet NSButton *dismissEventButton;
 
 ///
-///   Reference to the "Check Event" button in the XIB. Used to help the user
-///   check the binary.
-///
-@property(weak) IBOutlet NSButton *checkEventButton;
-
-///
 ///  The execution event that this window is for
 ///
 @property(readonly) SNTStoredEvent *event;
--- a/Source/gui/SNTBinaryMessageWindowController.h
+++ b/Source/gui/SNTBinaryMessageWindowController.h
@@ -66,6 +66,17 @@
 @property(readonly) SNTStoredEvent *event;
 
 ///
+/// TODO(PLM) Fix comment
+/// The reply block to call when the user has authenticated via touch ID in standalone mode.
+///
+@property(readonly, nonatomic) void (^replyBlock)(BOOL authenticated);
+
+///
+/// TODO(PLM) Fix this comment
+///
+//@property dispatch_semaphore_t replyBlockSemaphore;
+
+///
 ///  The root progress object. Child nodes are vended to santad to report on work being done.
 ///
 @property NSProgress *progress;
--- a/Source/gui/SNTBinaryMessageWindowController.h
+++ b/Source/gui/SNTBinaryMessageWindowController.h
@@ -1,4 +1,5 @@
 /// Copyright 2015 Google Inc. All rights reserved.
+/// Copyright 2024 North Pole Security, Inc.
 ///
 /// Licensed under the Apache License, Version 2.0 (the "License");
 /// you may not use this file except in compliance with the License.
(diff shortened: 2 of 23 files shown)


Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
                                                showGroup:YES];
 }
 
+// When running in standalone mode, the user is prompted to approve the binary.
+- (void)approveBinaryForStandaloneMode {
+  LAContext *context;
```
tburgin, line 155: nit: remove this and just declare and initialize on L159

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
...
+  NSError *err;
+
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
```
tburgin, line 158: move this to after the potential -canEvaluatePolicy early return

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
...
+  context = [[LAContext alloc] init];
+
+  if (![context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:&err]) {
```
tburgin, line 161: I would do this check even before changing the UI in -loadWindow. Or better yet, when standalone mode is enabled, but bio auth is not available, we display an error in the block popup letting the user know.

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
...
+                          addStandaloneRule:newRule
+                                      reply:^(NSError *error) {
+                                        if (error) {
+                                          LOGE(@"Failed to modify rules standalone: %s",
+                                               [error.localizedDescription UTF8String]);
+                                          LOGE(@"Failure reason: %@", error.localizedFailureReason);
+                                        }
+                                        dispatch_semaphore_signal(sema2);
+                                      }];
+                        dispatch_semaphore_wait(sema2, DISPATCH_TIME_FOREVER);
```
tburgin, line 213: We should add some timeouts.

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
...
+                      } else {
+                        LOGE(@"Authentication Error: %@", error);
+                      }
+                      dispatch_semaphore_signal(sema);
+                    }];
+  dispatch_semaphore_wait(sema, DISPATCH_TIME_FOREVER);
```
tburgin, line 219: Same here

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
...
+                          LOGE(@"Connection invalidated");
+                          dispatch_semaphore_signal(sema2);
+                        };
+
+                        [daemonConn resume];
+                        [[daemonConn remoteObjectProxy]
+                          addStandaloneRule:newRule
+                                      reply:^(NSError *error) {
+                                        if (error) {
+                                          LOGE(@"Failed to modify rules standalone: %s",
```
tburgin, line 207: Need some user feedback in the UI for this case.

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -135,13 +150,96 @@ - (IBAction)showCertInfo:(id)sender {
...
+
   NSURL *url = [SNTBlockMessage eventDetailURLForEvent:self.event customURL:self.customURL];
 
   [self closeWindow:sender];
   [[NSWorkspace sharedWorkspace] openURL:url];
 }
 
+- (IBAction)checkEventDetails:(id)sender {
+  NSString *eventCheckURL =
+    [NSString stringWithFormat:@"https://www.virustotal.com/gui/search/%@", self.event.fileSHA256];
```
tburgin, line 237: I like this idea. Though we should make it configurable. Similar to what we do for event details.

I don't think we want to hard code virus total here.

Source/gui/SNTBinaryMessageWindowController.m
pmarkowsky: Good call, should this be configurable?

Source/gui/SNTBinaryMessageWindowController.h
```diff
@@ -59,6 +59,12 @@
 ///
 @property(weak) IBOutlet NSButton *dismissEventButton;
 
+///
+///   Reference to the "Check Event" button in the XIB. Used to help the user
+///   check the binary.
+///
+@property(weak) IBOutlet NSButton *checkEventButton;
+
```
pmarkowsky, line 67: This was an accidental squash error, will replace this and move it to another PR.

Source/gui/SNTBinaryMessageWindowController.m
pmarkowsky: > Or better yet, when standalone mode is enabled, but bio auth is not available, we display an error in the block popup letting the user know

This seems like a good idea. Should we also disable the `Approve` button so users don't get frustrated?

Source/common/SNTXPCUnprivilegedControlInterface.h
```diff
@@ -51,6 +51,11 @@ struct RuleCounts {
 - (void)databaseEventCount:(void (^)(int64_t count))reply;
 - (void)staticRuleCount:(void (^)(int64_t count))reply;
 
+///
+/// Standalone rules
+///
+- (void)addStandaloneRule:(SNTRule *)rule reply:(void (^)(NSError *))reply;
```
tburgin, line 57: I don’t think we want to expose adding arbitrary rules over the unprivileged interface. Instead, we could have the daemon drive the flow, sending a message to the GUI for auth.

pmarkowsky: Admittedly I went back and forth on this. What were you thinking the flow would look like?

The main reason I added this was to keep the flow simpler then the santad flows I sketched out. Felt like I had to keep state in the daemon which felt wrong. 

If you enable standalone mode then our flow looks like this.

```mermaid
sequenceDiagram
santad ->> gui: We blocked a binary pop up message and display approve button
gui ->> gui: User clicks approve
gui ->> gui: Get auth from user
gui -->> santad: addStandaloneRule
```

If the user clicks ignore then nothing happens.

Source/gui/SNTBinaryMessageWindowController.m
tburgin: Yeah, if the desired auth is not available we shouldn't offer standalone buttons.

Source/gui/SNTBinaryMessageWindowController.m
tburgin: Most other XPC calls we use a 5 sec timeout. That should work fine here too.

Source/common/SNTXPCUnprivilegedControlInterface.h
tburgin: > The main reason I added this was to keep the flow simpler then the santad flows I sketched out. Felt like I had to keep state in the daemon which felt wrong.

Having the GUI push in rules is simpler, but I don't think we want to expose adding rules from an unprivileged interface. With that as a requirement, we will need to store state with the daemon.

sequenceDiagram
santad ->> santad: blocks an executable - checks if standalone mode and desired auth is available
santad ->> rules.db: adds a pending rule
santad ->> gui: prompt user and display approve button
gui ->> gui: User clicks approve
gui ->> gui: Get auth from user
gui ->> santad: activateStandaloneRule:
santad ->> rules.db: converts pending rule to active

The idea of a "pending rule" is new and we would need to think though the implications.

The other thing we need to consider is the key used with activateStandaloneRule. The obvious choice is to just use the blockable event. This is probably fine. However, it does not strongly guarantee that a user approved the event. We could consider using a hard to guess nonce, or figure out if LocalAuthentication can mint an attestation that we verify in the activateStandaloneRule: implementation.

tburgin: Chatted offline - using a reply block would eliminate the need for exposing an additional interface in the daemon. This is a great idea!

Source/gui/SNTBinaryMessageWindowController.m
pmarkowsky: I removed the semaphore since it's unnecessary now.

Source/gui/SNTBinaryMessageWindowController.m
pmarkowsky: I made the button disappear. But haven't figured out how to update the UI to display to the user that things are not operating as intended.

pmarkowsky:
Going to close this PR and start a new one based on the new UI code.

PR #6, opened 2024-10-03
Hook up CS INVALIDATED event

mlw:
The CS INVALIDATED event had previously been implemented here: https://github.com/google/santa/pull/1210

However it was never subscribed to. This PR adds the subscriptions and hooks up the appropriate metrics.


pmarkowsky (approved):
LGTM.

PR #17, opened 2024-10-04
Fix tree aware clients handleContextMessage default return

mlw:
This fixes an issue for tree aware clients that have not enabled any process tree annotations but also do no subscribe to the set of force-enabled tree aware client event types (fork/exec/exit).

Before this change, `handleContextMessage` would always return false when process trees weren't in use. Instead it should return whether or not the event type was force added.

This bug does not currently affect Santa's current implementation as the only tree aware client (the recorder) subscribes to the full set of force-enabled events. But it does affect various debug routines where event subscriptions are manipulated.

This PR also begins including tree aware client tests.


pmarkowsky (approved):
LGTM.

PR #18, opened 2024-10-05
[DRAFT] Add codesigning information to the ProcessInfoLight message

mlw:
This adds the `CodeSignature` message ([proto](https://github.com/northpolesec/santa/blob/e43219f07437ca558e6defc2aff37d3f8bcf40e1/Source/common/santa.proto#L32)) to the `ProcessInfoLight` message ([proto](https://github.com/northpolesec/santa/blob/e43219f07437ca558e6defc2aff37d3f8bcf40e1/Source/common/santa.proto#L182)).

Currently there are no good identifiers (except for path) when looking at a telemetry log to understand what the process might be. The `CodeSignature` message exists in the larger `ProcessInfo` message, but that is only used by EXEC and FAA events.

Pros: Adding the SID and TID will make processes more recognizable. The cdhash will also help make queries easier.
Cons: It does add overhead small overhead to event sizes. Also, while using cdhash to look for execs would be easier to get more information about a given process (e.g. the hash), you'd still need to run sub queries looking for pid/pidver and parent pid/pidver to find the exact EXEC for a process (e.g. to see args).

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -1,4 +1,17 @@
-// Important: This schema is currently in BETA
+/// Copyright 2024 Google LLC
+/// Copyright 2024 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     https://www.apache.org/licenses/LICENSE-2.0
+///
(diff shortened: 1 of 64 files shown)


pmarkowsky:
@mlw any reason not to put this in 2025.1?

pmarkowsky:
This is pretty stale. Marking this as closed. We can reopen if we want to bring this back.

PR #19, opened 2024-10-07
installer: support side-by-side install of NPS Santa [1/2]

tburgin:
This CL changes the installer to lay down Santa on disk as Santa_NPS.app. The post install or daemon will rename Santa_NPS.app to Santa.app when it is safe to do so.

A follow up CL will add logic in the daemon to idle while another install of Santa is running.

To test the installer behavior as new label has been added `package-dev`, which will package up a dev build. To support this, `package.sh` is now a separate script that can be called for both release and dev builds.

```
bazel build :package-dev -c opt
```

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -186,14 +186,13 @@ genrule(
 )
 
 genrule(
-  name = "package-dev",
-  srcs = [ ":release" ],
-  outs = ["santa-dev.pkg"],
-  cmd = """
+    name = "package-dev",
+    srcs = [":release"],
+    outs = ["santa-dev.pkg"],
+    cmd = """
   tar -xzvf $(<)
-  RELEASE_ROOT=. SCRATCH=. BUILD_DEV_DISTRIBUTION_PKG=1 \
+  RELEASE_ROOT=. PKG_OUT_DIR=$(@D) BUILD_DEV_DISTRIBUTION_PKG=1 \
     ./conf/package.sh
-  mv santa-dev.pkg $(@)
   """,
 )
 
--- a/BUILD
+++ b/BUILD
@@ -108,6 +108,7 @@ genrule(
         "Conf/com.northpolesec.santa.newsyslog.conf",
         "Conf/Package/Distribution.xml",
         "Conf/Package/notarization_tool.sh",
+        "Conf/Package/package.sh",
         "Conf/Package/package_and_sign.sh",
         "Conf/Package/postinstall",
         "Conf/Package/preinstall",
@@ -182,6 +183,19 @@ genrule(
     """,
     }),
     heuristic_label_expansion = 0,
+)
+
+genrule(
+  name = "package-dev",
+  srcs = [ ":release" ],
+  outs = ["santa-dev.pkg"],
+  cmd = """
+  tar -xzvf $(<)
+  # chmod a+x conf/package.sh
+  RELEASE_ROOT=. SCRATCH=. BUILD_DEV_DISTRIBUTION_PKG=1 \
+    ./conf/package.sh
+  mv santa-dev.pkg $(@)
+  """,
 )
 
 test_suite(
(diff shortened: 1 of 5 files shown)


Conf/Package/package.sh
```diff
@@ -0,0 +1,76 @@
...
+# The dev package is helpful when testing installer behavior for dev builds.
+#
+# All of the following environment variables are required.
+
+# RELEASE_ROOT is a required environment variable that points to the root
+# of an extracted release tarball produced with the :release and :release_driver
+# rules in Santa's main BUILD file.
+[[ -n "${RELEASE_ROOT}" ]] || die "RELEASE_ROOT unset"
+
+[[ -n "${SCRATCH}" ]] || die "SCRATCH unset"
```
mlw, line 16: nit: would be nice to comment this as required above with `RELEASE_ROOT`

Conf/Package/postinstall
```diff
@@ -13,8 +13,16 @@
 mkdir -p /usr/local/bin
 /bin/ln -sf /Applications/Santa.app/Contents/MacOS/santactl /usr/local/bin/santactl
 
-# Load com.northpolesec.santa.daemon
-/Applications/Santa.app/Contents/MacOS/Santa --load-system-extension
+if /bin/launchctl list EQHXZ8M8AV.com.google.santa.daemon > /dev/null 2>&1; then
```
mlw, line 16: Are we guaranteed to be running as root here?

BUILD
```diff
@@ -184,6 +185,18 @@ genrule(
...
 )
 
+genrule(
+  name = "package-dev",
+  srcs = [ ":release" ],
+  outs = ["santa-dev.pkg"],
+  cmd = """
+  tar -xzvf $(<)
+  RELEASE_ROOT=. SCRATCH=. BUILD_DEV_DISTRIBUTION_PKG=1 \
+    ./conf/package.sh
```
russellhancox, line 195: Conf

It'll only matter on a case sensitive system but just in case

Conf/Package/package.sh
```diff
@@ -0,0 +1,76 @@
...
+# Disable bundle relocation.
+/usr/bin/pkgbuild --analyze --root "${APP_PKG_ROOT}" "${SCRATCH}/component.plist"
+/usr/bin/plutil -replace BundleIsRelocatable -bool NO "${SCRATCH}/component.plist"
+/usr/bin/plutil -replace BundleIsVersionChecked -bool NO "${SCRATCH}/component.plist"
+/usr/bin/plutil -replace BundleOverwriteAction -string upgrade "${SCRATCH}/component.plist"
+/usr/bin/plutil -replace ChildBundles -json "[]" "${SCRATCH}/component.plist"
+
+# Build app package
+/usr/bin/pkgbuild --identifier "com.northpolesec.santa" \
+  --version 9999.1.1 \
```
russellhancox, line 63: Can this be pulled from the app info plist?

Conf/Package/postinstall
russellhancox: Yes

Conf/Package/package.sh
tburgin: Good call. When writing the comment, I realized `SCRATCH` really isn't required. Instead, what we were using it for should have its own variable. `PKG_OUT_DIR` - added a comment.

BUILD
tburgin: :) actually `conf` is correct. The script is being launched from the release tar.gz, which uses lowercase for its directory names.

Conf/Package/package.sh
tburgin: Ah, thanks for the reminder. I punted this because `defaults read` was failing and I forgot to look back into it.

It turns out `defaults` can not read from relative paths, which `:package-dev` makes use of. I switch to using `plutil` throughout.

tburgin (commented):
PTAL

mlw (approved):

PR #20, opened 2024-10-07
installer: support side-by-side install of NPS Santa [2/3]

tburgin:
This CL adds logic to the daemon that waits for previous versions of Santa to be removed.

TODO: Check /Applications/Santa_NPS.app signature before finalizing the install.
TODO: Handle supporting services.
TODO: Add tamper protection to /Applications/Santa.app.


Source/santad/main.mm
```diff
@@ -85,6 +87,52 @@ static void SantaWatchdog(void *context) {
   }
 }
 
+static void FinishInstall() {
+  // Wait for com.google.santa.daemon to be removed.
+  while (1) {
+    // Record if com.google.santa.daemon is an active and enabled system extension.
+    BOOL google_santa_active_enabled = NO;
+    NSDictionary *system_extensions =
+      [NSDictionary dictionaryWithContentsOfFile:@"/Library/SystemExtensions/db.plist"];
```
mlw, line 96: I think I have concerns about stability here. Would something like `propertiesRequestForExtension:queue:` work here? https://developer.apple.com/documentation/systemextensions/ossystemextensionrequest/propertiesrequest(forextensionwithidentifier:queue:)?language=objc

pmarkowsky: Looks like OSQuery's also using it as well https://github.com/osquery/osquery/issues/6738

https://github.com/osquery/osquery/blob/8920f3343613b8a7a9521d740af5e6e4650ee848/osquery/tables/system/darwin/system_extensions.mm#L29

mlw: We ran some experiments here. Unfortunately that API only returns properties for extensions with the same team ID

Source/santad/main.mm
```diff
@@ -87,27 +87,51 @@ static void SantaWatchdog(void *context) {
...
+  NSArray *system_extensions = system_extensions_db[@"extensions"];
+  if (![system_extensions isKindOfClass:[NSArray class]]) return false;
+
+  for (NSDictionary *sysx in system_extensions) {
+    if (![sysx isKindOfClass:[NSDictionary class]]) return false;
+
+    NSString *identifier = sysx[@"identifier"];
+    if (![identifier isKindOfClass:[NSString class]]) return false;
+
+    NSString *state = sysx[@"identifier"];
```
mlw, line 108: `sysx[@"state"];`

mlw (changes requested):

Source/santad/MachServiceDeadWaiter.h
```diff
@@ -0,0 +1,37 @@
...
+#include <string>
+
+// MachServiceDeadWaiter is a small waiter class that looks up a mach service,
+// then waits for all the receive rights of that service to reach zero
+// (MACH_NOTIFY_DEAD_NAME).
+class MachServiceDeadWaiter {
+ public:
+  explicit MachServiceDeadWaiter(std::string service_name);
+  MachServiceDeadWaiter(const MachServiceDeadWaiter &) = delete;
+  MachServiceDeadWaiter &operator=(const MachServiceDeadWaiter &) = delete;
```
mlw, line 29: nit: Should also delete the move constructor and move assignment operator

tburgin: Since I have a user declared copy constructor, even though it's just a delete, the implicit move constructor will not be created. Same goes for the copy assignment operator and move assignment operator.

https://en.cppreference.com/w/cpp/language/move_constructor

mlw: TIL, nice

mlw (approved):

PR #22, opened 2024-10-09
Add mirroring workflow for code signing repository

pmarkowsky:
This PR adds a mirroring workflow that runs every 10 minutes. This is useful for our codesigning repository.

The mirror workflow is gated by a GitHub secret `ENABLE_MIRRORING` which must be equal to `"true"` for the workflow to start.


pmarkowsky:
Closing this in favor a having the release process slurp up the the latest on demand rather than keeping a mirror.

PR #23, opened 2024-10-09
installer: support side-by-side install of NPS Santa [3/3] 

tburgin:
This CL adds side-by-side support for the various helper services.

The installer no longer lays down the launch job configs, rather they are installed and loaded by the sysx when appropriate.


Conf/finish_install.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Load com.northpolesec.santa.bundleservice
+/bin/launchctl load -w /Library/LaunchDaemons/com.northpolesec.santa.bundleservice.plist
+
+# Load com.northpolesec.santa.metricservice
+/bin/launchctl load -w /Library/LaunchDaemons/com.northpolesec.santa.metricservice.plist
+
+# Load com.northpolesec.santa.syncservice
+/bin/launchctl load -w /Library/LaunchDaemons/com.northpolesec.santa.syncservice.plist
+
```
mlw, line 34: Should we also here disable and/or remove the Google Santa launchd plists?

tburgin:
It is getting increasingly complex to handle the edge cases or the sysx finishing the install. Closing this. We will figure out a simpler approach.

PR #30, opened 2024-10-12
installer: generate migration pkg [2/2]

tburgin:


Conf/MigrationPackage/migration.sh
```diff
@@ -1,4 +1,37 @@
...
-# TODO: Add migration logic
-while true; do logger "migration"; sleep 1; done;
+while true;
+do
+  GOOGLE_SANTA_ACTIVATED="$(/usr/bin/systemextensionsctl list com.apple.system_extension.endpoint_security |
+                            /usr/bin/grep -E '^.+EQHXZ8M8AV.+com\.google\.santa\.daemon.+activated.+$')"
+  if [ -z "${GOOGLE_SANTA_ACTIVATED}" ]; then
+    break;
+  else
+    sleep 1
```
russellhancox, line 10: Probably doesn't need to be this tight a loop, 10s or even 30s would be plenty quick enough

tburgin: good call, upped it to 10s

russellhancox (approved):

PR #40, opened 2024-10-16
Add new install command

mlw:


Source/santactl/Commands/SNTCommandInstall.mm
```diff
@@ -0,0 +1,87 @@
...
+                                                 }];
+
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 10 * NSEC_PER_SEC)) > 0) {
+    LOGW(@"Timed out waiting for install to complete.");
+  }
+
+  exit(EXIT_SUCCESS);
+}
+
+@end
```
tburgin, line 87: nit: add a new line at the end

tburgin (dismissed):

Source/santactl/Commands/SNTCommandInstall.mm
```diff
@@ -36,7 +36,7 @@ + (BOOL)requiresDaemonConn {
 }
 
 + (NSString *)shortHelpText {
-  return @"Install a given package";
+  return @"Instruct the daemon to install Santa.app."
```
tburgin, line 39: oops, missing `;`

tburgin (dismissed):

Source/santactl/Commands/SNTCommandInstall.mm
```diff
@@ -0,0 +1,87 @@
...
+  LOGI(@"Asking daemon to install: %@", path);
+
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+  [[self.daemonConn remoteObjectProxy] installSantaApp:(NSString *)path
+                                                 reply:^(BOOL success) {
+                                                   LOGI(@"Got reply from daemon: %d", success);
+                                                   dispatch_semaphore_signal(sema);
+                                                 }];
+
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, 10 * NSEC_PER_SEC)) > 0) {
```
russellhancox, line 62: Is 10s enough?

Source/santactl/Commands/SNTCommandInstall.mm
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  if (!path) {
+    [self printErrorUsageAndExit:@"No path specified"];
+  }
+
+  LOGI(@"Asking daemon to install: %@", path);
+
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+  [[self.daemonConn remoteObjectProxy] installSantaApp:(NSString *)path
```
russellhancox: You could use synchronousRemoteObjectProxy here instead of using a semaphore. The only downside is you lose the ability to specify how long to wait.

tburgin (approved):

Source/santactl/Commands/SNTCommandInstall.mm
mlw: I feel like it should be, but idk. We can tweak this on the next PR once the install code is in place

Source/santactl/Commands/SNTCommandInstall.mm
mlw: I think the timeout makes sense here to provide some sort of expectation of bounds to the caller.

russellhancox (approved):

PR #41, opened 2024-10-16
package: tamper resistant install [1/2]

tburgin:
This CL adds logic to the .pkg to support installation and updates to a tamper resistant Santa.app.

- The .pkg no longer installs to /Applications/Santa.app. When appropriate, the sysx will move the updated Santa.app into place.
- The .pkg no longer installs the launch agent and daemon configurations. When appropriate, the sysx will handle this task. This is coming in a follow up CL.
- Migrating from an unsupported version of Santa is now handled by the main .pkg, the wrapper migration .pkg has been removed.


pmarkowsky (approved):
LGTM.

PR #44, opened 2024-10-17
Implement daemon install command

mlw:
This PR implements the daemon side logic for the Santa install flow.

### DIFF
--- a/Source/santad/SNTDaemonControlController.mm
+++ b/Source/santad/SNTDaemonControlController.mm
@@ -15,6 +15,7 @@
 
 #import "Source/santad/SNTDaemonControlController.h"
 
+#include <Foundation/Foundation.h>
 #import <MOLCodesignChecker/MOLCodesignChecker.h>
 #import <MOLXPCConnection/MOLXPCConnection.h>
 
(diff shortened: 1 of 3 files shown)


Source/santad/SNTDaemonControlController.mm
```diff
@@ -387,8 +389,98 @@ - (void)syncBundleEvent:(SNTStoredEvent *)event relatedEvents:(NSArray<SNTStored
...
+
+  NSError *err;
+  MOLCodesignChecker *cc = [[MOLCodesignChecker alloc] initWithBinaryPath:path error:&err];
+
+  if (err) {
+    LOGE(@"Failed to validate install path: %@", err);
+    return NO;
+  }
+
+  if (![cc.teamID isEqualToString:@"ZMCG7MLDV9"] ||
```
russellhancox, line 412: Instead of hardcoding this, it might be good to get this from the running binary. MOLCodesignChecker has an initWithSelf method you can use for this

tburgin:
Tested locally with two version built with `bazel build :package-dev -c opt`. Works great!

Source/santad/SNTDaemonControlController.mm
mlw: Spoke offline. While getting the TID dynamically is easy, pairing with the appropriate SID is hard and would require changes to the MOLXPCConnection repo to expose information about the GUI endpoint. Since this would only affect users attempting to rebuild Santa as their own, renamed product, and because that would require several other code changes as well, punting on this effort for now.

tburgin (approved):

PR #45, opened 2024-10-17
Set associated BundleIDs in launchd plists

mlw:
This change allows the icon for Santa to be properly displayed in the "Login Items & Extensions" settings page for "Allow in Background" items.


tburgin:
Awesome!

PR #46, opened 2024-10-17
GUI: Re-write binary block window with SwiftUI [1/3]

russellhancox:
This doesn't include any bundle hashing UI, including progress while hashing and presentation of the bundle hash once complete. That will be added in a future commit.

The primary changes to the UI:

* The application icon is now shown next to the title
* A lot of information has been moved into a modal accessible from the "More Details" link
* All the text fields are now selectable (and therefore copyable)
* All of the information can be copied to the clipboard with a single button

Main window:
image

More Details:
image

Text from the "Copy Details" button:
image

Part of #14


Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -79,88 +73,44 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
...
-        setKeyEquivalentModifierMask:NSEventModifierFlagCommand];  // Command Key
-    }
-  }
-
-  NSString *dismissButtonText = [[SNTConfigurator configurator] dismissText];
-  if (dismissButtonText.length) {
-    [self.dismissEventButton setTitle:dismissButtonText];
-  }
-
-  if (!self.event.needsBundleHash) {
```
tburgin, line 107: Mind adding this if/else block as a comment in SNTBinaryMessageWindowView.swift? Just so we can remember what is should be doing when we go to implement.

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -0,0 +1,220 @@
...
+struct SNTBinaryMessageWindowView: View {
+  let window: NSWindow?
+  let event: SNTStoredEvent?
+  let customMsg: NSString?
+  let customURL: NSString?
+  let uiStateCallback: ((Bool) -> Void)?
+
+  let c = SNTConfigurator()
+
+  @State public var checked = false
```
tburgin, line 160: name this something more descriptive

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -0,0 +1,220 @@
...
+import santa_common_SNTStoredEvent
+
+@objc public class SNTBinaryMessageWindowViewFactory : NSObject {
+  @objc public static func createWith(window: NSWindow,
+                                      event: SNTStoredEvent,
+                                      customMsg: NSString?,
+                                      customURL: NSString?,
+                                      uiStateCallback: ((Bool) -> Void)?) -> NSViewController {
+    return NSHostingController(rootView:SNTBinaryMessageWindowView(
+      window:window, event:event, customMsg:customMsg, customURL:customURL, uiStateCallback:uiStateCallback).frame(minWidth:520, minHeight:340).fixedSize())
```
tburgin, line 29: make `minWidth` and `minHight` constants since they are used elsewhere

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -0,0 +1,220 @@
...
+
+  @Environment(\.presentationMode) var presentationMode
+
+  func addLabel(@ViewBuilder closure: () -> some View) -> some View {
+    HStack(spacing:5.0) {
+      VStack(alignment:.leading, spacing:2.0) {
+        closure()
+      }.frame(alignment:.leading)
+      Spacer()
+    }.frame(width:400).fixedSize()
```
tburgin, line 44: same for width, make it a constant

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -0,0 +1,220 @@
...
+
+      Toggle(isOn: $checked) {
+        Text("Prevent future notifications for this application for a day").font(Font.system(size: 11.0));
+      }
+
+      HStack(spacing:15) {
+        if c.eventDetailURL != nil {
+          Button(action: openButton, label: {
+            if let edt = c.eventDetailText {
+              Text(edt).frame(maxWidth:150.0)
```
tburgin, line 180: same with maxWidth

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -174,7 +124,7 @@ - (void)updateBlockNotification:(SNTStoredEvent *)event withBundleHash:(NSString
       self.event.fileBundleHash = bundleHash;
       [self.foundFileCountLabel removeFromSuperview];
       [self.hashingIndicator setHidden:YES];
-      [self.openEventButton setEnabled:YES];
+      // [self.openEventButton setEnabled:YES];
```
mlw, line 127: Is this meant as another reminder for future work or should this be removed?

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -0,0 +1,313 @@
...
+      .keyboardShortcut("m", modifiers: .command)
+
+      Button(action: copyDetailsToClipboard) {
+        HStack(spacing:2.0) {
+          Text("\(Text("C").underline())opy Details").foregroundColor(.blue)
+          Image(systemName:"pencil.and.list.clipboard").foregroundColor(.blue)
+        }
+      }
+      .buttonStyle(ScalingButtonStyle())
+      .keyboardShortcut("c", modifiers: .command)
```
mlw, line 206: Does Cmd+C shortcut still work when the detail sheet is displayed? And can you highlight text from the sheet? If so, I'm wondering what happens, does just the highlighted text get copied or does the whole sheet content get copied?

Source/gui/SNTAccessibleTextField.h
```diff
@@ -1,23 +0,0 @@
...
-///    limitations under the License.
-
-#import <Cocoa/Cocoa.h>
-
-/**
-   An NSTextField subclass that provides an accessiblity label equal to:
-   (self.toolTip + self.stringValue) where available. It also sets the
-   accessibilityRoleDescription to "label".
-*/
-@interface SNTAccessibleTextField : NSTextField
```
tburgin, line 22: Hmm, the current uses `SNTAccessibleTextField`. The .xib uses it as a custom class for each text field, along with adding a tooltip.

In what order are the fields traversed when accessibility is enabled with the new Swift UI? I assume they follow the V stack. If that is the case, we should add tooltips and accessibility labels for the values in the main popup. The details modal should be okay with out additional labels, since the field name will come right before the value in the accessibility tree.

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -0,0 +1,313 @@
...
+        Divider()
+
+        if let signingID = e?.signingID {
+          addLabel {
+            Text("Signing ID").bold().font(Font.system(size:12.0))
+            Text(signingID).font(Font.system(size:12.0).monospaced()).textSelection(.enabled)
+          }
+        }
+
+        Divider()
```
tburgin, line 76: If there is no SID, two dividers will be added. I don't think that is what we want?

Maybe move `Divider()` into the if block?

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -79,88 +73,44 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
...
+}
 
-  if (!self.event.fileBundleName) {
-    [self.applicationNameLabel removeFromSuperview];
-  }
+- (void)showWindow:(id)sender {
+  if (self.window) [self.window orderOut:sender];
+
+  self.window =
+    [[NSWindow alloc] initWithContentRect:NSMakeRect(0, 0, 0, 0)
```
tburgin, line 84: I think we want this to be an `SNTMessageWindow` to match existing behavior (see .xib) of an NSPanel. It also gets us the fade in and out animations.

russellhancox: SNTMessageWindow is gone; the fade in/out haven't been working for a while and are superfluous and the flags passed to create this window re-create a window without a title bar or traffic light buttons.

Source/gui/SNTAccessibleTextField.h
russellhancox: It reads them in the correct order; from the screenshot:

AppIcon, Image -> Santa, text -> This application has been banned, text -> Application, text -> Spotify, text -> Publisher, text -> Spotify - Developer ID Application: Spotify (2FNC3A47ZF), text -> User, text -> rah, text -> More details, button -> Copy details, button, Prevent future notifications for this application for a day, checkbox -> Open sync server, button -> Dismiss, button

Source/gui/SNTBinaryMessageWindowView.swift
russellhancox: Done, and the same for CDHash

Source/gui/SNTBinaryMessageWindowView.swift
russellhancox: > Does Cmd+C shortcut still work when the detail sheet is displayed?
No

> And can you highlight text from the sheet? 
Yes

> If so, I'm wondering what happens, does just the highlighted text get copied or does the whole sheet content get copied?
You can copy the highlighted text.

Now that you've pointed it out, overloading Cmd+C like this feels wrong, so I've updated to use Cmd+D instead. The answers above still apply.

Source/gui/SNTBinaryMessageWindowController.m
russellhancox: It's a reminder that the button should be disabled while hashing is in progress and re-enabled afterward

Source/gui/SNTBinaryMessageWindowController.m
tburgin: Ah, I see. Even though the .xib refers to it, we were not compiling in `SNTMessageWindow` as of
https://github.com/google/santa/pull/1010.

Source/gui/SNTAccessibleTextField.h
tburgin: Awesome!

tburgin:
code LGTM

tburgin (approved):

PR #47, opened 2024-10-17
build: fix bazel run :reload

tburgin:


Conf/install.sh
```diff
@@ -18,56 +18,26 @@ if [[ -z "${BINARIES}" || -z "${CONF}" ]]; then
...
+  # Removal was successful.
+  # Install Santa and load the system extension. The system extension will
+  # finish loading the rest of Santa's configs and helper services.
+  /bin/cp -r ${BINARIES}/Santa.app /Applications/
+  /Applications/Santa.app/Contents/MacOS/Santa --load-system-extension
+else
+  # Tamper protections are enabled, ask Santa to install the update. If the
+  # update is valid, the system extension will take care of finishing the
+  # install.
+  /bin/mkdir /Library/Caches/com.northpolesec.santa
```
russellhancox, line 36: `-p`, just in case

Conf/install.sh
```diff
@@ -18,56 +18,26 @@ if [[ -z "${BINARIES}" || -z "${CONF}" ]]; then
...
+  /bin/cp -r ${BINARIES}/Santa.app /Library/Caches/com.northpolesec.santa/
+  /Applications/Santa.app/Contents/MacOS/santactl install
+  # Cleanup cache dir.
+  /bin/rm -rf /Library/Caches/com.northpolesec.santa
```
russellhancox, line 40: Is this necessary? If it is, should santad do it?

tburgin: Currently only the installer and the migration.sh clean it up.

In santad, `-installSantaApp:reply:` isn't a great spot to do the removal, because it takes an arbitrary path for future flexibility (the path in `santactl install` is hardcoded to `/Library/Caches/com.northpolesec.santa/Santa.app`).

We could do a cleanup from [install_services.sh](https://github.com/northpolesec/santa/blob/15d6de1262329046bb48fd890cbb3d64bb8e57b5/Conf/install_services.sh), but that also doesn't feel quite right.

I think leaving the cleanup to the install scripts is okay for now.

russellhancox (approved):

PR #51, opened 2024-10-21
Workflows: Use the runner OS as the key for disk caches

russellhancox:
Also remove the workflow_dispatch option, being able to manually trigger the CI workflow isn't useful, continuous can be used for that instead.


pmarkowsky (approved):
LGTM.

PR #52, opened 2024-10-21
Project: Use `matrix.os` as cache key instead of `runner.os`

russellhancox:
runner.os is always `macOS`, so it doesn't do what we need


pmarkowsky (approved):
LGTM.

PR #54, opened 2024-10-22
GUI: Allow customization of silence period

russellhancox:
Allow selecting a silence period of 1 day, 1 week or 1 month. Changing the period will automatically check the box. When the box is checked the "Dismiss" text will change to "Dismiss & Silence" (unless the Dismiss text has been overridden in the config)

image
image


Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -238,12 +238,27 @@ struct SNTBinaryMessageWindowView: View {
...
   let customMsg: NSString?
   let customURL: NSString?
-  let uiStateCallback: ((Bool) -> Void)?
+  let uiStateCallback: ((TimeInterval) -> Void)?
 
   let c = SNTConfigurator()
 
   @State public var preventFutureNotifications = false
 
+  let preventNotificationPeriods: [TimeInterval] = [86400, 604800, 2592000]
```
mlw, line 247: Consider changing: `s/2592000/2678400/` to get to "one month"

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -238,12 +238,27 @@ struct SNTBinaryMessageWindowView: View {
...
+
+  init(window: NSWindow, event: SNTStoredEvent?, customMsg: NSString?, customURL: NSString?, uiStateCallback: ((TimeInterval) -> Void)?) {
+    self.window = window
+    self.event = event
+    self.customMsg = customMsg
+    self.customURL = customURL
+    self.uiStateCallback = uiStateCallback
+
+    dateFormatter.unitsStyle = .spellOut
+    dateFormatter.allowedUnits = [.hour, .day, .month]
```
mlw, line 259: consider replacing `.hour` with `.weekOfMonth`. With the time interval change above, my local tests now show "one day, one week, one month". I think it looks a bit nicer but debatable and not a blocker if you prefer the current days-only output.

russellhancox: Interesting, I didn't realize `.weekOfMonth` would do that. That's much nicer.

mlw (approved):

PR #55, opened 2024-10-22
tamper client: protect service configs

tburgin:
This CL protects Santa's service configs.


tburgin:
Preventing unsupported versions of Santa from loading services will be handled in a separate CL.

mlw (approved):

PR #67, opened 2024-10-24
GUI: Refactor common UI elements into a single module.

russellhancox:
At the same time, update the FAA dialog to match the new style

FAA main view:
image

FAA "More Details":
image

### DIFF
--- a/Source/common/SNTBlockMessage.h
+++ b/Source/common/SNTBlockMessage.h
@@ -1,4 +1,5 @@
 /// Copyright 2016 Google Inc. All rights reserved.
+/// Copyright 2014 North Pole Security, Inc.
 ///
 /// Licensed under the Apache License, Version 2.0 (the "License");
 /// you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


pmarkowsky:
Main concern here is that I'd want the process and username that performed the file access rather than the rule meta data in the main part of the FAA window.

Source/common/SNTBlockMessage.h
```diff
@@ -1,4 +1,5 @@
 /// Copyright 2016 Google Inc. All rights reserved.
+/// Copyright 2014 North Pole Security, Inc.
```
pmarkowsky, line 2:
```suggestion
/// Copyright 2024 North Pole Security, Inc.
```

Source/common/SNTFileAccessEvent.h
```diff
@@ -37,17 +38,4 @@
...
-///
-@property(readonly) NSString *publisherInfo;
-
-///
-/// Return an array of the underlying SecCertificateRef's of the signingChain
-///
-/// WARNING: If the refs need to be used for a long time be careful to properly
-/// CFRetain/CFRelease the returned items.
-///
-@property(readonly) NSArray *signingChainCertRefs;
```
pmarkowsky, line 51: Don't we need this for the process that caused the file access policy violation?

russellhancox: I've moved it into the superclass

russellhancox:
> Main concern here is that I'd want the process and username that performed the file access rather than the rule meta data in the main part of the FAA window.

I've added user (which was never previously a thing in this UI). 

Process is ambiguous: the PID is never shown (nor is it for binary blocks) but the process name is, and the parent is shown in the more details window.

tburgin (approved):

PR #76, opened 2024-10-25
Localization: Add script to regen and validate localizations

russellhancox:
localization.py will automatically re-generate the English localization file and then print out the status of the other localizations inside the source tree. This script is now also used in the CI workflow to return an error if the English localization is out-of-date.


tburgin:
This is great!

tburgin (dismissed):

tburgin (approved):

mlw (approved):

PR #77, opened 2024-10-25
Project: Disable pyink checks

russellhancox:
pyink isn't installed by default and isn't available in homebrew, making using it a bit annoying.

I've checked in a .pylintrc so we can manually lint if we want to (`pylint .` is enough), given that pylint seems to be installed by default when Homebrew installs python3. I didn't add this to the fix.sh/lint.sh scripts because pylint won't auto-fix problems and 2 of the scripts checked in do not validate.


pmarkowsky (approved):
LGTM.

PR #86, opened 2024-10-27
Standalone Mode (Beta) (1/3)

pmarkowsky:
This PR creates a new Santa operating mode -- standalone

This adds a new operating mode to Santa called standalone mode.

When running in standalone mode TouchID can be used to approved binaries. If a
binary is properly signed a SigningID rule is generated otherwise a SHA256 rule
is generated. 

When using a sync service standalone mode won't override explicit block rules, but will allow a user to approve anything blocked by default.

Video demo-ing standlone.

[Standalone Santa Operation](https://www.youtube.com/watch?v=Hd4t9mW-C-c)

>[!Note]
> This lacks a GUI for browsing rules in the local rule db

This supersedes #5.

Also the sync protocol changes are saved for a future PR. 

## To Do
- [x] Rework this as a new mode.
- [x] rework this to use new states for rule creation -- `SNTEventAllowLocalSHA256`, `SNTEventAllowLocalSigningID`
- [ ] rework this to emit two events (Future PR)
  - [x] First event is the normal event that is uploaded immediately
  - [ ] Create a new event when a new rule is created that is only a subset of the old event data to indicate that the user has created that rule (Future PR)
  - [ ] Add a new clean sync type to target standalone rules (Future PR)
  - [ ] make `sync_type` repeatable in sync protocol. (Future PR)
  - [x] make this only apply to unknown blocks.


Source/common/SNTConfigurator.h
```diff
@@ -310,6 +311,14 @@
 ///
 @property(readonly, nonatomic) BOOL enableSilentTTYMode;
 
+///
+///  When standalonemode is enabled, Santa will let a user allow a
+///  binary using biometric authentication.
```
pmarkowsky, line 316: This should read:

```
///  When standalonemode is enabled, Santa will let a user allow a
///  binary using TouchID.
```

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -269,41 +273,51 @@ struct SNTBinaryMessageWindowView: View {
...
-          {
-            OpenEventButton(customText: c.eventDetailText, action: openButton)
-          }
-          DismissButton(
-            customText: c.dismissText,
-            silence: preventFutureNotifications,
-            action: dismissButton
-          )
+      // Display the standalone error message to the user if one is provided.
+      if let errorMessage = standaloneErrorMessage {
```
mlw, line 291: consider using the same logic as for displaying the button here (calling `addStandaloneButton`) instead of the `if let` statement. Might be good to change the method name too so it's less generic.

mlw (changes requested):

Source/gui/SNTBinaryMessageWindowView.swift
pmarkowsky: Fair the `if let` was to handle the optional string as the `if let` keyword unboxes it from the optional same as Rust. 

I can wrap it in in a `addStandaloneButton` check

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -120,4 +141,45 @@ - (void)updateBlockNotification:(SNTStoredEvent *)event withBundleHash:(NSString
...
+    return;
+  }
+
+  LAContext *context = [[LAContext alloc] init];
+
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+  [context evaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics
+          localizedReason:[NSString stringWithFormat:@"Approve %@", self.event.signingID]
+                    reply:^(BOOL success, NSError *error) {
+                      if (self.replyBlock == nil) {
```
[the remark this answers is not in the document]
pmarkowsky: This was here originally as as defensive measure in case ew'd released the window before the block got scheduled in.

Will fix.

Source/common/SNTConfigurator.m
```diff
@@ -775,6 +778,11 @@ - (BOOL)enableSilentTTYMode {
   return number ? [number boolValue] : NO;
 }
 
+- (BOOL)enableStandaloneMode {
```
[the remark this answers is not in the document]
pmarkowsky: Can do.

Source/santad/SNTExecutionController.mm
```diff
@@ -375,7 +375,7 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
         // message to santad to perform the upload logic for bundles.
         // See syncBundleEvent:relatedEvents: for more info.
         se.needsBundleHash = YES;
-      } else if (config.syncBaseURL) {
+      } else if (config.syncBaseURL && !config.enableStandaloneMode) {
```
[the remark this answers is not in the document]
pmarkowsky: Should be fixed.

Source/gui/SNTBinaryMessageWindowView.swift
pmarkowsky: Wrapped this in a simple check

```swift
      // Display the standalone error message to the user if one is provided.
      if c.enableStandaloneMode {
        if let errorMessage = standaloneErrorMessage {
          if errorMessage != "" {
            Text(errorMessage).foregroundColor(.red)
          }
        }
      }
```

Source/santad/SNTExecutionController.mm
```diff
@@ -404,8 +404,28 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
           self->_ttyWriter->Write(targetProc, msg);
         }
 
+        void (^replyBlock)(BOOL) = ^void(BOOL authenticated) {
+        };
+
+        // Only allow a user in standalone mode to override a block if an
+        // explicit block rule is not set, when using a sync service.
+        if (config.enableStandaloneMode) {
+          if (!config.syncBaseURL || se.decision == SNTEventStateBlockUnknown) {
```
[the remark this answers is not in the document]
pmarkowsky: Fixed.

Source/santad/SNTExecutionController.mm
```diff
@@ -375,7 +375,7 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
         // message to santad to perform the upload logic for bundles.
         // See syncBundleEvent:relatedEvents: for more info.
         se.needsBundleHash = YES;
-      } else if (config.syncBaseURL) {
+      } else if (config.syncBaseURL && (!config.enableStandaloneMode || cd.silentBlock)) {
```
mlw, line 378: I think there might be an additional case here we didn't handle last time we looked at this. On line 413 below, we only setup the custom reply block if there isn't an explicit block rule (`se.decision == SNTEventStateBlockUnknown`). 

If there was an explicit block rule, (e.g. the decision was `SNTEventStateBlockBinary` or something) and `enableStandaloneMode` is set, then it seems the event would never be added to the sync queue. It wouldn't be added here because if check would fail. Then down below the custom `reply`block wouldn't be set either, resulting in no event being uploaded.

Source/santad/SNTExecutionController.mm
```diff
@@ -404,8 +404,27 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
           self->_ttyWriter->Write(targetProc, msg);
         }
 
+        void (^replyBlock)(BOOL) = ^void(BOOL authenticated) {
+        };
+
+        // Only allow a user in standalone mode to override a block if an
+        // explicit block rule is not set when using a sync service.
+        if (config.enableStandaloneMode &&
+            (!config.syncBaseURL || se.decision == SNTEventStateBlockUnknown)) {
```
mlw, line 413: I'm not sure this is the condition we want - "standalone enabled and no sync server configured" would be enough to succeed the condition without checking the decision, potentially allowing for a standalone rule to be created. But what about the case of static rules? We'd go thru the flow of letting the user approve some binary, but it would still fail to run due to static rules having higher precedence which would likely cause confusion.

mlw (changes requested):

docs/deployment/configuration.md
```diff
@@ -31,6 +31,7 @@ also known as mobileconfig files, which are in an Apple-specific XML format.
 | EnableBadSignatureProtection       | Bool       | If true, binaries with a bad signing chain will be blocked even in MONITOR mode, **unless** the binary is allowed by an explicit rule. Defaults to false. |
 | EnablePageZeroProtection           | Bool       | If true, 32-bit binaries that are missing the `__PAGEZERO` segment will be blocked even in MONITOR mode, **unless** the binary is allowed by an explicit rule. Defaults to true. |
 | EnableSilentMode                   | Bool       | If true, Santa will not post any GUI notifications. This can be a very confusing experience for users, use with caution. Defaults to false. |
+| EnableStandaloneMode                   | Bool       | If true, Santa will allow the local user to approve binaries provided they successfully authenticate with TouchID. Use with caution. Defaults to false. |
```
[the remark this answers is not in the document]
pmarkowsky: This was a general warning as in "beware, you're giving users lots of power"

Source/santad/SNTExecutionController.mm
pmarkowsky: Yeah thought about that and I think we need to figure out the right thing here or if I need to subclass the `SNTExecutionController` and just override it's behavior.

There's a few options here if we don't subclass:

1. Make the immediate upload happen if the decision is not `SNTEventStateBlockUnknown` and then add logging to the default replyBlock if `enableStandalone` mode and a `syncBaseURL` are both set for when a user dismisses the block dialog or fails authentication.

This is pretty easy and probably requires a helper function for the conditional as it's a little hairy and should be tested.

2. In standalone mode we can just defer all event uploads to the sync server until after the user has made the decision. 

This has the disadvantage of delaying when the sync service can display the event for approvals  / log or trigger bundle service interactions.

3. We can upload 2 events to the server, the first one if a `syncBaseURL` is configured (Santa's default behavior) and a second one if the user authenticates and approves the block. 

Option 3 would put an added burden on the sync service to determine what happened, but shouldn't be that crazy as the event identifiers and timestamps would all match.

Source/santad/SNTExecutionController.mm
pmarkowsky: If standalone is enabled and you do not have a sync service then local rule configuration is enabled via the command line via `sudo santactl rule`.

In which case the ability to bypass lockdown is tied to do you have root access (or at least sudo), admittedly this isn't always the same as having enrolled in Touch ID, so I probably need to add a check in the authorization part to ensure the user is in the `admin` group. 

If we want to prevent bypassing static rules we need a way to identify a rule as a static rule or make the check explicitly tied to `SNTEventStateBlockUnknown`.

pmarkowsky: > We'd go thru the flow of letting the user approve some binary, but it would still fail to run due to static rules having higher precedence which would likely cause confusion.

Ah I misread this. That'd be super frustrating, to avoid this I think we need to know that the rule is from the static rules and cannot be modified. I'll open a separate PR for this as it's useful to have in the `santactl fileinfo`  output.

Source/santad/SNTExecutionController.mm
pmarkowsky: We've decided to go with option 4 a new event type to notify the sync service of local rule creation. 

This will come in a future PR.

Source/common/SNTConfigurator.h
```diff
@@ -399,6 +399,12 @@
 ///
 @property(nullable, readonly, nonatomic) NSString *modeNotificationLockdown;
 
+///
+///  The notification text to display when the client goes into STANDALONE mode.
+///  Defaults to "Switching into Lockdown mode"
```
russellhancox, line 404:
```suggestion
///  Defaults to "Switching into Standalone mode"
```

Source/common/SNTCommonEnums.h
```diff
@@ -63,13 +63,15 @@ typedef NS_ENUM(NSInteger, SNTRuleState) {
...
+  SNTRuleStateAllowLocalBinary = 7,
+  SNTRuleStateAllowLocalSigningID = 8,
 };
 
 typedef NS_ENUM(NSInteger, SNTClientMode) {
   SNTClientModeUnknown,
-
   SNTClientModeMonitor = 1,
   SNTClientModeLockdown = 2,
+  SNTClientModeStandalone = 3,
```
russellhancox, line 74: Not necessary for this PR but we should get rid of this enum at some point and use the one from the sync proto.

Source/common/SNTConfigurator.m
```diff
@@ -14,6 +14,7 @@
 /// limitations under the License.
 
 #import "Source/common/SNTConfigurator.h"
+#include "Source/common/SNTCommonEnums.h"
```
russellhancox, line 17: This is already imported in the header

Source/common/SNTStoredEvent.m
```diff
@@ -70,6 +70,7 @@ - (void)encodeWithCoder:(NSCoder *)coder {
   ENCODE(self.quarantineRefererURL, @"quarantineRefererURL");
   ENCODE(self.quarantineTimestamp, @"quarantineTimestamp");
   ENCODE(self.quarantineAgentBundleID, @"quarantineAgentBundleID");
+  ENCODE([NSNumber numberWithBool:self.standaloneApproval], @"standaloneApproval");
```
russellhancox, line 73:
```suggestion
  ENCODE(@(self.standaloneApproval), @"standaloneApproval");
```

Source/common/SNTXPCNotifierInterface.h
```diff
@@ -25,7 +26,8 @@
 @protocol SNTNotifierXPC
 - (void)postBlockNotification:(SNTStoredEvent *)event
             withCustomMessage:(NSString *)message
-                 andCustomURL:(NSString *)url;
+                 andCustomURL:(NSString *)url
```
russellhancox, line 29: "and" should only be on the last argument, this can be just `customURL:`

Source/gui/BUILD
```diff
@@ -88,6 +89,7 @@ objc_library(
         "SecurityInterface",
         "SystemExtensions",
         "UserNotifications",
+        "LocalAuthentication",
```
russellhancox, line 92: Alphabetize

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -285,6 +287,7 @@ static void UpdateCachedDecisionSigningInfo(
 
   switch (mode) {
     case SNTClientModeMonitor: cd.decision = SNTEventStateAllowUnknown; return cd;
+    case SNTClientModeStandalone:  // Fallthrough
```
russellhancox, line 290:
```suggestion
    case SNTClientModeStandalone:  [[fallthrough]]
```

Source/gui/SNTMessageView.swift
```diff
@@ -149,6 +149,19 @@ public func OpenEventButton(customText: String? = nil, action: @escaping () -> V
...
+  Button(
+    action: action,
+    label: {
+      Text(NSLocalizedString("Approve", comment: "Default text for Approve")).frame(maxWidth: 200.0)
+    }
+  )
+  .keyboardShortcut(.return, modifiers: .command)
+  .help("⌘ Return")
+}
+
```
russellhancox, line 164: If this button is only used in the BinaryMessageWindow it should be declared there. SNTMessageView is intended for "shared" components

Source/gui/SNTBinaryMessageWindowController.m
```diff
@@ -120,4 +141,40 @@ - (void)updateBlockNotification:(SNTStoredEvent *)event withBundleHash:(NSString
   });
 }
 
+#pragma mark - Standlone mode methods
```
russellhancox, line 144: Could these two methods be pushed into the SwiftUI code?

It seems like `isAbleToAuthenticateInStandaloneMode:` is used to send an error into the UI but if that were in the UI we could more easily localize the error message itself. 

`approveBinaryForStandaloneMode` is a little less obvious but I think you could do the TouchID evaluation inside the SwiftUI code and then pass self.replyBlock() directly into the UI code instead of wrapping that in another callback?

russellhancox (commented):
Nits

Source/gui/SNTBinaryMessageWindowController.m
pmarkowsky: I can make that change. This is mostly left over from before we made the move to Swift UI. 

Part of this was also to make sure the window closed only after the request was processed. Hence the block wrapping back and forth.

Source/gui/SNTMessageView.swift
pmarkowsky: I put it here because I assumed at some point we'd extend to file access and USB.

Source/common/SNTStoredEvent.m
pmarkowsky: This should be entirely removed, as this field is no longer required after we move to the audit event.

docs/deployment/configuration.md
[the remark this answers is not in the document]
pmarkowsky: Resolving as switching this to another mode removes the option.

Source/gui/BUILD
pmarkowsky: Fixed.

Source/santad/SNTPolicyProcessor.mm
pmarkowsky: Fixed.

Source/gui/SNTBinaryMessageWindowController.m
pmarkowsky: The current code pushes the local authorization checks into swift.

Source/gui/SNTMessageView.swift
```diff
@@ -149,8 +150,56 @@ public func OpenEventButton(customText: String? = nil, action: @escaping () -> V
...
+    success,
+    authenticationError in
+    {
+      addRule = success
+      error = authenticationError as NSError?
+      semaphore.signal()
+    }()
+  }
+
+  semaphore.wait()
```
russellhancox, line 175: AuthorizeViaTouchID is being called from inside SwiftUI so it's running on the main thread - we should never block the main thread. Given that the callback is being passed into this function to call, why not remove the return value and leave it to run asynchronously?

Source/gui/SNTNotificationManager.m
```diff
@@ -355,7 +355,7 @@ - (void)postRuleSyncNotificationWithCustomMessage:(NSString *)message {
 
 - (void)postBlockNotification:(SNTStoredEvent *)event
             withCustomMessage:(NSString *)message
-                 andCustomURL:(NSString *)url
+                    CustomURL:(NSString *)url
```
russellhancox, line 358:
```suggestion
                    customURL:(NSString *)url
```

Source/gui/SNTNotificationManagerTest.m
```diff
@@ -60,7 +60,7 @@ - (void)testPostBlockNotificationSendsDistributedNotification {
 
   [sut postBlockNotification:ev
            withCustomMessage:@""
-                andCustomURL:@""
+                   CustomURL:@""
```
russellhancox, line 63:
```suggestion
                   customURL:@""
```

Source/santad/SNTExecutionController.mm
```diff
@@ -426,7 +426,7 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
         // Let the user know what happened in the GUI.
         [self.notifierQueue addEvent:se
                    withCustomMessage:cd.customMsg
-                        andCustomURL:cd.customURL
+                           CustomURL:cd.customURL
```
russellhancox, line 429:
```suggestion
                           customURL:cd.customURL
```

Source/santad/SNTNotificationQueue.h
```diff
@@ -24,7 +24,7 @@
 
 - (void)addEvent:(SNTStoredEvent *)event
   withCustomMessage:(NSString *)message
-       andCustomURL:(NSString *)url
+          CustomURL:(NSString *)url
```
russellhancox, line 27:
```suggestion
          customURL:(NSString *)url
```

Source/santad/SNTNotificationQueue.m
```diff
@@ -38,7 +38,7 @@ - (instancetype)init {
 
 - (void)addEvent:(SNTStoredEvent *)event
   withCustomMessage:(NSString *)message
-       andCustomURL:(NSString *)url
+          CustomURL:(NSString *)url
```
russellhancox, line 41:
```suggestion
          customURL:(NSString *)url
```

Source/santad/SNTNotificationQueue.m
```diff
@@ -81,7 +81,7 @@ - (void)flushQueue {
     for (NSDictionary *d in self.pendingNotifications) {
       [rop postBlockNotification:d[@"event"]
                withCustomMessage:d[@"message"]
-                    andCustomURL:d[@"url"]
+                       CustomURL:d[@"url"]
```
russellhancox, line 84:
```suggestion
                       customURL:d[@"url"]
```

Source/gui/SNTMessageView.swift
pmarkowsky: The main issue here is that if the replyBlock call doesn't complete before the window is closed then the rules may not apply. 

This has led to some funky behavior with standalone mode in the past.

Previously I was using a callback from the replyBlock to close the window to avoid this.

tmp-rules.json
mlw, line 1: This file seems inadvertently added.

mlw (changes requested):

Source/common/SNTRule.m
```diff
@@ -254,6 +259,8 @@ - (NSString *)ruleStateToPolicyString:(SNTRuleState)state {
     case SNTRuleStateSilentBlock: return kRulePolicySilentBlocklist;
     case SNTRuleStateRemove: return kRulePolicyRemove;
     case SNTRuleStateAllowTransitive: return @"AllowTransitive";
+    case SNTRuleStateAllowLocalBinary: return @"ALLOWLIST_LOCAL_BINARY";
```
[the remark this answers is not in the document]
pmarkowsky: I explicitly left this out as it was unclear if we wanted this to be in the sync protocol or treated the same as transitive rules. 

Happy to add this to the sync constants.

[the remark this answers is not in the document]
pmarkowsky: Fixed.

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -312,11 +341,50 @@ struct SNTBinaryMessageWindowView: View {
...
     }
+
+  }
+
+  // This button is only shown when the standalone mode is enabled in place of
+  // the "Open Event" button.
+  func standAloneButton() {
+
+    var msg = "Allow execution of "
+
```
russellhancox, line 363:
```suggestion
    var msg = "allow execution of "

```

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -312,11 +341,50 @@ struct SNTBinaryMessageWindowView: View {
...
+    if let e = self.event {
+      let bundleName = e.fileBundleName ?? ""
+      let filePath = e.filePath ?? ""
+      let signingID = e.signingID ?? ""
+
+      if !bundleName.isEmpty {
+        msg = NSLocalizedString("the application " + msg + bundleName, comment: "Bundle name")
```
russellhancox, line 370:
```suggestion
        msg = NSLocalizedString(msg + "the application " + bundleName, comment: "Bundle name")
```

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -312,11 +341,50 @@ struct SNTBinaryMessageWindowView: View {
...
+      if let cb = self.replyCallback {
+        cb(false)
+      }
+      return
+    }
+
+    if let callback = self.replyCallback {
+      _ = AuthorizeViaTouchID(reason: msg, replyBlock: callback)
+    }
+    window?.close()
```
russellhancox, line 387:
```suggestion
    if let callback = self.replyCallback {
      AuthorizeViaTouchID(reason: msg, replyBlock: { success in 
        callback(success)
        DispatchQueue.main.sync {
          window?.close()
        }
      })
    }
```

Source/gui/SNTMessageView.swift
```diff
@@ -149,6 +150,67 @@ public func OpenEventButton(customText: String? = nil, action: @escaping () -> V
...
+  if error != nil {
+    replyBlock(false)
+    return (false, error)
+  }
+
+  replyBlock(addRule)
+  return (addRule, nil)
+}
```
russellhancox, line 183:
```suggestion
public func AuthorizeViaTouchID(reason: String, replyBlock: @escaping (Bool) -> Void) {
  LAContext().evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, localizedReason: reason) { success, error in
    if error != nil {
      replyBlock(false)
    } else {
      replyBlock(success)
    }
  }
}
```

russellhancox (changes requested):
These 4 changes go together to avoid blocking the main thread. I've tested them locally

Source/gui/SNTBinaryMessageWindowView.swift
russellhancox: You might actually have to copy this into the NSLocalizedString on lines 371 and 373, otherwise the localization tools won't pick it up properly

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -312,11 +341,50 @@ struct SNTBinaryMessageWindowView: View {
...
   }
```
russellhancox, line 388:
```suggestion
  func standAloneButton() {
    guard let e = self.event else {
      // If we don't have an event, we can't fill in the message.
      if let cb = self.replyCallback {
        cb(false)
      }
      return
    }
 
    let bundleName = e.fileBundleName ?? ""
    let filePath = e.filePath ?? ""
    let signingID = e.signingID ?? ""
    
    var msg = "authorize execution"    

    if !bundleName.isEmpty {
      msg = NSLocalizedString("authorize execution of the application \(bundleName)", comment: "Bundle name")
    } else if !signingID.isEmpty {
      msg = NSLocalizedString("authorize execuction of \(signingID)", comment: "Signing ID")
    } else if !filePath.isEmpty {
      msg = NSLocalizedString("authorize execution of \((filePath as NSString).lastPathComponent)", comment: "File path")
    }

    if let callback = self.replyCallback {
      _ = AuthorizeViaTouchID(reason: msg, replyBlock: callback)
    }
    window?.close()
  }
```

Source/gui/SNTMessageView.swift
pmarkowsky: The solution was to schedule the window close on the main thread from the callback using dispatch.

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -358,33 +358,47 @@ struct SNTBinaryMessageWindowView: View {
...
+        comment: "Signing ID"
+      )
+    } else if !filePath.isEmpty {
+      msg = NSLocalizedString(
+        "authorize execution of " + filePath,
+        comment: "File path"
+      )
+    }
+
     if let callback = self.replyCallback {
```
russellhancox, line 391: Need to handle the else case here. Or force-unwrap, because it seems to me it's a programming error if there is no callback

russellhancox (approved):

mlw (approved):

PR #88, opened 2024-10-28
Project: Ensure Swift code is properly formatted

russellhancox:


.swift-format
```diff
@@ -0,0 +1,18 @@
...
+    "indentConditionalCompilationBlocks": false,
+    "prioritizeKeepingFunctionOutputTogether": true,
+    "rules": {
+        "AlwaysUseLowerCamelCase": false,
+        "AmbiguousTrailingClosureOverload": false,
+        "NoBlockComments": false,
+        "OrderedImports": true,
+        "UseLetInEveryBoundCaseVariable": false,
+        "UseSynthesizedInitializer": false
+    }
```
mlw, line 17: Consider: `blankLinesBetweenImports: false` - I like how the original version separated SDK imports vs. other imports.

russellhancox: `blankLinesBetweenImports` didn't work, so I had to disable `OrderedImports` instead

mlw: Oh bummer :( But sounds good re: `OrderedImports`. Maybe things will be fixed in the future...

mlw (approved):

PR #92, opened 2024-10-31
Move away from /Library/Caches for migration

mlw:
This change only moves the scratch dir used for migration out of /Library/Caches due to external factors that can cause that directory to be blown away (such as OS upgrades).


pmarkowsky (approved):
LGTM.

PR #98, opened 2024-11-01
Only set AT_SYMLINK_NOFOLLOW_ANY on supported OS versions

mlw:
`unlinkat(2)` didn't gain support for `AT_SYMLINK_NOFOLLOW_ANY` until macOS 14. Don't set this flag on earlier OS versions.


pmarkowsky (approved):
LGTM.

PR #106, opened 2024-11-06
Add proto for new authentication events

mlw:


russellhancox (approved):

Source/common/santa.proto
```diff
@@ -757,6 +757,126 @@ message OpenSSH {
...
+// Information related to OpenDirectory authentication
+message AuthenticationOD {
+  // The process that emitted the authentication event
+  optional ProcessInfoLight instigator = 1;
+
+  // The process that triggered authentication
+  // Note: Due internal system limitations, the process that triggered the
+  // event may have already exited before the event could be generated. This
+  // results in only a small subset of the information from the triggering
+  // process to be reported.
```
pmarkowsky, line 769: Cna we use the process tree to fill this in after the fact?

mlw: potentially. Would need to store a bunch of information though on all executing processes (i.e. most everything in the `es_message_t`) to be able to do the lookup after the fact and then have the available data to populate the telemetry event. Would need to be careful about mem usage for a rare use case. We'd already be able to do the mapping on the backend cause we'd have the original exec log so the savings here would be a backend join.

PR #114, opened 2024-11-11
UI: Make images work in messages again

### LINKED ISSUE #113: UI: Images in customizable messages are not rendered
Since switching Santa's UI over to SwiftUI, custom messages (from `BannedBlockMessage`, `UnknownBlockMessage` and the like) that contain `` tags do not have the images rendered.

E.g., this used to work:

```
<key>BannedBlockMessage</key>
<string>&lt;img src="data:image/png;base64,<YOUR IMAGE HERE>" /&gt;&lt;br /&gt;Hello from Santa. Your toys have been taken away.</string>
```

Resulting in:

image

With the new SwiftUI updates, we instead see this:

image

This appears to be an issue with the `Text()` type in SwiftUI throwing away images that are "attachments" to an NSAttributedString.

russellhancox:
Fixes #113

### DIFF
--- a/Source/gui/SNTMessageView.swift
+++ b/Source/gui/SNTMessageView.swift
@@ -1,4 +1,5 @@
 import SwiftUI
+
 import santa_common_SNTConfigurator
 
 public let MAX_OUTER_VIEW_WIDTH = 560.0
@@ -48,7 +49,7 @@ public struct SNTMessageView<Content: View>: View {
 
       Spacer()
 
-      Text(AttributedString(blockMessage))
+      AttributedText(blockMessage)
         .multilineTextAlignment(.center)
         .padding([.leading, .trailing], 15.0)
         .fixedSize()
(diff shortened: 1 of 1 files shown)


Source/gui/SNTMessageView.swift
```diff
@@ -161,3 +162,19 @@ public func DismissButton(
   .keyboardShortcut(.escape, modifiers: .command)
   .help("⌘ Esc")
 }
+
+// AttributedText is like Text() but it supports all the features of NSAttributedString()
+// by using NSTextField under the hood.
```
pmarkowsky, line 167: Does this mean we can use markdown here? 

Asking because I skimmed https://developer.apple.com/documentation/foundation/attributedstring/instantiating_attributed_strings_with_markdown_syntax and it looks like we might support it with this change.

mlw (approved):

pmarkowsky (approved):
LGTM.

Source/gui/SNTMessageView.swift
russellhancox: No, it doesn't. The actual message goes through SNTBlockMessage first where either HTML tags are stripped out or more HTML is added to make it look right. 

Markdown has to be initialized with a different initializer and it's hard to embed in a plist with newlines, so it would be much harder to use correctly than the HTML support we currently have.

PR #123, opened 2024-11-13
santad/santactl: Add comment field for rules, populate for manual rules

### LINKED ISSUE #112: Manually added rules should automatically populate the `message` field with associated identifiers
When a rule is manually added via `santactl rule`, identifiers used to create the rule (such as path, signing ID, etc) should be automatically added into the `custommsg` column of the `rules` table. While not perfect (e.g., paths may not be descriptive and can change), it would generally help users identify source of rules when they're manually reviewed.

Created from discussion: #111

russellhancox:
Fixes #112

### DIFF
--- a/Source/common/SNTRule.h
+++ b/Source/common/SNTRule.h
@@ -53,13 +53,19 @@
 @property(readonly) NSUInteger timestamp;
 
 ///
+///  A comment attached to this rule. This is intended only for local rules.
+///
+@property(copy) NSString *comment;
+
+///
 ///  Designated initializer.
 ///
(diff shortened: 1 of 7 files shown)


mlw (approved):
Some high level comments, not really blockers:

1. Would be nice to support this in the sync protocol so comments could be sent down
2. Would be nice to have some way to have this information be dumped outside of opening up the DB. Maybe `santactl fileinfo` and/or `santactl rule --check` should do this.

Don't need to implement these now, but if not opposed to these actions, we should file a couple of issues.

PR #141, opened 2024-11-19
gui: Fix crash unwrapping optional in EventDetailURL

russellhancox:


Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -288,7 +288,7 @@ struct SNTBinaryMessageWindowView: View {
         }
 
         HStack(spacing: 15.0) {
-          if c.eventDetailURL.count > 0
+          if c.eventDetailURL?.count ?? 0 > 0
             && !(event?.needsBundleHash ?? false && !bundleProgress.isFinished)
           {
             OpenEventButton(customText: c.eventDetailText, action: openButton)
```
tburgin, line 294: Do we need to do similar unwrapping for c.eventDetailText and c.dismissText? I wonder what this is not a compile error? Is this an implicit force unwrap?

tburgin (approved):

Source/gui/SNTBinaryMessageWindowView.swift
russellhancox: > Do we need to do similar unwrapping for c.eventDetailText and c.dismissText?

We are, inside OpenEventButton and DismissButton

> I wonder what this is not a compile error? Is this an implicit force unwrap?

IIUC, because SNTConfigurator doesn't annotate nullable/nonnull on any of these properties, Swift imports them as implicitly unwrapped optionals. I'm gonna look at fixing that in a follow-up

pmarkowsky: Do we always set eventDetailText and dismissText with a default.

tburgin: > We are, inside OpenEventButton and DismissButton
Ah, I see. Perfect!

PR #146, opened 2024-11-19
Config: Add nullability annotations to configurator

russellhancox:


Source/common/SNTConfigurator.h
```diff
@@ -18,6 +18,8 @@
 
 @class SNTRule;
 
+NS_ASSUME_NONNULL_BEGIN
```
mlw, line 21: I think it would help clarity to explicitly mark all call sites. Given this PR has marked almost everything already, does it save much?

russellhancox: Done

PR #154, opened 2024-11-30
Fix: Database Rule Table Version number

pmarkowsky:
This PR fixes an issue with the SNTRuleTable's `currentSupportedVersion` method returning the incorrect value due to the constant it wraps not being updated. 

It makes the `currentSupportedVersion` method part of the interface, so as to expose for testing and metrics. 

A unit test is added to ensure that this constant is updated in the future.

This  bug was introduced in #123 was causing all rules to be dropped across builds / restarts.

This PR also adds the imports for the BUILD files to satisfy the linter.

### DIFF
--- a/Source/santad/DataLayer/SNTRuleTable.h
+++ b/Source/santad/DataLayer/SNTRuleTable.h
@@ -122,4 +122,9 @@
 ///
 - (uint32_t)currentSupportedVersion;
 
+///
+/// The current version of the rule table schema.
+///
+@property(readonly, nonatomic) uint32_t currentVersion;
+
 @end
--- a/Source/santad/DataLayer/SNTRuleTable.h
+++ b/Source/santad/DataLayer/SNTRuleTable.h
@@ -117,9 +117,4 @@
 @property(readonly, nonatomic)
   NSDictionary<NSString *, SNTCachedDecision *> *criticalSystemBinaries;
 
-///
-///  Current supported version of the rule table schema.
-///
-- (uint32_t)currentSupportedVersion;
-
 @end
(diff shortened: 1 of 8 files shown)


Source/santad/DataLayer/SNTRuleTableTest.m
```diff
@@ -533,4 +534,10 @@ - (void)testCriticalBinariesProduceFullSigningInformation {
   XCTAssertEqualObjects(teamID, cd.teamID, @"team IDs should match");
 }
 
+// This test ensures that we bump the constant on updates to the rule table
+// schema.
+- (void)testConstantVersionIsUpdated {
```
russellhancox, line 539: This test wouldn't have prevented this bug. The test needs to initialize a database and check the version matches the expected version

Source/santad/DataLayer/SNTRuleTable.h
```diff
@@ -117,4 +117,14 @@
...
 
+///
+///  Current supported version of the rule table schema.
+///
+- (uint32_t)currentSupportedVersion;
+
+///
+/// The current version of the rule table schema.
+///
+@property(readonly, nonatomic) uint32_t currentVersion;
```
russellhancox, line 128: I think this should be a readonly property on SNTDatabaseTable that returns the version from SQLite (`[db userVersion]`) as that will always match what the initializer is provided. This would replace the `currentSchemaVersion` property

Source/santad/DataLayer/SNTRuleTable.h
```diff
@@ -117,4 +117,14 @@
 @property(readonly, nonatomic)
   NSDictionary<NSString *, SNTCachedDecision *> *criticalSystemBinaries;
 
+///
+///  Current supported version of the rule table schema.
+///
+- (uint32_t)currentSupportedVersion;
```
russellhancox, line 123: Move this up to SNTDatabaseTable as it's defined in all subclasses

russellhancox (approved):

PR #159, opened 2024-12-05
standalone: Allow password fallback

russellhancox:
This makes standalone mode usable in clamshell mode or on a desktop machine.

This change also modifies the authorization dialog for unsigned, non-bundled binaries to just show the binary name instead of the full path. In many cases the full path is very long and as the dialog is limited in width it makes the path squash up and become nearly unreadable.

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -801,7 +801,7 @@ static SNTConfigurator *sharedConfigurator = nil;
 
 - (BOOL)enableStandalonePasswordFallback {
   NSNumber *number = self.configState[kEnableStandalonePasswordFallbackKey];
-  return number ? [number boolValue] : NO;
+  return number ? [number boolValue] : YES;
 }
 
 - (BOOL)enableSilentMode {
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -92,6 +92,7 @@ static NSString *const kMachineOwnerPlistKeyKey = @"MachineOwnerKey";
 static NSString *const kMachineIDPlistFileKey = @"MachineIDPlist";
 static NSString *const kMachineIDPlistKeyKey = @"MachineIDKey";
 
+static NSString *const kEnableStandalonePasswordFallbackKey = @"EnableStandalonePasswordFallback";
 static NSString *const kEnableSilentModeKey = @"EnableSilentMode";
 static NSString *const kEnableSilentTTYModeKey = @"EnableSilentTTYMode";
 static NSString *const kAboutTextKey = @"AboutText";
@@ -226,6 +227,7 @@ static NSString *const kSyncTypeRequired = @"SyncTypeRequired";
       kOnStartUSBOptions : string,
       kEnablePageZeroProtectionKey : number,
       kEnableBadSignatureProtectionKey : number,
+      kEnableStandalonePasswordFallbackKey : number,
       kEnableSilentModeKey : number,
       kEnableSilentTTYModeKey : number,
       kAboutTextKey : string,
@@ -416,6 +418,10 @@ static SNTConfigurator *sharedConfigurator = nil;
 }
 
 + (NSSet *)keyPathsForValuesAffectingEnablePageZeroProtection {
+  return [self configStateSet];
+}
+
++ (NSSet *)keyPathsForValuesAffectingEnableStandalonePasswordFallback {
   return [self configStateSet];
 }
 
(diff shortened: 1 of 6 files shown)


Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -383,7 +383,7 @@ struct SNTBinaryMessageWindowView: View {
       )
     } else if !filePath.isEmpty {
       msg = NSLocalizedString(
-        "authorize execution of " + filePath,
+        "authorize execution of " + (filePath as NSString).lastPathComponent,
```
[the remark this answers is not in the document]
russellhancox: I see your point but the full path is in the "More Details" section of the original dialog. And I think the majority of users aren't going to actually look at the path in this prompt because it's too long:

image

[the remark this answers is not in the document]
pmarkowsky: I don't disagree. The right answer here is probably doc updates to emphasize being careful.

docs/concepts/mode.md
```diff
@@ -27,14 +27,27 @@ Running Santa in Lockdown Mode will stop all blocked binaries and additionally
...
+as a mouse/keyboard. Unfortunately this makes Standalone mode unusable for users
+on most desktop machines or using a laptop in clamshell mode. If supporting that
+use-case is important there is a configuration option
+`EnableStandalonePasswordFallback` that can be enabled to allow falling back to
+password authentication when TouchID is not available.
 
 {: .note}
-Standalone mode will not override explicit block rules when Santa is configured to use a sync service nor will it override static rules.
+Standalone mode will not override explicit block rules when Santa is configured
+to use a sync service nor will it override static rules.
```
pmarkowsky, line 50:
```suggestion
Standalone mode will not override explicit block rules nor will it override static rules.
```

We'd changed this behavior so this is more accurate.

pmarkowsky (changes requested):
Minor things.

Source/common/SNTConfigurator.m
```diff
@@ -793,6 +799,11 @@ - (BOOL)enableBadSignatureProtection {
   return number ? [number boolValue] : NO;
 }
 
+- (BOOL)enableStandalonePasswordFallback {
+  NSNumber *number = self.configState[kEnableStandalonePasswordFallbackKey];
+  return number ? [number boolValue] : NO;
```
pmarkowsky, line 804:
```suggestion
  return number ? [number boolValue] : YES;
```

docs/concepts/mode.md
```diff
@@ -27,14 +27,26 @@ Running Santa in Lockdown Mode will stop all blocked binaries and additionally
...
+This allows a user to quickly self approve in lieu of using a sync service. If
+one is using a sync service Events will still be sent up to that sync service.
+
+Standalone mode typically requires TouchID for self-approval because this
+requires a physical interaction and cannot be spoofed by other software acting
+as a mouse/keyboard. Unfortunately this makes Standalone mode unusable for users
+on most desktop machines or using a laptop in clamshell mode. If supporting that
+use-case is important there is a configuration option
+`EnableStandalonePasswordFallback` that can be enabled to allow falling back to
+password authentication when TouchID is not available.
```
pmarkowsky, line 46:
```suggestion
Standalone mode typically requires TouchID  or password to approve binary execution.

The configuration option `EnableStandalonePasswordFallback` can be disabled to make Standalone mode require physical access. When disabled only TouchID will be used for self-approval because this requires a physical interaction and cannot be spoofed by other software acting as a mouse/keyboard.
```

docs/deployment/configuration.md
```diff
@@ -30,6 +30,7 @@ also known as mobileconfig files, which are in an Apple-specific XML format.
 | FileChangesPrefixFilters           | Array      | Array of path prefix strings. When an event is logged, if the target path (e.g. the file being written/removed/etc ) matches a prefix it will not be logged. |
 | EnableBadSignatureProtection       | Bool       | If true, binaries with a bad signing chain will be blocked even in MONITOR mode, **unless** the binary is allowed by an explicit rule. Defaults to false. |
 | EnablePageZeroProtection           | Bool       | If true, 32-bit binaries that are missing the `__PAGEZERO` segment will be blocked even in MONITOR mode, **unless** the binary is allowed by an explicit rule. Defaults to true. |
+| EnableStandalonePasswordFallback   | Bool       | If true, Santa will fallback to password authorization for Standalone mode. |
```
pmarkowsky, line 33:
```suggestion
| EnableStandalonePasswordFallback   | Bool       | If true, Santa will fallback to password authorization for Standalone mode. Defaults to YES |
```

pmarkowsky (approved):
LGTM.

PR #161, opened 2024-12-06
Restructure Watch Items objects to prepare for Proc FAA

mlw:
This PR starts the process of breaking up `WatchItemPolicy` into `DataWatchItemPolicy` and `ProcessWatchItemPolicy` to as a first step in implementing Proc FAA.

`WatchItems` is also being broken up to abstract out the `DataWatchItems` into their own object so that it can be operated on singularly. A future PR will add `ProcessWatchItems` to work similarly.

### DIFF
--- a/Source/santad/DataLayer/WatchItemPolicy.h
+++ b/Source/santad/DataLayer/WatchItemPolicy.h
@@ -24,6 +24,9 @@
 #include <string_view>
 #include <vector>
 
+#import "Source/common/PrefixTree.h"
+#import "Source/common/Unit.h"
+
 namespace santa {
 
 enum class WatchItemPathType {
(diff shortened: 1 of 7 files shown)


Source/santad/DataLayer/WatchItemPolicy.h
```diff
@@ -120,7 +118,56 @@ struct WatchItemPolicy {
...
+  }
+
+  bool operator!=(const WatchItemPolicyBase &other) const override { return !(*this == other); }
+
+  std::string path;
+  WatchItemPathType path_type;
+  std::vector<WatchItemProcess> processes;
+};
+
+struct ProcessWatchItemPolicy : public WatchItemPolicyBase {
```
mlw, line 156: Note: This is very incomplete right now. Just meant to be a stub to start getting interfaces worked out.

Source/santad/DataLayer/WatchItems.h
```diff
@@ -67,16 +67,47 @@ struct WatchItemsState {
...
+  std::unique_ptr<santa::PrefixTree<std::shared_ptr<DataWatchItemPolicy>>> tree_;
+  std::set<std::pair<std::string, WatchItemPathType>> paths_;
+};
+
 class WatchItems : public std::enable_shared_from_this<WatchItems> {
  public:
   using VersionAndPolicies =
-      std::pair<std::string, std::vector<std::optional<std::shared_ptr<WatchItemPolicy>>>>;
-  using WatchItemsTree = santa::PrefixTree<std::shared_ptr<WatchItemPolicy>>;
+      std::pair<std::string, std::vector<std::optional<std::shared_ptr<DataWatchItemPolicy>>>>;
```
russellhancox, line 106: WDYT about adding more using decls to make this nesting a little simpler and as some of the types are used in other places? E.g. `using DataWatchItemPolicySharedPtr = std::shared_ptr<DataWatchItemPolicy` can be used on line 133 and would make the nesting here shorter. And then perhaps `using VectorOptionals = std::vector<std::optional>`?

Just a thought, feel free to ignore

russellhancox (approved):

Source/santad/DataLayer/WatchItems.h
mlw: I think I over-indexed on the Google style guide of not having `using` in header files, but that isn't really the guidance. I've used some in the impl files, but I'm not consistent. For now, I'll keep as-is, but will look into simplifying these types in part 2!

PR #174, opened 2024-12-16
Standalone: Switch to a hold-and-ask model (part 1)

russellhancox:
With these changes, instead of denying execution and then making a rule after authorization, now when in Standalone mode the execution will be allowed but the binary immediately paused with `pid_suspend()`. If the user authorizes the execution, the binary will be resumed with `pid_resume()` (as well as a rule being created). While a process is stopped pending user authorization, any other process attempting to send `pid_resume()` will be blocked. If the GUI is unavailable for any reason, the process will immediately be killed.

This PR does not change logging, so currently all held executions will be logged as if they were an allowed execution even though they may be killed later before the binary ever had the change to execute anything. This will be fixed in a follow-up PR.


Source/santad/SNTNotificationQueue.m
```diff
@@ -69,12 +69,30 @@ - (void)addEvent:(SNTStoredEvent *)event
   @synchronized(self.pendingNotifications) {
     [self.pendingNotifications addObject:d];
   }
-  [self flushQueue];
+  [self flushQueueWithReplies:(reply != nil)];
```
mlw, line 72: As written today, `reply` is never `nil`, an empty block is provided. Will need to rework this flow. I think we probably should support nil block, but then the agent side will need some more smarts to first check for non-nil.

Source/santad/SNTExecutionController.mm
```diff
@@ -92,6 +93,7 @@ @implementation SNTExecutionController {
   absl::Mutex _entitlementFilterMutex;
   std::set<std::string> _entitlementsTeamIDFilter;
   std::unique_ptr<PrefixTree<Unit>> _entitlementsPrefixFilter;
+  std::unique_ptr<SantaCache<pid_t, bool>> _procSignalCache;
```
mlw, line 96: We should really make sure to use pid/pidver pair here. The concern is a bit far fetched, but there are a few paths through where a SIGSTOPped process could exit before an XPC reply to cleanup (either due to connection being invalidated, or one of the few ways through the kernel where a signal can be sent without hitting the MAC hook). If a PID is recycled, we would easily get into a weird state.

Source/santad/SNTNotificationQueue.m
russellhancox: Good point. I've made it so that now it defaults to being nil coming from SNTExecutionController but an empty reply block will be added for those cases in SNTNotificationQueue as passing a nil reply block over XPC is not allowed.

Source/santad/SNTExecutionController.mm
russellhancox: Fair enough, I've switched to using a std::pair to hold both. This necessitated some changes to SantaCache - I could have added a SantaCacheHasher specialization but I've switched over to using absl::Hash instead as that handles most things we want to hash and almost certainly does a better job than what I came up with several years ago.

mlw: Love this change to `absl::Hash`!

Source/santad/SNTExecutionController.mm
```diff
@@ -538,4 +603,17 @@ - (void)createRuleForStandaloneModeEvent:(SNTStoredEvent *)se {
   // TODO: Notify the sync service of the new rule.
 }
 
+extern "C" int pid_suspend(pid_t pid);
+extern "C" int pid_resume(pid_t pid);
+
+// Wrapper around the pid_suspend() / pid_resume() / kill() functions that uses signal numbers
+// to determine which to use and which can be easily mocked out in tests;
+- (void)sendSignal:(int)signal toPID:(pid_t)pid {
```
mlw: Conceptually I'm not a fan of this naming since 2 of 3 operations aren't a signal. Maybe add some new enum type:
```
enum class ProcessControl {
  kSuspend,
  kResume,
  kKill,
};
```
and change the signature to something like
```
- (void)manipulateProcess:(pid_t)pid withControl:(ProcessControl)control
```

Source/santad/SNTExecutionController.mm
```diff
@@ -312,8 +316,22 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
...
 
-  // Respond with the decision.
-  postAction(action);
+  pid_t newProcPid = audit_token_to_pid(targetProc->audit_token);
+  std::pair<pid_t, int> pidAndVersion =
+      std::make_pair(newProcPid, audit_token_to_pidversion(targetProc->audit_token));
+  if (cd.decision == SNTEventStateBlockUnknown && config.clientMode == SNTClientModeStandalone) {
+    // In standalone mode we want hold off on making a decision until the user has had a chance to
+    // approve. ES won't let us do this, we'd hit the response deadline. Instead, we send a SIGSTOP
+    // to hold stop the binary from executing but we respond to ES with an allow decision. If the
```
mlw, line 326: comment needs updating

mlw (changes requested):

Source/santad/SNTNotificationQueue.m
```diff
@@ -69,20 +69,45 @@ - (void)addEvent:(SNTStoredEvent *)event
...
 }
 
-- (void)flushQueue {
+- (void)flushQueueWithReplies:(BOOL)hasReplies {
```
mlw, line 75: If a machine is configured to use standalone mode, I think this will result in no notifications being queued when a user is logged out. Is that what we want? We might instead want to perform the `reply`, and add the notification back so that a user can be informed when logging back in.

Also might want to consider if this `hasReplies` is necessary. We likely want to flush (or reply and re-add) entries any time there isn't an XPC connection, not just when trying to add a new event.
I think this could maybe be simplified. It shouldn't matter if the current event has a reply block. I think we should always remove notifications

russellhancox: Yeah this wasn't the best. I've re-worked things to put flushQueue back to how it was before but now we only save the reply block in the dict if a notifier connection is currently available. There's still the possibility that the connection goes away after the block is saved in the dict but this should be very rare.

Source/santad/SNTNotificationQueue.m
```diff
@@ -79,10 +83,17 @@ - (void)flushQueue {
   @synchronized(self.pendingNotifications) {
     NSMutableArray *postedNotifications = [NSMutableArray array];
     for (NSDictionary *d in self.pendingNotifications) {
+      void (^reply)(BOOL authenticated) = d[@"reply"];
+      if (reply == nil) {
+        // The reply block sent to the GUI cannot be nil.
+        reply = ^(BOOL _) {
```
mlw, line 89: We should maybe revert to the old style and expect a block be passed in (then also add code to check for nil replies and bail). We previously needed to ensure we always copied blocks to ensure stack-based blocks got moved to the heap. We would presumably hit the same issue here.

russellhancox: I've added a copy of the reply block. I like the idea of SNTNotificationQueue being able to tell whether there is a reply block even if we're not currently using it.

mlw: This does do two copies now for blocks passed in, but not really a huge deal.

mlw (dismissed):

mlw (approved):

PR #177, opened 2024-12-17
initial Japanese localization

arubdesu:
All strings as of late 2024 AFAICT, spot-checked by Koji Arita.
May have issue on "Prevent future notifications for this application for " because sentence doesn't end that way grammatically in Japanese, trusting that `\n\n`'s & `<br />`'s will be interpreted correctly.

### DIFF
--- a/Source/gui/Resources/ja.lproj/Localizable.strings
+++ b/Source/gui/Resources/ja.lproj/Localizable.strings
@@ -1,0 +1,81 @@
+/* The default message to show the user when access to a file is blocked */
+"Access to a file has been denied" = "ファイルへのアクセスが拒否されました";
+
+/* No comment provided by engineer. */
+"Application" = "アプリケーション";
+
+/* No comment provided by engineer. */
+"Bundle Hash" = "バンドルハッシュ";
+
+/* No comment provided by engineer. */
+"CDHash" = "CDハッシュ";
+
+/* Copy Details button in more details dialog */
+"Copy Details" = "詳細をコピー";
+
+/* Dismiss button in more details dialog */
+"Dismiss" = "閉じる";
+
+/* No comment provided by engineer. */
+"Dismiss & Silence" = "閉じて通知を停止";
+
+/* No comment provided by engineer. */
+"Filename" = "ファイル名";
+
+/* More Details button in binary block dialog */
+"More Details" = "さらに詳細";
+
+/* No comment provided by engineer. */
+"More Info..." = "詳細情報…";
+
+/* No comment provided by engineer. */
+"Parent" = "親プロセス";
+
+/* No comment provided by engineer. */
+"Path" = "パス";
+
+/* No comment provided by engineer. */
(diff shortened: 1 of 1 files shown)


russellhancox:
image

image

russellhancox:
```----- Language ja -----
Missing: 4, ['Approve', 'Open...', 'authorize execution of ', 'authorize execution of the application ']
Extraneous: 0, []
Complete: 87.10%
```

3 of these missing strings are new and only apply to the new standalone mode but `Open...` is very important

arubdesu:
Pardon, this should match en.lproj's strings now.
Please note again that grammatically 'authorize execution _space_' strings wouldn't work to have the app name follow afterwards (with the space boundary) as spaces are omitted in Japanese text, this is another place where an inline variable would/should be used

Source/gui/Resources/ja.lproj/Localizable.strings
```diff
@@ -0,0 +1,94 @@
...
+"Santa is an application control system for macOS.\n\nThere are no user-configurable settings." = "SantaはmacOS向けアプリケーション制御システムです\n\n　ユーザーによる設定項目はありません";
+
+/* No comment provided by engineer. */
+"SHA-256" = "SHA-256";
+
+/* No comment provided by engineer. */
+"Signing ID" = "署名ID";
+
+/* The default message to show the user when an unknown application is blocked */
+"The following application has been blocked from executing<br />because its trustworthiness cannot be determined" = "次のアプリケーションは信頼性を判断できないため、<br />実行がブロックされました
```
russellhancox, line 78: This string isn’t terminated

russellhancox:
> Pardon, this should match en.lproj's strings now. Please note again that grammatically 'authorize execution _space_' strings wouldn't work to have the app name follow afterwards (with the space boundary) as e.g. the order of the phonetic word 'application' goes in the beginning/front of the sentence AND spaces are omitted in Japanese text, this is another place where an inline variable would/should be used

Interesting, I’ll have a look at updating the translations to include the app name tomorrow. Let’s get this in as-is for now and I’ll @ you when we can update to be more grammatically correct

russellhancox (changes requested):

Source/gui/Resources/ja.lproj/Localizable.strings
arubdesu: fixicated, pardon

russellhancox (approved):

russellhancox:
Thanks!

arubdesu:
Just to reiterate, I mentioned in the previous comment where variables either aren't currently used in English or other localizations or wouldn't be appended to the end of the sentence for Japanese, please take into account that:
```"Prevent future notifications for this application for " = "このアプリケーションからの通知を停止します```
should have the variable of the time period be between このアプリケーションからの通知をand 停止します
Also "authorize execution of the application " should instead be pre-pended, so it would be for example "Slackアプリケーションの実行を許可する"

PR #188, opened 2024-12-27
Adopt ring buf for queued notifications

mlw:
This change makes the SNTNotificationQueue use a ring buffer for storing queued messages instead of an array. This means that the most recent notifications will be displayed when a user logs in and the oldest ones will be dropped.


Source/santad/SNTNotificationQueue.mm
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+
+    [self flushQueueLocked];
+  });
+}
+
+/// For each pending notification, call the reply block if set then clear the
+/// reply so it won't be called again when the notification is eventually sent.
+- (void)clearAllPendingRepliesLocked {
```
russellhancox, line 84: The "locked" in this method name (and flushQueueLocked) is a little confusing as there isn't a lock held

pmarkowsky: We have a few of those in the code base, by convention they assume that you've already locked the resource.

- [ OpenFileHandleLocked();](https://github.com/northpolesec/santa/blob/3a5df9e0ae69b973bb46f51b9a1d0e146d7784ce/Source/santad/Logs/EndpointSecurity/Writers/File.h#L50)
- [FlushLocked()](https://github.com/northpolesec/santa/blob/3a5df9e0ae69b973bb46f51b9a1d0e146d7784ce/Source/santad/Logs/EndpointSecurity/Writers/File.h#L52)
- [ExportLocked()](https://github.com/northpolesec/santa/blob/3a5df9e0ae69b973bb46f51b9a1d0e146d7784ce/Source/santad/Metrics.mm#L330)
- [TryResetLocked](https://github.com/northpolesec/santa/blob/3a5df9e0ae69b973bb46f51b9a1d0e146d7784ce/Source/santad/EventProviders/RateLimiter.mm#L56)
- [ShouldRateLimitLocked](https://github.com/northpolesec/santa/blob/3a5df9e0ae69b973bb46f51b9a1d0e146d7784ce/Source/santad/EventProviders/RateLimiter.mm#L44)
- and so on.

russellhancox: Discussed offilne: none of these cases with the 'Locked' suffix are using locks explicitly but are only being done on serialized queues, so safety-wise it's OK. Switching the naming to be clearer might be beneficial but is not pressing.

russellhancox (approved):

Source/santad/SNTNotificationQueue.mm
mlw: Will followup in a separate PR with a consistent naming approach.

PR #192, opened 2024-12-30
Add serializations for clone and copyfile

mlw:
The bulk of the PR is new test golden files.

The main thing to call out is that the string serialization doesn't include "mode", "flags", and "previously existed" like is done for proto. This is largely because encoding these types is more cumbersome. We often don't add these for the string serializations, but could add them (maybe as hex and true/false strings?) if anyone feels strongly.


pmarkowsky (approved):
LGTM.

PR #194, opened 2024-12-31
Add a parquet writer using Rust and CXX for FFI

the80srobot:
This adds a fairly MVP writer for Parquet files implemented in Rust and usable from C++. An actual Santa logger is just a stub, because this diff is already large. (And yet, I managed to shave off 1000 lines compared to my original PR from 2023.)

This XL PR could be split into one that adds Rust build support and one that adds the Parquet writer, but both of those things informed each other, so it actually seems easier to review them together. Sorry for the size, I won't make it a habit.

Some notes:

* The patch for cxx is unfortunate, however the WORKSPACE -> MODULE migration seemingly broke everything and I couldn't find a combination of rules_rust and cxx that was both stable and easy to use with MODULE.bazel. The good news is that rules_rust 0.57.0 *should* fix everything, allowing me to update cxx and drop the patch. (Probably in a few weeks.)
* The writer is fairly low-level. Pete's experiments AFAIK are with the more high-level arrow API. We should chat about trade-offs. I don't have a strong opinion on which is better. This PR does not lock us into one approach.

Test plan:

There is an e2e test shell script, which requires Docker.


pmarkowsky:
@russellhancox @tburgin  @mlw @the80srobot I believe we agreed to shelve this in favor of a future one no?

pmarkowsky:
Talked with @the80srobot. Closing this in favor of a future PR.

PR #198, opened 2025-01-08
Add licenses back to santa/BUILD

kubalaguna:
This was deleted in https://github.com/northpolesec/santa/commit/3c728b6, hopefully by accident


pmarkowsky:
Yup the license function was removed from the BUILD file on accident on our part. This [just notifies you that the license](https://docs.bazel.build/versions/0.24.0/be/functions.html#licenses) requires an original copyright notice or advertising clause and that external distributions require the notice to be included.

All of the code is covered by the license file in the repo.

tburgin (approved):

PR #199, opened 2025-01-08
santactl status: fix hang when sync server is unresponsive

tburgin:
Because install.sh (and postinstall) now use output from `santactl status`, this CL also fixes the intermittent hanging we have noticed during `bazel :reload`.

https://github.com/northpolesec/santa/blob/ce65385ca5234e071bdefac96cbf9ef46ff69605/Conf/install.sh#L40


russellhancox (approved):

pmarkowsky (approved):
LGTM.

PR #201, opened 2025-01-10
Ensure ES doesn't cacheincorrect hold and ask responses

mlw:
This change addresses the following issues:

1. Ensures that responses to ES AUTH EXECS are no cached during hold-and-ask
2. Blocks duplicate executions of a binary while another instance is pending authorization
3. Changes `Message` lifetimes to ensure proper ES message retains/releases


pmarkowsky (approved):
LGTM. 

Only thought is, are there any other checks or safety rails we need to consider since we've moved from the const reference in the endpoint security clients.

pmarkowsky (approved):

PR #203, opened 2025-01-10
Add an optional step to add a TeamID rule for NPS Santa

pmarkowsky:
This PR adds an optional step of adding a `Team ID` rule for NPS to Google Santa in the migration guide.

### DIFF
--- a/docs/deployment/migration.md
+++ b/docs/deployment/migration.md
@@ -24,7 +24,7 @@ straight to [Getting Started](getting-started.md).
 ## Migration Steps
 
 ### 1. Configure System Extensions
-- (Optional) Add a Team ID Rule for North Pole Security's Team ID (ZMCG7MLDV9) to Google Santa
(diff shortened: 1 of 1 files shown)


docs/deployment/migration.md
```diff
@@ -24,6 +24,8 @@ straight to [Getting Started](getting-started.md).
 ## Migration Steps
 
 ### 1. Configure System Extensions
+- (Optional) Add a Team ID Rule for North Pole Security's Team ID (ZMCG7MLDV9) to Google Santa
```
russellhancox, line 27: My only concern with this is that being an optional step, how does an admin following these steps decide whether they should or should not do it? Could we add another bullet outlining the reason to do it?

pmarkowsky: Makes sense. How about a sentence like the following:

`(Optional) Add a Team ID Rule for North Pole Security's Team ID (ZMCG7MLDV9) to Google Santa. This is to guarantee that complex MDM setups don't allow Google Santa to block NPS Santa`

russellhancox: Works for me

PR #205, opened 2025-01-14
slightly less blinding red

discentem:
The sidebar text is slightly hard to read with such a blinding red. This makes it a bit less blinding (I think). 

- I haven't tested this though. I'm happy to test if there are instructions somewhere for running the webpage on a local port

### DIFF
--- a/docs/_sass/custom/custom.scss
+++ b/docs/_sass/custom/custom.scss
@@ -14,7 +14,7 @@
 }
 
 .side-bar {
-  background-color: #FF0000; /* Pure red */
+  background-color: #c44549;
 }
 
 .side-bar,


russellhancox:
> instructions somewhere for running the webpage on a local port

In the docs folder:

```
$ bundle install
$ bundle exec jekyll serve
```

discentem:
> > instructions somewhere for running the webpage on a local port
> 
> In the docs folder:
> 
> ```
> $ bundle install
> $ bundle exec jekyll serve
> ```

Thanks, I was able to get it running and can see this probably too muted. I can try a few other options though @pmarkowsky noted that y'all may have larger color scheme changes coming soon so no worries if you can't merge some change here right now.

pmarkowsky:
First off really appreciate the PR. However we need to close this as we're gonna be doing some work here for branding in the near future.

PR #209, opened 2025-01-15
Clarify uniqueness properties of PID + pidversion

gnoack:
This is, among other places, documented in the ESMessage.h header, where it says that (pid, pidversion) "is not meant to be unique across reboots or across multiple systems".


pmarkowsky (approved):
LGTM.

mlw (approved):

PR #210, opened 2025-01-15
common: Add bootSessionUUID to SNTSystemInfo

russellhancox:
This fetches the stable boot session UUID from sysctl but it is currently unused.

I also updated the serialNumber and hardwareUUID properties to only fetch once instead of on every call, and removed some deprecated uses of kIOMasterPortDefault


Source/common/SNTSystemInfo.h
```diff
@@ -29,6 +29,14 @@
 ///
 + (NSString *)hardwareUUID;
 
+///
+///  The boot session UUID is generated by IOKit during early startup and is
+///  persistent until the team reboots.
```
pmarkowsky, line 34:
```suggestion
///  persistent until the system reboots.
```

Source/common/SNTSystemInfo.m
```diff
@@ -18,34 +18,47 @@
...
+  return uuid;
+}
 
++ (NSString *)bootSessionUUID {
+  static NSString *uuid;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    uuid_string_t bootSessionUUID = {};
+    size_t uuidLength = sizeof(bootSessionUUID);
+    sysctlbyname("kern.bootsessionuuid", bootSessionUUID, &uuidLength, NULL, 0);
```
mlw, line 59: Need to handle failure

mlw (changes requested):

Source/common/SNTSystemInfo.m
pmarkowsky, line 59:
```suggestion
    if (sysctlbyname("kern.bootsessionuuid", bootSessionUUID, &uuidLength, NULL, 0) != 0) {
       LOGE(@"Unable to get kern.bootsessionuuid sysctl");
    }
```

You'd need to add the import for SNTLogging to the SNTSystemInfo.m and update the BUILD file as well.

mlw (approved):

PR #212, opened 2025-01-16
Ensure that non-glob faa path patterns are watched

mlw:
Fixes an issue where configured non-glob watch paths (i.e. those without any glob magic characters) in FAA rules would not be monitored until after the file was created and Santa re-applied the config to see the new path (default 10m, defined by `FileAccessPolicyUpdateIntervalSec`).


pmarkowsky (approved):
LGTM.

russellhancox (approved):

PR #200, opened 2025-01-08
config: use CFPreferencesCopyAppValue and CFPreferencesAppValueIsForced

tburgin:

PR #214, opened 2025-01-17
Add back code to watch changes to our defaults domain

mlw:
Removed inadvertently here: https://github.com/northpolesec/santa/pull/200


pmarkowsky (approved):
LGTM

Source/common/SNTConfigurator.m
```diff
@@ -296,6 +298,9 @@ - (instancetype)initWithSyncStateFile:(NSString *)syncStateFilePath
     _syncStateFilePath = syncStateFilePath;
     _syncStateAccessAuthorizerBlock = syncStateAccessAuthorizer;
 
+    _defaults = [NSUserDefaults standardUserDefaults];
```
russellhancox, line 301: Might be worth adding a comment for our future selves that this is to keep KVO on changes but we're using CF* for reading.

PR #216, opened 2025-01-21
Enforce direct BUILD dependencies by adding layering check (1/2)

kubalaguna:
It ensures that code includes the headers it needs directly instead of transitively.


russellhancox:
Is there any reason not to apply this to all relevant BUILD files instead of just the santad one?

kubalaguna:
Ultimately I want to apply this to all BUILD files. I wanted to see if you would be willing to accept this change before doing the whole thing. We can do it in a single PR.

PR #217, opened 2025-01-21
gui: Fix open button appearing when it shouldn't.

russellhancox:
Also fix a bug in SNTTestGUI where it didn't catch this


mlw (approved):

pmarkowsky (approved):
LGTM.

PR #218, opened 2025-01-22
Config: Adjust perms to 0644 for sync-state.plist.

russellhancox:
Nothing in this plist is sensitive and preventing the non-root procs from reading it can cause config mismatches


pmarkowsky (approved):
LGTM.

Issue #227, opened 2025-01-27
Feature Request: Config key to remove ability to snooze notifications


4rm:
I would like to request a config key to choose if we want the "Prevent future notifications for this application for [...]" option to appear or not.

We have some public machines where guest users sign in to the same account, and I would like to prevent one user snoozing notifications for a future user.

Per @russellhancox in the MacAdmins Slack, we are currently running `rm /Users/X/Library/Preferences/com.northpolesec.santa.plist` as a logout script to get around this.

A company may want user-owned machines to not be able to snooze notifications either. In this case, a logout script would create an inconsistent experience, allowing the user to set a snooze time, but having that time reset on logout without the user knowing.

pmarkowsky:
@4rm this should be fixed in the next release 2025.2

4rm:
Amazing, thank you!

PR #228, opened 2025-01-27
gui: Add config to disable notification silences

russellhancox:
Also fixed a few bugs in SNTTestGUI that made it hard to test this


pmarkowsky:
Addresses #227

Source/gui/SNTNotificationManager.m
```diff
@@ -378,7 +381,7 @@ - (void)postBlockNotification:(SNTStoredEvent *)event
                                                   configState:configState
                                                         reply:replyBlock];
 
-  [self queueMessage:pendingMsg];
+  [self queueMessage:pendingMsg ignoringSilences:(configState.enableNotificationSilences != YES)];
```
mlw, line 384: nit: consider changing this param name to `enableSilences` or something, having the logic flip confused when reading the code.

Source/santad/Santad.mm
```diff
@@ -151,11 +151,13 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
 
     access_authorizer_client.fileAccessBlockCallback = ^(
         SNTFileAccessEvent *event, NSString *customMsg, NSString *customURL, NSString *customText) {
+      SNTConfigState *cs = [[SNTConfigState alloc] initWithConfig:[SNTConfigurator configurator]];
```
mlw, line 154: Note for future self: This is fine for now, but when we actually make config states valid for the life of processing an event, the config state will need to come as an argument to this block. Nothing to do about this today though.

mlw (dismissed):

Source/gui/SNTNotificationManager.m
russellhancox: Done

Source/santad/Santad.mm
russellhancox: I added a TODO

mlw (approved):

PR #229, opened 2025-01-27
project: Add simple Makefile

russellhancox:
The Makefile just wraps existing commands so we can run `make` to auto-format and build before doing a PR instead of running multiple commands.


pmarkowsky:
What about making rule or `sh_binary` target that does all of this or using something like [rules_multirun](https://github.com/keith/rules_multirun) 

I'll admit to being a little hesitant to add yet another build tool on top a build tool.

pmarkowsky (approved):
LGTM.

After talking about this. This seems like the easiest path for us to get what we want.

Issue #124, opened 2024-11-13
Process Based FAA


mlw:
In some scenarios what you actually want to control is what a process can and cannot open.  You want to know exactly which files a specific process interacted with e.g. every file that tar or zip touched. 

This issue is to extend FAA to cover those scenarios.

PR #236, opened 2025-02-02
Better type representation in FAA, standardize type alias naming

mlw:
The goal of this PR is to better represent data with types that match how the FAA configuration is understood. This largely means things like converting vectors to sets to ensure no duplicate data.

Some additional tests were added, but there are no new features with this change.

Finally, type alias names were standardized for better readability.

### DIFF
--- a/Source/santad/DataLayer/WatchItemPolicy.h
+++ b/Source/santad/DataLayer/WatchItemPolicy.h
@@ -33,12 +33,25 @@ namespace santa {
 // Forward declarations
 enum class WatchItemPathType;
 enum class WatchItemRuleType;
+template <typename T>
+struct SharedPtrValueHash;
+template <typename T>
+struct SharedPtrValueEqual;
+struct DataWatchItemPolicy;
+struct ProcessWatchItemPolicy;
 struct WatchItemProcess;
 
 // Helper type aliases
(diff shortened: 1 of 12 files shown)


Source/santad/DataLayer/WatchItemPolicy.h
```diff
@@ -33,12 +33,25 @@ namespace santa {
 // Forward declarations
 enum class WatchItemPathType;
 enum class WatchItemRuleType;
+template <typename T>
+struct SharedPtrValueHash;
+template <typename T>
+struct SharedPtrValueEqual;
```
russellhancox, line 39:
```suggestion
template <typename T> struct SharedPtrValueHash;
template <typename T> struct SharedPtrValueEqual;
```

russellhancox: If clang-format doesn't like this we either need to modify its config or find some other way to make this clearer as it's hard when scanning down to see that this is 2 declarations instead of 4.

mlw: Spoke offline. clang-format 19 looks to have what we want, but it's not currently supported in the github runners. Will reevaluate later. For now, regrouped things to hopefully be more clear.

mlw:
Part of: #124

PR #244, opened 2025-02-08
Proc FAA - hookup probe and proc policy iteration interfaces

mlw:
This PR sets the stage for the probe interface to do meaningful work by hooking it up to the exec authorizer client as well as adding the interfaces in WatchItems to support iterating policies.  The "block taking a block" design is to mitigate the need for the WatchItems class to be aware of EndpointSecurity, and keeps policy matching in the FAA clients.

Next PR will begin abstracting functions from the existing Data FAA client for reuse in the Proc FAA client.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
@@ -22,7 +22,7 @@ using santa::ProcessWatchItemPolicy;
 
 @interface SNTEndpointSecurityProcessFileAccessAuthorizer ()
 @property bool isSubscribed;
-@property (copy) IterateProcessPoliciesBlock iterateProcessPoliciesBlock;
+@property(copy) IterateProcessPoliciesBlock iterateProcessPoliciesBlock;
 @end
 
 @implementation SNTEndpointSecurityProcessFileAccessAuthorizer
--- a/Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
@@ -22,7 +22,7 @@ using santa::ProcessWatchItemPolicy;
 
 @interface SNTEndpointSecurityProcessFileAccessAuthorizer ()
 @property bool isSubscribed;
-@property IterateProcessPoliciesBlock iterateProcessPoliciesBlock;
+@property (copy) IterateProcessPoliciesBlock iterateProcessPoliciesBlock;
 @end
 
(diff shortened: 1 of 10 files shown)


pmarkowsky (approved):
LGTM.

Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
```diff
@@ -13,21 +13,29 @@
...
+#include "Source/santad/DataLayer/WatchItemPolicy.h"
+#include "Source/santad/EventProviders/SNTEndpointSecurityEventHandler.h"
 
+using santa::IterateProcessPoliciesBlock;
 using santa::Message;
+using santa::ProcessWatchItemPolicy;
 
 @interface SNTEndpointSecurityProcessFileAccessAuthorizer ()
 @property bool isSubscribed;
+@property IterateProcessPoliciesBlock iterateProcessPoliciesBlock;
```
russellhancox, line 25: Make this a `(copy)`, and as you're not using the setter in init, you'll need to manually copy there.

Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
```diff
@@ -56,6 +72,7 @@ - (void)enable {
 
   if (!self.isSubscribed) {
     if ([super subscribe:events]) {
+      LOGD(@"Proc FAA subscribed");
```
russellhancox, line 75: Still needed?

Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
```diff
@@ -67,10 +84,19 @@ - (void)enable {
 - (void)disable {
   if (self.isSubscribed) {
     if ([super unsubscribeAll]) {
+      LOGD(@"Proc FAA unsubscribed");
```
russellhancox, line 87: Same

Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
mlw: I like having these here for development, being debug it shouldn't matter too much and it's lightweight - it only prints when policies go from 0->1+ or 1+->0. Would like to keep them.

russellhancox: That's fine, was just checking they weren't leftover debugging while writing

russellhancox (dismissed):

russellhancox (approved):

PR #247, opened 2025-02-10
Add back the ability to block scripts by hash

mlw:
Several years ago when Santa was kext-based (prior to adopting EndpointSecurity), it had the ability to block scripts by hash under specific scenarios. This change adds back this functionality. The same limitations as before still apply:

1. Block only - Santa will allow all scripts by default unless a specific block rule exists
2. Block rules for scripts only support the SHA256 identifier since they are unsigned. (note: `BlockedPathRegex` configuration key may also be used)
3. As with before, Santa can only evaluate script execution when the script is directly executed (e.g., `./foo.sh`, not `/bin/sh ./foo.sh`).

### DIFF
--- a/Source/common/SNTCachedDecision.h
+++ b/Source/common/SNTCachedDecision.h
@@ -45,6 +45,7 @@
 @property BOOL entitlementsFiltered;
 @property uint32_t codesigningFlags;
 @property SNTSigningStatus signingStatus;
+@property BOOL scriptExecutionDenied;
 
 @property NSString *quarantineURL;
 
--- a/Source/common/SNTFileInfo.m
+++ b/Source/common/SNTFileInfo.m
@@ -76,6 +76,10 @@ extern NSString *const NSURLQuarantinePropertiesKey WEAK_IMPORT_ATTRIBUTE;
 }
 
 - (instancetype)initWithEndpointSecurityFile:(const es_file_t *)esFile error:(NSError **)error {
+  if (!esFile) {
+    return nil;
+  }
+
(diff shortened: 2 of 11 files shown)


russellhancox (approved):

mlw:
We will not be able to provide this feature at this time due to system API limitations. The EndpointSecurity framework has no way to distinguish in its cache layer whether or not some EXEC event has an associated `script` field set. This means that in Santa we would never be able to make use of the ES cache layer so that we would be able to ensure that we could hash scripts for future execs of interpreters.

It's also worth being clear about what "script blocking" means. On macOS, there isn't a great way to do this. Sure we can block by hash under certain scenarios (namely, someone executing the script directly, e.g. ./foo.sh). But outside of this, things become difficult, fast. When a script is passed as an argument to some interpreter (e.g. /bin/sh foo.sh), we then have to start parsing all exec args for things that look like paths, not to mention understanding what various program switches exist that can change how a path may be presented. We could also potentially monitor for interpreters making OPEN calls, but this gets unwieldy fast as well given the number of non-script opens that occur, and needing to know beforehand which binaries are interpreters. There's also then needing to handle people actively trying to bypass the control, such as using switches to execute commands without some path (e.g. /bin/bash -c), adding a newline to the end of a file to change the hash, or even just copy/pasting a script's contents into a terminal to execute it.

Given these limitations, we are not currently moving forward with drastic options, such as disabling the ES cache which would allow us to implement the basics of this feature at the cost of degraded system performance.

PR #259, opened 2025-02-13
Updated sync-protocol docs for northpole.dev

pmarkowsky:
This PR updates the sync protocol documentation to reflect the additions added in Santa 2025.1

### DIFF
--- a/docs/development/sync-protocol.md
+++ b/docs/development/sync-protocol.md
@@ -258,7 +258,7 @@ If the binary has entitlements, the entitlements are stored in the `Event` as an
 | Key | Required | Type | Meaning | Example Value |
 |---|---|---|---|---|
 | entitlementsFiltered | Y | boolean | Whether or not Santa filtered some of the reported entitlements based on its configuration. | true |
-| entitlements | Y | list of entitlements see below | See Entitlement |
+| entitlements | Y | List of entitlements see below | See Entitlement |
 
 ##### Entitlement
 
--- a/docs/development/sync-protocol.md
+++ b/docs/development/sync-protocol.md
@@ -253,11 +253,11 @@ sequenceDiagram
 
 #### Entitlement Info
 
-If the binary is signed and has entitlements, they are stored in the Event as an EntitlementInfo object that indicates if the entitlements were filtered followed by a list of entitlements.
+If the binary has entitlements, the entitlements are stored in the `Event` as an `EntitlementInfo` object, that indicates whether or not the set was filtered and includes the set of captured entitlements.
 
 | Key | Required | Type | Meaning | Example Value |
 |---|---|---|---|---|
-| entitlements_filtered | Y | boolean | Did Santa filter out some of the entitlements | true |
+| entitlementsFiltered | Y | boolean | Whether or not Santa filtered some of the reported entitlements based on its configuration. | true |
 | entitlements | Y | list of entitlements see below | See Entitlement |
 
 ##### Entitlement
--- a/docs/development/sync-protocol.md
+++ b/docs/development/sync-protocol.md
@@ -487,15 +487,15 @@ The request consists of the following JSON keys:
 |---|---|---|---|---|
 | rules_received    | YES | int | The number of rules the client received from all ruledownlaod requests. | 211 |
 | rules_processed      | YES | int | The number of rules that were processed from all ruledownload requests. | 212 |
-| machine_id | YES | The UUID of the machine that is sending this postflight. | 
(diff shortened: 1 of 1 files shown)


docs/development/sync-protocol.md
```diff
@@ -248,65 +251,285 @@ sequenceDiagram
...
-        "sha256": "7ae80b9ab38af0c63a9a81765f434d9a7cd8f720eb6037ef303de39d779bc258"
-      },
-      {
-        "cn": "Apple Worldwide Developer Relations Certification Authority",
-        "valid_until": 1897776000,
-        "org": "Apple Inc.",
-        "valid_from": 1582136027,
-        "ou": "G3",
-        "sha256": "dcf21878c77f4198e4b4614f03d696d89c66c66008d4244e1b99161aac91601f"
+  "events": [
```
mlw, line 277: I'm not sure it's useful to have a 250 line example here. We could probably still use Santa.app which has a few entitlements to get the same point across.

docs/development/sync-protocol.md
```diff
@@ -236,6 +236,9 @@ sequenceDiagram
 | signing_id | NO | string | Signing ID of the binary that was executed | "EQHXZ8M8AV:com.google.Chrome" |
 | team_id | NO | string | Team ID of the binary that was executed | "EQHXZ8M8AV" |
 | cdhash | NO | string | CDHash of the binary that was executed | "dbe8c39801f93e05fc7bc53a02af5b4d3cfc670a" |
+| entitlementInfo | NO | Entitlement Info Object (see below) | Entitlement information about the target executable | see below |
+| csFlags | NO | int32 | The raw codesigning flags set for the execution |
+| signingStatus | NO | The signing status of the executing binary | **Must be one of the examples** | "SIGNING_STATUS_UNSPECIFIED", "SIGNING_STATUS_UNSIGNED", "SIGNING_STATUS_ADHOC", "SIGNING_STATUS_DEVELOPMENT", "SIGNING_STATUS_PRODUCTION" |
```
mlw, line 241: Is this the right casing? I would've thought it was words separated by underscores. If not, we should probably fix to match the others.

docs/development/sync-protocol.md
```diff
@@ -248,65 +251,285 @@ sequenceDiagram
 | valid_from | YES | int | Unix timestamp of when the cert was issued |  1647447514 |
 | valid_until | YES | int | Unix timestamp of when the cert expires |  1678983513 |
 
+#### Entitlement Info
+
+If the binary is signed and has entitlements, they are stored in the Event as an EntitlementInfo object that indicates if the entitlements were filtered followed by a list of entitlements.
```
mlw, line 256: nit: Wording feels off. Maybe something like:
```
If the binary has entitlements, the entitlements are stored in the Event as an EntitlementInfo object, which indicates whether or not the set was filtered and includes the set of captured entitlements.
```

docs/development/sync-protocol.md
```diff
@@ -248,65 +251,285 @@ sequenceDiagram
...
+
+| Key | Required | Type | Meaning | Example Value |
+|---|---|---|---|---|
+| entitlements_filtered | Y | boolean | Did Santa filter out some of the entitlements | true |
+| entitlements | Y | list of entitlements see below | See Entitlement |
```
mlw, line 261: nit: Capitalization.

docs/development/sync-protocol.md
```diff
@@ -248,65 +251,285 @@ sequenceDiagram
 | valid_from | YES | int | Unix timestamp of when the cert was issued |  1647447514 |
 | valid_until | YES | int | Unix timestamp of when the cert expires |  1678983513 |
 
+#### Entitlement Info
+
+If the binary is signed and has entitlements, they are stored in the Event as an EntitlementInfo object that indicates if the entitlements were filtered followed by a list of entitlements.
+
+| Key | Required | Type | Meaning | Example Value |
+|---|---|---|---|---|
+| entitlements_filtered | Y | boolean | Did Santa filter out some of the entitlements | true |
```
mlw, line 260: nit: Wording feels off. Maybe something like:
```
Whether or not Santa filtered some of the reported entitlements based on its configuration.
```

mlw (changes requested):

docs/development/sync-protocol.md
pmarkowsky: I've swapped it for a smaller one.

docs/development/sync-protocol.md
pmarkowsky: It is the right casing per what Santa does today as of 2025.1

pmarkowsky: For 2025.2 we should fix it to be consistent.

mlw (approved):

PR #262, opened 2025-02-14
stats: Fix allocation of protobuf message.

russellhancox:
Also fixes an issue where `santactl status` hangs if syncservice isn't running

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.m
+++ b/Source/santactl/Commands/SNTCommandStatus.m
@@ -140,7 +140,7 @@ REGISTER_COMMAND_NAME(@"status")
     // so we run the request asynchronously with a semaphore timer; if we have
     // no response within 2s, give up and move on.
     dispatch_semaphore_t sema = dispatch_semaphore_create(0);
-    dispatch_async(dispatch_get_global_queue(0, 0), ^{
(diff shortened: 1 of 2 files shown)


Source/santactl/Commands/SNTCommandStatus.m
```diff
@@ -135,9 +135,18 @@ - (void)runWithArguments:(NSArray *)arguments {
...
   if ([[SNTConfigurator configurator] syncBaseURL]) {
-    [rop pushNotifications:^(BOOL response) {
-      pushNotifications = response;
-    }];
+    // The request to santad to discover whether push notifications are enabled
+    // makes a call to santasyncservice. If it's unavailable the call can hang
+    // so we run the request asynchronously with a semaphore timer; if we have
+    // no response within 2s, give up and move on.
+    dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+    dispatch_async(dispatch_get_global_queue(0, 0), ^{
```
mlw, line 143: nit: Would recommend setting `QOS_CLASS_DEFAULT` instead of `QOS_CLASS_UNSPECIFIED`

russellhancox: Done, except I went with `QOS_CLASS_USER_INITIATED` because I think that's arguably more correct for this case.

mlw (approved):

PR #265, opened 2025-02-17
Add Contains method to PrefixTree

mlw:
Adds a new `Contains` method to the `PrefixTree` which will return true if the tree contains any prefix or literal string that matches the input, otherwise false.


pmarkowsky (approved):
LGTM.

PR #269, opened 2025-02-19
Remove stat change detection metrics that are no longer necessary

mlw:
This removes the metrics collection that was largely added here:
* https://github.com/google/santa/pull/1348
* https://github.com/google/santa/pull/1354

### DIFF
--- a/Source/santad/SNTExecutionController.mm
+++ b/Source/santad/SNTExecutionController.mm
@@ -268,12 +268,11 @@ static NSString *const kPrinterProxyPostMonterey =
   // TODO(markowsky): Maybe add a metric here for how many large executables we're seeing.
   // if (binInfo.fileSize > SomeUpperLimit) ...
 
-  SNTCachedDecision *cd = [self.policyProcessor decisionForFileInfo:binInfo
-      targetProcess:targetProc
-      configState:configState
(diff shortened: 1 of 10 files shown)


Source/santad/SNTExecutionController.mm
```diff
@@ -268,12 +268,11 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
...
-      targetProcess:targetProc
-      configState:configState
-      preCodesignCheckCallback:^(void) {
-        esMsg.UpdateStatState(StatChangeStep::kCodesignValidation);
-      }
+  SNTCachedDecision *cd = [self.policyProcessor
+             decisionForFileInfo:binInfo
+                   targetProcess:targetProc
+                     configState:configState
+        preCodesignCheckCallback:nil
```
russellhancox, line 275: WDYT about removing this callback?

russellhancox (approved):

pmarkowsky (approved):
LGTM.

Source/santad/SNTExecutionController.mm
mlw: Oh whoops, that didn't make the commit, needs to be removed cause won't build otherwise. Fixed.

PR #281, opened 2025-02-27
Update config docs for stats key

mlw:


docs/deployment/configuration.md
```diff
@@ -92,6 +92,7 @@ also known as mobileconfig files, which are in an Apple-specific XML format.
 | EntitlementsPrefixFilter           | Array      | Array of strings of entitlement prefixes that should not be logged (for example: `com.apple.private`). No default. |
 | EntitlementsTeamIDFilter           | Array      | Array of TeamID strings. Entitlements from processes with a matching TeamID in the code signature will not be logged. Use the value `platform` to filter entitlements from platform binaries. No default. |
 | [StaticRules](#static-rules)       | Array      | Array of rule dictionaries. The rules defined in this key take precedence over any rules in the rules database. |
+| EnableStatsCollection              | Bool       | If true, Santa will periodically collect and send basic, non-identifying stats to the maintainers at North Pole Security to help better support Santa. See [Stats documentation](https://northpole.dev/deployment/stats.html) for complete details. Defaults to false. |
```
russellhancox, line 95: WDYT about documenting the stats org key too?

mlw: Thought about it. I decided not to since no one should be setting it outside of those we're working with directly, which is documented on the Stats page.

mlw: Added

PR #282, opened 2025-02-28
Add fallback methods for gathering mounted DMG path

mlw:
This fixes a long-standing issue where the DMG path is not always collected for disk telemetry events depending on how the disk was mounted.

### DIFF
--- a/Source/santad/Logs/EndpointSecurity/Serializers/Utilities.mm
+++ b/Source/santad/Logs/EndpointSecurity/Serializers/Utilities.mm
@@ -130,34 +130,33 @@ static NSDictionary *PropertiesForDevice(NSString *devPath) {
 NSString *DiskImageForDevice(NSString *devPath) {
   NSString *result;
 
-  // First try getting the image-path property of the parent path of
-  // the given device path. Since at least macOS 13, this mainly works
-  // for `hdiutl attach` only.
-  NSDictionary *properties = PropertiesForDevice([devPath stringByDeletingLastPathComponent]);
+  // First, lookup properties of the full given path then look for either the
+  // DiskImageURL or Virtual Interface Location Path keys. Since at least
+  // macOS 13, these keys work for both DiskImageMounter and `hdiutil attach`.
+  NSDictionary *properties = PropertiesForDevice(devPath);
 
-  if (properties[@"image-path"]) {
-    result = [[NSString alloc] initWithData:properties[@"image-path"]
-                                   encoding:NSUTF8StringEncoding];
+  if ([properties[@"DiskImageURL"] isKindOfClass:[NSString class]]) {
+    // Note: This is often a file URL. Check and convert to a path to have
+    // consistent output compared to the other properties.
+    result = properties[@"DiskImageURL"];
+    if ([result hasPrefix:@"file://"]) {
+      result = [[NSURL URLWithString:result] path];
+    }
+  } else if ([properties[@"Protocol Characteristics"] isKindOfClass:[NSDictionary class]] &&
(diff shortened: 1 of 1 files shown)


pmarkowsky (dismissed):
LGTM.

Source/santad/Logs/EndpointSecurity/Serializers/Utilities.mm
```diff
@@ -125,14 +124,44 @@ const mach_port_t GetDefaultIOKitCommsPort() {
...
   IOObjectRelease(device);
+  return properties;
+}
+
+NSString *DiskImageForDevice(NSString *devPath) {
+  NSString *result;
+
+  // First try getting the image-path property of the parent path of
+  // the given device path. Since at least macOS 13, this mainly works
+  // for `hdiutl attach` only.
```
russellhancox, line 135:
```suggestion
  // for `hdiutil attach` only.
```

Source/santad/Logs/EndpointSecurity/Serializers/Utilities.mm
```diff
@@ -125,14 +124,44 @@ const mach_port_t GetDefaultIOKitCommsPort() {
...
   NSDictionary *properties = CFBridgingRelease(device_properties);
   IOObjectRelease(device);
+  return properties;
+}
+
+NSString *DiskImageForDevice(NSString *devPath) {
+  NSString *result;
+
+  // First try getting the image-path property of the parent path of
+  // the given device path. Since at least macOS 13, this mainly works
```
russellhancox, line 134:
```suggestion
  // First try getting the image-path property of the parent device of
  // the given device path. Since at least macOS 13, this mainly works
```

Source/santad/Logs/EndpointSecurity/Serializers/Utilities.mm
```diff
@@ -125,14 +124,44 @@ const mach_port_t GetDefaultIOKitCommsPort() {
...
 
   if (properties[@"image-path"]) {
-    NSString *result = [[NSString alloc] initWithData:properties[@"image-path"]
-                                             encoding:NSUTF8StringEncoding];
-    return [result stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+    result = [[NSString alloc] initWithData:properties[@"image-path"]
+                                   encoding:NSUTF8StringEncoding];
   } else {
-    return nil;
+    // As a fallback, lookup properties of the full given path then look
```
russellhancox, line 142: Given that this works for both mount routes on all recent macOSes, it would be better to do _this_ method first and fallback to the old way.

mlw: My hesitation here is that while I think this is probably a true statement, I don't really know for sure... With that said, I could find and boot some old macOS images and see what things look like.

pmarkowsky: That's why I didn't flag this earlier as well. The change that Matt made guarantees that we at least have the same behavior we've always had. 

Without stronger testing / a better understood root cause I'd be reluctant to change this to the default.

mlw: So fwiw, I tested back to 13.6 and the new methods work the same way as they do in 15.3.1. Given we'll have some time to bake the changes before the next release, I went ahead and switched up the order. The nice thing about the new order is that we likely only ever do one fetch/copy of properties, where if we checked things the old way first we would often require two fetch/copies.

russellhancox (approved):

PR #288, opened 2025-03-05
Add update method to SantaCache to update value under lock

mlw:
The new `update` method operates similarly to `set` in the following ways:
1. A new element will be created if it doesn't already exist
2. If the element exists the value will be updated (or at least the update block called for the opportunity to be updated)
3. If the cache is at capacity, it will first be cleared prior to the addition of a new element.

This PR also fixes test-only leaks. SantaCacheTest.xctest now runs with no reported leaks under the `leaks(1)` command.

### DIFF
--- a/Source/common/SantaCache.h
+++ b/Source/common/SantaCache.h
@@ -325,7 +325,6 @@ class SantaCache {
 
     // Allocate a new entry, set the key and value, then put this new entry at
     // the head of this bucket's linked list.
-    // struct entry *new_entry = (struct entry *)malloc(sizeof(struct entry));
     struct entry *new_entry = new struct entry(std::move(key));
     if (update_block) {
       update_block(new_entry->value);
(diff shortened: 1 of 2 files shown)


Source/common/SantaCache.h
```diff
@@ -294,10 +325,13 @@ class SantaCache {
 
     // Allocate a new entry, set the key and value, then put this new entry at
     // the head of this bucket's linked list.
-    struct entry *new_entry = (struct entry *)malloc(sizeof(struct entry));
-    bzero(new_entry, sizeof(struct entry));
-    new_entry->key = key;
-    new_entry->value = value;
+    // struct entry *new_entry = (struct entry *)malloc(sizeof(struct entry));
```
russellhancox, line 328: Can delete?

russellhancox (dismissed):

PR #292, opened 2025-03-11
Add reads cache support to the FAA Policy Processor

mlw:
This PR adds reads cache to the FAA policy processor. Because moving the cache to this layer will require differentiating between Data and Proc FAA clients, a new `FAAPolicyProcessorProxy` class is introduced that is appropriately subclassed and hides the details of setting the proper FAAClientType enum so that callers cannot accidentally use the wrong value.


russellhancox:
This is clean, very nice.

PR #295, opened 2025-03-12
Add Operation ID / fingerprint hash to FileAccess telemetry events

mlw:
It is possible for a single operation to emit multiple FileAccess telemetry log messages if multiple rules are violated. This PR adds an Operation ID (e.g. "fingerprint") value that would allow consumers of telemetry to link all telemetry from a single operation.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -552,6 +552,11 @@ message FileAccess {
     POLICY_DECISION_ALLOWED_AUDIT_ONLY = 3;
   }
   optional PolicyDecision policy_decision = 6;
+
+  // Used to link a single operation emitting multiple FileAccess messages.
+  // This can happen, for example, when a single operation violates both Data
+  // and Process File Access Authorization rules
+  optional string fingerprint = 7;
 }
 
 // Session identifier for a graphical session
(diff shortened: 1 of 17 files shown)


Source/common/santa.proto
```diff
@@ -552,6 +552,11 @@ message FileAccess {
     POLICY_DECISION_ALLOWED_AUDIT_ONLY = 3;
   }
   optional PolicyDecision policy_decision = 6;
+
+  // Used to link a single operation emitting multiple FileAccess messages.
+  // This can happen, for example, when a single operation violates both Data
+  // and Process File Access Authorization rules.
+  optional string fingerprint = 7;
```
russellhancox, line 559: If you're still looking for alternative names for `fingerprint`, how about `operation_id`?

russellhancox (approved):

russellhancox:
Sorry this took a while to get to

Source/common/santa.proto
mlw: Ooo I think i like this name better, it's more self-describing. Will update.

russellhancox (approved):

PR #306, opened 2025-03-19
Add foreach and clear with block support to SantaCache

mlw:
Adds new methods to operate on SantaCache key/value pairs while under lock.

### DIFF
--- a/Source/common/SantaCache.h
+++ b/Source/common/SantaCache.h
@@ -147,6 +147,39 @@ class SantaCache {
   }
 
   /**
+    Iterate all key and value pairs in the cache.
+
+    @param foreach_block Called for all key and value pairs.
+  */
+  void foreach (std::function<void(KeyT &, ValueT &)> foreach_block) {
+    assert(foreach_block != nullptr);
+
+    // Prevent the cache from being cleared while iterating
+    lock(&clear_bucket_);
+
+    // Lock all buckets
+    for (uint32_t i = 0; i < bucket_count_; ++i) {
+      lock(&buckets_[i]);
+    }
+
+    // Operate on all k/v pairs
+    for (uint32_t i = 0; i < bucket_count_; ++i) {
+      struct entry *entry = (struct entry *)((uintptr_t)buckets_[i].head - 1);
+      while (entry != nullptr) {
+        foreach_block(entry->key, entry->value);
+        entry = entry->next;
+      }
+    }
+
+    // Unlock all buckets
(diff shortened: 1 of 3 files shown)


Source/common/SantaCacheTest.mm
```diff
@@ -455,4 +455,93 @@ - (void)testContains {
...
+  sut.set(4, 44);
+  sut.set(5, 55);
+
+  sut.set(6, 66);
+  sut.remove(6);
+
+  NSSet *wantKeys = [[NSSet alloc] initWithArray:@[ @(1), @(2), @(3), @(4), @(5) ]];
+  NSSet *wantVals = [[NSSet alloc] initWithArray:@[ @(11), @(22), @(33), @(44), @(55) ]];
+  __block NSMutableSet *gotKeys = [[NSMutableSet alloc] init];
+  __block NSMutableSet *gotVals = [[NSMutableSet alloc] init];
```
russellhancox, line 472: Why 2 pairs of sets instead of 2 dicts?

Source/common/SantaCache.h
```diff
@@ -146,6 +146,39 @@ class SantaCache {
...
+  /**
+    Iterate all key and value pairs in the cache.
+
+    @param foreach_block Called for all key and value pairs.
+  */
+  void foreach (std::function<void(KeyT &, ValueT &)> foreach_block) {
+    assert(foreach_block != nullptr);
+
+    // Prevent the cache from being cleared while iterating
+    lock(&clear_bucket_);
```
russellhancox, line 158: I'm not sure this is strictly necessary, though admittedly it probably doesn't hurt much.

The clear function locks every bucket before it clears and as you're locking all the buckets too, either you got a bucket lock and a clear will wait to grab it, or vice-versa. The clear bucket is to prevent back-to-back clears when the cache overflows.

mlw: Yeah, this is true. Will remove.

If multiple threads call both functions simultaneously - whichever method is first to lock the first bucket, that method will then complete it's entire task before the other method can start. This works since, as you mention, both methods lock all buckets and, importantly, both start from index 0 and lock sequentially. I wanted to ensure that a cache couldn't be "partially cleared", but that can't happen here.

Source/common/SantaCacheTest.mm
mlw: Heh, cause I didn't really think about it :) Fixed.

russellhancox (approved):

PR #309, opened 2025-03-20
Stop watching processes when all Proc FAA rules are removed

mlw:


Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
```diff
@@ -213,6 +213,12 @@ - (void)handleMessage:(Message &&)esMsg
   return interest;
 }
 
+- (void)stopWatching:(const std::pair<pid_t, int> &)pidPidver {
+  audit_token_t stubToken = santa::MakeStubAuditToken(pidPidver.first, pidPidver.second);
+  [self unmuteProcess:&stubToken];
```
pmarkowsky, line 218: nit: Might be worth calling out in a comment that unmuting is actually removing the process from being watched, since you're inverting the mute.

pmarkowsky (dismissed):
LGTM.

pmarkowsky (approved):
LGTM.

PR #311, opened 2025-03-20
Handle FAA policy changes affecting watched processes

mlw:
With this change, when the FAA config changes Santa will now re-evaluate processes that are currently being watched because of the old policies to ensure they should still be watched under the new policies.

This is done by simply clearing the rule cache and having policy lookup be re-triggered the next time some event from a watched process comes in. This also has the nice side effect that cache eviction is now handled the same way.

This also fixes tracking/cleanup issues that existed on FORK/EXEC events.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
@@ -121,28 +121,23 @@ using ProcessRuleCache = SantaCache<PidPidverPair, std::shared_ptr<ProcessWatchI
 
   switch (esMsg->event_type) {
     case ES_EVENT_TYPE_NOTIFY_EXEC: {
-      // If the exec was allowed, remove the pid/pidver of the process that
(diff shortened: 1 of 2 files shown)


Source/santad/EventProviders/SNTEndpointSecurityProcessFileAccessAuthorizer.mm
```diff
@@ -121,28 +121,21 @@ - (void)handleMessage:(Message &&)esMsg
...
-      if (esMsg->action.notify.result.auth == ES_AUTH_RESULT_ALLOW) {
-        _procRuleCache->remove(PidPidversion(esMsg->process->audit_token));
-      } else {
-        _procRuleCache->remove(PidPidversion(esMsg->event.exec.target->audit_token));
-      }
-
+      // On EXEC, the previous process was replaced with the new one. The old
+      // process must be cleaned up since there will be no EXIT event for it.
+      // There will be a corresponding EXIT for the newly executed process
+      // regardless of whether or not it was allowed or denied.
```
russellhancox, line 127: This comment is a little confusing to me. IIUC the process wasn't replaced, just the pid version was incremented.

PR #317, opened 2025-03-24
Add proto schema definition for BTM LaunchItems

mlw:


Source/common/santa.proto
```diff
@@ -948,6 +948,72 @@ message GatekeeperOverride {
...
+  optional ProcessInfoLight instigator = 1;
+
+  // Whether or not the launch item is being added or removed
+  enum Action {
+    ACTION_UNKNOWN = 0;
+    ACTION_ADD = 1;
+    ACTION_REMOVE = 2;
+  };
+
+  Action action = 2;
```
russellhancox, line 965:
```suggestion
  Action action = 2;
```

Source/common/santa.proto
```diff
@@ -948,6 +948,72 @@ message GatekeeperOverride {
...
+  enum ItemType {
+    ITEM_TYPE_UNKNOWN = 0;
+    ITEM_TYPE_USER_ITEM = 1;
+    ITEM_TYPE_APP = 2;
+    ITEM_TYPE_LOGIN_ITEM = 3;
+    ITEM_TYPE_AGENT = 4;
+    ITEM_TYPE_DAEMON = 5;
+  }
+
+  ItemType item_type = 7;
```
russellhancox, line 996:
```suggestion
  ItemType item_type = 7;
```

Source/common/santa.proto
```diff
@@ -948,6 +948,72 @@ message GatekeeperOverride {
...
+
+  // Whether or not the launch item is a legacy plist
```
russellhancox, line 998: Can we expand on what legacy means? Do we know?

mlw: This seems to correlate with items backed by plists on disk vs SMAppService. But i'm not certain. Will leave largely undocumented for now since this is all the ES docs say.

russellhancox (approved):

PR #319, opened 2025-03-25
Add indirection layer between Configurator and CFPreferences API

kubalaguna:
While working on a new config option, I couldn't find a way to unit-test if it would be propagated correctly from MDM because `SNTConfigurator` is tightly coupled to system APIs. An additional layer between `SNTConfigurator` and the `CFPreferences` API makes it possible to inject MDM configuration in unit tests.

Added `MetricExportTimeout` config unit test as an example.

### DIFF
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -318,6 +318,7 @@ objc_library(
     ],
     deps = [
         ":SNTCommonEnums",
+        ":SNTMdmConfigSource",
         ":SNTRule",
         ":SNTStrengthify",
         ":SNTSystemInfo",
@@ -609,10 +610,12 @@ santa_unit_test(
 
 santa_unit_test(
     name = "SNTConfiguratorTest",
+    size = "small",
     srcs = ["SNTConfiguratorTest.m"],
     deps = [
         ":SNTCommonEnums",
         ":SNTConfigurator",
+        ":SNTMdmConfigSource",
         "@OCMock",
     ],
 )
@@ -744,6 +747,13 @@ objc_library(
     hdrs = ["MOLAuthenticatingURLSession.h"],
     sdk_frameworks = ["Security"],
     deps = [":MOLCertificate"],
+)
+
+objc_library(
+    name = "SNTMdmConfigSource",
+    srcs = ["SNTMdmConfigSource.m"],
+    hdrs = ["SNTMdmConfigSource.h"],
+    sdk_frameworks = ["Foundation"],
 )
 
 santa_unit_test(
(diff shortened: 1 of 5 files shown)


mlw:
Could you provide a bit more info about your end goals here, it's not entirely clear from the description/code. Are you trying to be able to independently mock the results of `CFPreferencesAppValueIsForced` and `CFPreferencesCopyAppValue`? Why wouldn't mocking `forcedConfigValueForKey` be enough?

kubalaguna:
My end goal is to test whether `SNTConfigurator` correctly processes a config variable provided by MDM.

Mocking `forcedConfigValueForKey` directly is also an option, if you prefer. But I find it clearer from an OOP perspective to decouple the system dependency (MDM) from the application logic (`SNTConfigurator`), as this enables better test coverage.

We could also move `forcedConfigValueForKey` entirely to `SNTMdmConfigSource`, but I'd like to keep this class super simple since it's not tested.

mlw:
I would also recommend against mocking `forcedConfigValueForKey` and prefer directly mocking the getter as we do throughout the tests in the code base. Mocking `forcedConfigValueForKey` requires creating new configurator instances (or simulating changes) to change a value which can sometimes be cumbersome depending on the test and/or also introduces a bunch of unrelated code paths to a test.

kubalaguna:
Thanks for your perspectives guys. Would you consider SNTConfigurator well-tested, or do you have other ideas on how we can improve?

If you consider the getters simple enough that they don't require tests, then probably the current coverage is ok. But when I first went through the codebase, it seemed to me that there might be more logic involved than simple value retrieval.

>I would also recommend against mocking forcedConfigValueForKey and prefer directly mocking the getter as we do throughout the tests in the code base.

Yeah I agree that mocking `forcedConfigValueForKey` doesn't provide much value if the test is not a unit test for `SNTConfigurator`.

mlw:
There is typically little-to-no logic in getters. We should certainly test more complex getters where feasible, but that needs to be weighed against test complexity. It's true we've typically not spent time on unit tests within the configurator (the refactor to make the configurator even be testable only happened last year (https://github.com/google/santa/pull/1275/). But also worth noting that testing a getter would likely require mocking things within the getter method (and potentially `forcedConfigValueForKey`). The wrappers added here wouldn't be beneficial for testing individual getters - they only enable testing of `forcedConfigValueForKey` and I'm hesitant to add 70+ test lines for a 4 line method.

kubalaguna:
Sounds reasonable, closing

PR #179, opened 2024-12-17
Make SNTRule initWithDictionary case insensitive

pmarkowsky:
This PR makes SNTRule's initWithDictionary method case insensitive by normalizing the keys of the dictionary to be lowercase.

This avoids issues like typing `rule_Type` when adding static rules or when importing rules from a JSON file.

PR #339, opened 2025-04-03
general: Add SNTError, use where appropriate

russellhancox:
This is to unify all uses of NSError where we're generating errors for Santa-specific errors and to report more errors from some configuration-related methods.


Source/common/SNTError.h
```diff
@@ -0,0 +1,70 @@
...
+                            detail:(nonnull NSString *)detail;
+
+// Generate a new error with the provided code and message.
+// `msg` will populate the NSLocalizedErrorDescription key.
++ (nonnull NSError *)errorWithCode:(SNTErrorCode)code message:(nonnull NSString *)msg;
+
+// Generate a new error with the provided message.
+// `msg` will populate the NSLocalizedErrorDescription key.
+// The error code will be SNTErrorCodeUnknown.
++ (nonnull NSError *)errorWithMessage:(nonnull NSString *)msg;
```
mlw, line 68: It is very common throughout this PR to create a temporary string and pass into these methods. I think it'd be a big quality of life improvement to instead define these as:

```
// SNTError.h
+ (NSError *)errorWithCode:(NSInteger)code message:(NSString *)format, ... NS_FORMAT_FUNCTION(2, 3);

// SNTError.m
+ (NSError *)errorWithCode:(NSInteger)code message:(NSString *)format, ... {
    va_list args;
    va_start(args, format);
    NSString *formattedMessage = [[NSString alloc] initWithFormat:format arguments:args];
    va_end(args);
    
    // create error object...
}
```

mlw: Maybe a another quality of life improvement would be to remove the need to check `if (error) { ... }`. We could add another method or two like:

```
+ (void)populateError:(NSError**)error withCode:(NSInteger)code message:(NSString *)format, ... NS_FORMAT_FUNCTION(2, 3);
```

These two suggestions would change this common occurrence:
```
if (![policyString isKindOfClass:[NSString class]]) {
  if (error) {
    NSString *errStr = [NSString
        stringWithFormat:@"Rule received with missing/invalid policy '%@'", policyString];
    *error = [SNTError errorWithCode:SNTErrorCodeRuleMissingPolicy message:errStr];
  }
  return nil;
}
```
into:
```
  if (![policyString isKindOfClass:[NSString class]]) {
    [SNTError populateError:error withCode:SNTErrorCodeRuleMissingPolicy
                  message:@"Rule received with missing/invalid policy '%@'", policyString];
    return nil;
  }
```

Source/common/SNTRule.h
```diff
@@ -85,9 +86,11 @@
                               type:(SNTRuleType)type;
 
 ///
-///  Initialize with a dictionary received from a sync server.
+///  Initialize with a dictionary received from a static rule or santactl.
+///  If the passed dict cannot be parsed as a rule nil will be returned. Additionally, if the error
+///  parameter is a non-nil pointer then it will be populated with an appropriate error object.
 ///
-- (instancetype)initWithDictionarySlow:(NSDictionary *)dict;
+- (instancetype)initWithDictionary:(NSDictionary *)rawDict error:(NSError **)error;
```
mlw, line 93: I'm generally against this... This method, while not currently used in a hot path, used to be. Before it accidentally goes back into a hot path, I feel it would be better for the author and reviewer of some commit to know what it is they're calling.

mlw (changes requested):

Source/common/SNTRule.h
russellhancox: > This method, while not currently used in a hot path, used to be

I don't recall this method ever being in a hot path. `initWithDictionary` was originally added to support static rules and was a lift-and-shift from SNTSyncRuleDownload - the two places it was called were in RuleDownload and in Configurator when re-caching static rules, neither of which are particularly hot. RuleDownload stopped using this method when we moved to using protobuf for syncing, leaving just StaticRules as the user of this. The `slow` was added to the name in #179 when Pete made this more forgiving of case and while it _is_ slower than initializing directly I don't think having slow in the name is really buying us very much.

mlw: This is what I was referring to by "hot path" - a clean sync where 10's of thousands of rules would need to be normalized: https://github.com/google/santa/blob/2024.5/Source/santasyncservice/SNTSyncRuleDownload.m#L107

This older code would've definitely spiked memory and CPU churning thru all these dictionaries and string comparisons if we were normalizing back then.

I'm a big fan of the API names signaling potentially dangerous spots like this. Is the push back on the name "slow"? We could rename to something like `initWithNonNormalizedDictionary` to better indicate what will happen. Or is the pushback that authors/reviewers wouldn't notice the name?

Source/common/SNTError.h
russellhancox: Done. I've replaced all of the original `errorWith` methods to be `populateError:withX` methods as there were very few cases where we weren't populating a passed error and it's easy enough to re-use in those few cases without having 2x the methods in this interface.

I've also added formatting versions to all of the methods _except_ `populateError:withCode:message:detail` because there's no (reasonably usable) way to have 2 format strings in a single method.

Source/common/SNTRule.h
russellhancox: The change to make this normalizing happened long after it was no longer used in RuleDownload.

> I'm a big fan of the API names signaling potentially dangerous spots like this. Is the push back on the name "slow"? We could rename to something like initWithNonNormalizedDictionary to better indicate what will happen. Or is the pushback that authors/reviewers wouldn't notice the name?

Bit of both - having a method with the name "slow" in it implies it's either doing something horrendous (which this isn't, really) or there is a faster version that can be used in most cases , and I don't think calling it `initWithDictionarySlow:withError` or `initWithNonNormalizedDictionary:` is going to prevent much. I've added a comment above the method, which probably also won't do much but is at least clear and doesn't pollute the rest of the codebase.

Source/common/SNTError.h
```diff
@@ -0,0 +1,85 @@
...
+// `msg` will populate the NSLocalizedErrorDescription key.
++ (void)populateError:(NSError *_Nullable *_Nullable)error
+             withCode:(SNTErrorCode)code
+              message:(nonnull NSString *)msg;
+
+// Generate a new error with the provided code and format-string message, and populate into the
+// provided NSError; `msg` will populate the NSLocalizedErrorDescription key.
++ (void)populateError:(NSError *_Nullable *_Nullable)error
+             withCode:(SNTErrorCode)code
+               format:(nonnull NSString *)format, ... NS_FORMAT_FUNCTION(3, 4);
```
mlw, line 72: Why have both?

russellhancox: As discussed, I was avoiding the `format string is not a string literal (potentially insecure)` error but we can use a simple format string instead to avoid the extra method, so I've done that.

PR #340, opened 2025-04-03
santactl: Add 'doctor' subcommand

russellhancox:


Source/santactl/Commands/SNTCommandDoctor.m
```diff
@@ -0,0 +1,133 @@
...
+    proc_name(pid, name, sizeof(name));
+
+    NSString *nsName = @(name);
+    if ([nsName isEqualToString:@"Santa"]) foundSanta = YES;
+    if ([nsName isEqualToString:@"com.northpolesec.santa.daemon"]) foundSantad = YES;
+    if ([nsName isEqualToString:@"santasyncservice"]) foundSantaSyncService = YES;
+  }
+
+  if (!foundSanta) {
+    print(@"[-] Santa GUI process doesn't seem to be running");
```
gnoack, line 101: Regarding the output format: It would be a useful feature if such diagnostics would be machine readable -- then it would become possible to feed them to other tools that gather overall machine health information.

(It does not necessarily have to be JSON, I think a simple textual format is also fine to parse, but if it is a textual format, could we specify what the format is, so that we can rely on it to stay stable?)

gnoack (commented):
More high level remark:
Are some of these failure cases possible to remediate, e.g. by invoking a sync, restarting or reinstalling Santa ("turning it off and on again")?

If remediation is feasible, an outside tool would need to tell apart which remediation would be the appropriate one to use. - Does the output format give enough information to classify the failures so that we can pick the right one?

(Some of the configuration error messages look pretty free form, but I imagine that these are at best remediable by doing a sync... so it's probably not worth telling them apart in detail in a machine-readable form.)

tburgin (commented):
Overall this LGTM

Source/santactl/Commands/SNTCommandDoctor.m
russellhancox: I mean this format (`[-]` to mean "I found a bad thing", `[+]` to mean "all is well") is going to remain stable, as it's how we're printing in green/red. But the errors themselves are somewhat free-form; I don't see us changing them regularly except to possibly make the wording clearer.

russellhancox:
> Are some of these failure cases possible to remediate, e.g. by invoking a sync, restarting or reinstalling Santa ("turning it off and on again")?

Everything can be fixed, but none of the things this command finds are automatically remediable or we would do that.

At some point we'll probably add suggestions of what to do for various errors but as I don't want to tie remediation steps to versions of Santa that's likely to involve links to docs.

> If remediation is feasible, an outside tool would need to tell apart which remediation would be the appropriate one to use. - Does the output format give enough information to classify the failures so that we can pick the right one?

No, this command is really intended for users to run when contacting support for help because any remediation would need to involve the user anyway.

Source/santactl/Commands/SNTCommandDoctor.m
```diff
@@ -0,0 +1,133 @@
...
+
+- (BOOL)validateProcesses {
+  print(@"=> Validating processes...");
+
+  BOOL foundSanta = NO, foundSantad = NO, foundSantaSyncService = NO;
+
+  // Get the number of procs and then allocate a buffer large enough to receive
+  // data about all of them.
+  int n = proc_listpids(PROC_ALL_PIDS, 0, NULL, 0);
+  int *buffer = (int *)malloc(sizeof(int) * n);
```
mlw, line 81: The number returned by `proc_listpids` is already multiplied by `sizeof(int)`:  https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/kern/proc_info.c#L375-L379

(looks like you account for this after the second call in the for loop conditional).

As a tangent, likely out of scope for this change, but we do this `proc_listpids` dance already in the proc tree. Would be kinda nice to have some shared interface https://github.com/google/santa/blob/main/Source/santad/ProcessTree/process_tree_macos.mm#L140

mlw: nit: I know this is a short live process that is about to exit, but would be nice to free the buffer,

mlw (changes requested):

Source/common/SNTConfigurator.m
```diff
@@ -1504,4 +1504,60 @@ - (void)cacheStaticRules {
...
+        // Check that the key is known to us.
+        id type = self.forcedConfigKeyTypes[key];
+        if (!type) {
+          [errors addObject:[NSString stringWithFormat:@"The key %@ is not recognized", key]];
+          return;
+        }
+
+        // Check that the type of the value matches the expected type.
+        id value = CFBridgingRelease(CFPreferencesCopyAppValue(keyRef, kMobileConfigDomain));
+        if (![value isKindOfClass:type]) {
```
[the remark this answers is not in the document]
russellhancox: Good catch

Source/common/SystemResources.mm
```diff
@@ -77,3 +78,29 @@ uint64_t AddNanosecondsToMachTime(uint64_t ns, uint64_t machTime) {
...
+    LOGW(@"Failed to retrieve list of pids");
+    return std::nullopt;
+  }
+
+  // Re-size the vector down to the actual number of pids received.
+  n_procs /= sizeof(pid_t);
+  pids.resize(n_procs);
+
+  return std::make_optional(pids);
+}
```
mlw: nit: newline

mlw (changes requested):

mlw (dismissed):

mlw (approved):

PR #341, opened 2025-04-04
santad: use santa cache for event upload backoff

tburgin:
This CL replaces the NSCache used for event upload backoff with SantaCache.

On full cache boundaries, NSCache, has exhibited some odd behavior. Namely, it is possible to get into a situation where a hash is successfully added to the cache, but is immediately cleared. Once in this state, that hash is not able to be successfully retrieved. Even after repeatedly being successfully added to the cache.

If an execution is being blocked in a tight loop like `while true; do yes; done`, this could result in an event being uploaded every [1500ms](https://github.com/northpolesec/santa/blob/82b09a99f3440a9eb6456a859a32905c6e2a5f90/Source/santad/EventProviders/AuthResultCache.h#L58).

Switching to SantaCache gives us better guarantees around full cache boundaries. We know the cache will be cleared when overfilled, and we know we can start filling it back up again. Once a usable LRU is introduced to Santa's codebase, we could go that route, but it is not pressing.

Also this CL adds some tests for SNTSyncdQueue.

### DIFF
--- a/Source/santad/SNTSyncdQueue.mm
+++ b/Source/santad/SNTSyncdQueue.mm
@@ -15,23 +15,27 @@
 
 #import "Source/santad/SNTSyncdQueue.h"
 
+#include <memory>
+
 #import "Source/common/MOLXPCConnection.h"
 #import "Source/common/SNTLogging.h"
 #import "Source/common/SNTStoredEvent.h"
 #import "Source/common/SNTXPCSyncServiceInterface.h"
+#include "Source/common/SantaCache.h"
+#include "Source/common/String.h"
 
 @interface SNTSyncdQueue ()
-@property NSCache<NSString *, NSDate *> *uploadBackoff;
 @property dispatch_queue_t syncdQueue;
 @end
 
-@implementation SNTSyncdQueue
+@implementation SNTSyncdQueue {
+  std::unique_ptr<SantaCache<std::string, NSDate *>> _uploadBackoff;
+}
 
 - (instancetype)init {
   self = [super init];
   if (self) {
-    _uploadBackoff = [[NSCache alloc] init];
-    _uploadBackoff.countLimit = 128;
+    _uploadBackoff = std::make_unique<SantaCache<std::string, NSDate *>>(128);
     _syncdQueue = dispatch_queue_create("com.northpolesec.syncd_queue", DISPATCH_QUEUE_SERIAL);
   }
   return self;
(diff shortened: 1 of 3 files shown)


Source/santad/SNTSyncdQueue.mm
```diff
@@ -15,23 +15,28 @@
...
+#include "Source/common/String.h"
 
 @interface SNTSyncdQueue ()
-@property NSCache<NSString *, NSDate *> *uploadBackoff;
 @property dispatch_queue_t syncdQueue;
 @end
 
-@implementation SNTSyncdQueue
+@implementation SNTSyncdQueue {
+  // TODO: Eventually replace with an LRU.
```
pmarkowsky, line 32: nit: we should open an issue to track this.

pmarkowsky (approved):
LGTM.  Minor nits. Feel free to ignore.

Source/santad/SNTSyncdQueue.mm
```diff
@@ -15,23 +15,28 @@
...
+  std::unique_ptr<SantaCache<std::string, NSDate *>> _uploadBackoff;
+}
 
 - (instancetype)init {
   self = [super init];
   if (self) {
-    _uploadBackoff = [[NSCache alloc] init];
-    _uploadBackoff.countLimit = 128;
+    _uploadBackoff = std::make_unique<SantaCache<std::string, NSDate *>>(128);
```
mlw, line 39: I think we should consider bumping to 256 here. Seems maybe high since this cache should only be filled with data from true blocks. But given the object sizes are small, and given the non-LRU semantics it might be worth it.

Source/santad/SNTSyncdQueue.mm
```diff
@@ -73,12 +79,13 @@ - (void)dispatchBlockOnSyncdQueue:(void (^)(void))block {
 
 // The event upload is skipped if an event has been initiated for it in the last 10 minutes.
 // The passed-in hash is fileBundleHash for a bundle event, or fileSHA256 for a normal event.
+// Returns YES if backoff is needed, NO otherwise.
```
mlw, line 82: Nice, this is def more readable.

mlw (dismissed):

Source/santad/SNTSyncdQueue.mm
tburgin: Good idea, done.

Source/santad/SNTSyncdQueue.mm
tburgin: SGTM

Source/santad/SNTSyncdQueue.mm
pmarkowsky: saw https://github.com/northpolesec/santa/issues/344 thanks

tburgin:
PTAL

pmarkowsky (approved):
LGTM.

mlw (approved):

PR #343, opened 2025-04-04
Send all postflight results together

mlw:
This PR changes the sync service to send the daemon all postflight results within a single XPC message. The primary goal is to create a single choke point for the daemon to know when a sync in complete. This is then used to defer vacuuming the events database until the end of a sync, drastically increasing event processing throughput and decreasing CPU load in the daemon - especially on larger event databases.

### DIFF
--- a/Source/common/SNTPostflightResult.h
+++ b/Source/common/SNTPostflightResult.h
@@ -1,0 +1,38 @@
+/// Copyright 2025 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


Source/common/SNTPostflightResult.h
```diff
@@ -0,0 +1,38 @@
...
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+#include "Source/common/SNTCommonEnums.h"
+
+@class SNTSyncState;
+
+@interface SNTPostflightResult : NSObject <NSSecureCoding>
```
russellhancox, line 21: I'm not a fan of this being so closely tied to "postflight"; this class is ultimately a way to update many settings at once

Source/common/SNTPostflightResult.h
```diff
@@ -0,0 +1,38 @@
...
+
+- (instancetype)initWithSyncState:(SNTSyncState *)syncState;
```
russellhancox, line 23: I have mixed feelings about this. While being able to init directly from the sync state makes the code in SNTSyncPostflight much simpler, it requires importing SNTSyncState from santasyncservice which is crossing boundaries - nothing in common should should ever import from outside of common, nothing in each component folder should import from any other (only themselves or common).

mlw:
Superseded by: #346

PR #346, opened 2025-04-04
Add config bundling

mlw:
This PR introduces the "config bundles" which are a way to generically refer to some config state that can be passed between components. This is then used to pass settings between the sync service and main Santa daemon.

Vacuuming the events database has been removed. The reason is that we now vacuum on startup. Given it would be extremely rare for the database to grow large, this only has the impact of adding database contention unnecessarily. Additionally, constantly shrinking the database, even on the "happy path", results in a performance penalty of having to keep grow and shrink things.

### DIFF
--- a/Source/common/SNTConfigBundleTest.mm
+++ b/Source/common/SNTConfigBundleTest.mm
@@ -31,6 +31,8 @@
 @property NSNumber *enableAllEventUpload;
 @property NSNumber *disableUnknownEventUpload;
 @property NSString *overrideFileAccessAction;
+@property NSDate *fullSyncLastSuccess;
+@property NSDate *ruleSyncLastSuccess;
 @end
 
 @interface SNTConfigBundleTest : XCTestCase
@@ -40,7 +42,8 @@
 
 - (void)testGettersWithValues {
   __block XCTestExpectation *exp = [self expectationWithDescription:@"Result Blocks"];
-  exp.expectedFulfillmentCount = 11;
+  exp.expectedFulfillmentCount = 13;
+  NSDate *nowDate = [NSDate now];
 
   SNTConfigBundle *bundle = [[SNTConfigBundle alloc] init];
   bundle.clientMode = @(SNTClientModeLockdown);
@@ -54,6 +57,8 @@
   bundle.enableAllEventUpload = @(NO);
   bundle.disableUnknownEventUpload = @(YES);
   bundle.overrideFileAccessAction = @"disable";
+  bundle.fullSyncLastSuccess = nowDate;
+  bundle.ruleSyncLastSuccess = nowDate;
 
   [bundle clientMode:^(SNTClientMode val) {
     XCTAssertEqual(val, SNTClientModeLockdown);
@@ -110,7 +115,18 @@
     [exp fulfill];
   }];
 
-  [self waitForExpectationsWithTimeout:3.0 handler:NULL];
+  [bundle fullSyncLastSuccess:^(NSDate *val) {
+    XCTAssertEqualObjects(val, nowDate);
+    [exp fulfill];
+  }];
+
+  [bundle ruleSyncLastSuccess:^(NSDate *val) {
+    XCTAssertEqualObjects(val, nowDate);
+    [exp fulfill];
+  }];
+
+  // Low timeout because code above is synchronous
+  [self waitForExpectationsWithTimeout:0.1 handler:NULL];
 }
 
 - (void)testGettersWithoutValues {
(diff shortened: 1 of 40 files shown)


Source/common/SNTConfigBundleTest.mm
```diff
@@ -0,0 +1,181 @@
...
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import "Source/common/SNTConfigBundle.h"
+
+#include <Foundation/Foundation.h>
```
russellhancox, line 17:
```suggestion
#import <Foundation/Foundation.h>
```

Source/common/SNTXPCControlInterface.h
```diff
@@ -46,19 +47,9 @@ typedef NS_ENUM(NSInteger, SNTRuleAddSource) {
...
-- (void)setAllowedPathRegex:(NSString *)pattern reply:(void (^)(void))reply;
-- (void)setBlockedPathRegex:(NSString *)pattern reply:(void (^)(void))reply;
-- (void)setBlockUSBMount:(BOOL)enabled reply:(void (^)(void))reply;
-- (void)setRemountUSBMode:(NSArray *)remountUSBMode reply:(void (^)(void))reply;
-- (void)setEnableBundles:(BOOL)bundlesEnabled reply:(void (^)(void))reply;
-- (void)setEnableTransitiveRules:(BOOL)enabled reply:(void (^)(void))reply;
-- (void)setEnableAllEventUpload:(BOOL)enabled reply:(void (^)(void))reply;
-- (void)setDisableUnknownEventUpload:(BOOL)enabled reply:(void (^)(void))reply;
-- (void)setOverrideFileAccessAction:(NSString *)action reply:(void (^)(void))reply;
+- (void)postflightResult:(SNTConfigBundle *)result reply:(void (^)(void))reply;
```
russellhancox, line 52: Still not a fan of this name.

Maybe `updateSyncSettings`?

Source/santad/SNTDaemonControlController.mm
```diff
@@ -257,78 +247,90 @@ - (void)setSyncTypeRequired:(SNTSyncType)syncType reply:(void (^)(void))reply {
...
+    [configurator setDisableUnknownEventUpload:val];
+  }];
+
+  [result overrideFileAccessAction:^(NSString *val) {
+    [configurator setSyncServerOverrideFileAccessAction:val];
+  }];
+
+  [configurator setFullSyncLastSuccess:[NSDate now]];
+
+  // Vacuum the event databases when postflight is complete since it has just been drained.
```
russellhancox, line 329:
```suggestion
  // Vacuum the event database when postflight is complete since it has just been drained.
```

Source/santasyncservice/SNTPostflightConfigBundleTest.mm
```diff
@@ -0,0 +1,68 @@
...
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import "Source/santasyncservice/SNTPostflightConfigBundle.h"
+
+#include <Foundation/Foundation.h>
```
russellhancox, line 17:
```suggestion
#import <Foundation/Foundation.h>
```

Source/santasyncservice/SNTSyncPostflight.mm
```diff
@@ -19,6 +19,7 @@
 #import "Source/common/SNTSyncConstants.h"
 #import "Source/common/SNTXPCControlInterface.h"
 #import "Source/common/String.h"
+#include "Source/santasyncservice/SNTPostflightConfigBundle.h"
```
russellhancox, line 22:
```suggestion
#import "Source/santasyncservice/SNTPostflightConfigBundle.h"
```

Source/common/SNTXPCControlInterface.h
mlw: The intention here was to be less generic, and have an explicit callsite for the daemon to perform any actions once postflight was complete. So I'd prefer a name that made this clear. Maybe `postflightComplete`?

Source/common/SNTConfigBundleTest.mm
mlw: ugh, I always mess this up. `import` was such a bad idea that doesn't even have a benefit anymore in modern compilers. Was just a lazy, language-specific "solution" to a problem that has now been more generically solved. End of mini rant. will change.

mlw: Mass changed throughout the project.

russellhancox (approved):

PR #357, opened 2025-04-13
Fix crashing tests due to incorrect releasing behaviour

dzonder:
https://clang.llvm.org/extra/clang-tidy/checks/objc/nsinvocation-argument-lifetime.html


pmarkowsky (approved):
LGTM.

PR #360, opened 2025-04-15
Expand sync service connection state to handle more config keys

mlw:
This changes Santa to ensure the sync service is running if either a SyncBaseURL is set, or stats collection is enabled. The desired state of the sync service is reassessed whenever configuration is changed to determine if the service should be started, bounced, or stopped.

### DIFF
--- a/Source/santad/SNTSyncdQueue.mm
+++ b/Source/santad/SNTSyncdQueue.mm
@@ -74,34 +74,44 @@
       STRONGIFY(self);
       SNTConfigurator *configurator = [SNTConfigurator configurator];
 
+      NSURL *newSyncBaseURL = [configurator syncBaseURL];
+      BOOL statsCollectionEnabled = [configurator enableStatsCollection];
+
+      // Don't allow an empty string for the sync base URL
+      if (newSyncBaseURL && newSyncBaseURL.absoluteString.length == 0) {
+        newSyncBaseURL = nil;
+      }
+
       // If the SyncBaseURL was added or changed, and a connection already
       // exists, it must be bounced.
-      if (self.previousSyncBaseURL &&
-          ![self.previousSyncBaseURL isEqual:[configurator syncBaseURL]] &&
-          self.syncServiceConnected) {
+      if (self.previousSyncBaseURL.absoluteString.length &&
+          ![self.previousSyncBaseURL isEqual:newSyncBaseURL] && self.syncServiceConnected) {
         [self tearDownSyncServiceConnectionSerialized];
         // Return early. When the sync service spins down, it will trigger the
         // invalidation handler which will reassess the connection state.
         return;
       }
 
-      self.previousSyncBaseURL = [configurator syncBaseURL];
+      self.previousSyncBaseURL = newSyncBaseURL;
 
-      // If syncBaseURL or enableStatsCollection is set, start the sync service
-      if ([configurator syncBaseURL] || [configurator enableStatsCollection]) {
+      // If newSyncBaseURL or statsCollectionEnabled is set, start the sync service
+      if (newSyncBaseURL || statsCollectionEnabled) {
         if (!self.syncServiceConnected) {
-          [NSObject cancelPreviousPerformRequestsWithTarget:[SNTConfigurator configurator]
-                                                   selector:@selector(clearSyncState)
-                                                     object:nil];
-
           [self establishSyncServiceConnectionSerialized];
         }
-      } else if (self.syncServiceConnected) {
-        // Note: Only teardown the connection if we're connected. This helps
-        // prevent the condition where we would end-up reestablishing the
-        // connection when the invalidationHandler fired and a call to
-        // spindown would be performed.
-        [self tearDownSyncServiceConnectionSerialized];
+
+        // Ensure any pending requests to clear sync state are cancelled.
+        [NSObject cancelPreviousPerformRequestsWithTarget:[SNTConfigurator configurator]
+                                                 selector:@selector(clearSyncState)
+                                                   object:nil];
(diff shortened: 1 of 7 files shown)


Source/santad/SNTSyncdQueue.mm
```diff
@@ -37,11 +42,109 @@ - (instancetype)init {
...
+
+      // If syncBaseURL or enableStatsCollection is set, start the sync service
+      if ([configurator syncBaseURL] || [configurator enableStatsCollection]) {
+        if (!self.syncServiceConnected) {
+          [NSObject cancelPreviousPerformRequestsWithTarget:[SNTConfigurator configurator]
+                                                   selector:@selector(clearSyncState)
+                                                     object:nil];
+
+          [self establishSyncServiceConnectionSerialized];
+        }
```
pmarkowsky, line 98: There's a potential TOCTOU bug here, since we're async in a thread. If you were to change the `syncBaseURL` config directly after the check on line 80.  You could assign a version of it different from the version checked on line 88. 

KVO should handle this and call back to push the timer out to the next second, but if you're in the middle of executing the block  the cancelation won't interrupt it per https://developer.apple.com/documentation/dispatch/1385604-dispatch_source_cancel.

While I think this is very unlikely to trigger in practice, if you get the `syncBaseURL` once from the configurator before checking this is at least consistent and goes away.

Source/santad/SNTSyncdQueue.mm
```diff
@@ -37,11 +42,109 @@ - (instancetype)init {
...
+                              100 * NSEC_PER_MSEC);
+
+    // WEAKIFY(self);
+    dispatch_source_set_event_handler(self.timer, ^{
+      STRONGIFY(self);
+      SNTConfigurator *configurator = [SNTConfigurator configurator];
+
+      // If the SyncBaseURL was added or changed, and a connection already
+      // exists, it must be bounced.
+      if (self.previousSyncBaseURL &&
```
pmarkowsky, line 79: I believe we're using `.length` to check if the string is valid elsewhere. in the codebase.

pmarkowsky (commented):
I think there might be an edge case in `reassessSyncServiceConnection`

Source/santad/SNTSyncdQueue.mm
mlw: Yeah, this is correct, there is a data race here. I think only one or two transition types would be affected, but only pulling the value once should fix it. What you say about scheduling is also true - if a second config changes while the previous one is being processed, the current timer will be "cancelled" (but won't stop the thread), and a second timer will be started to get the "next" state. This means most transitions that occur during processing would be fine and eventually consistent.

There is one small race still that this won't address - we'll need to pull both values from the configurator we care about (and `enableStatsCollection` queries two values behind the scenes, so 3 total), but this won't be atomic. So there could still be a discrepancy within a single processing window. But I think this is fine since there aren't any interdependencies between these config values and things will cleanup on the next processing round. (this is another good reason for idempotent config).

Source/santad/SNTSyncdQueue.mm
russellhancox: It won't return nil even if line 819 didn't exist, `URLWithString:` returns a valid object even for an empty string 🤦

mlw: I partially take this back - I'm going to ensure that the new SyncBaseURL cannot be empty, and if so, force it to nil (tho after the configurator fix, this shouldn't be possible). Then will keep the existing nil check as is because it makes more sense and is (imho) cleaner.

Source/santad/SNTSyncdQueue.mm
```diff
@@ -74,34 +74,44 @@ - (void)reassessSyncServiceConnection {
       STRONGIFY(self);
       SNTConfigurator *configurator = [SNTConfigurator configurator];
 
+      NSURL *newSyncBaseURL = [configurator syncBaseURL];
+      BOOL statsCollectionEnabled = [configurator enableStatsCollection];
+
+      // Don't allow an empty string for the sync base URL
+      if (newSyncBaseURL && newSyncBaseURL.absoluteString.length == 0) {
```
russellhancox, line 81:
```suggestion
      if (newSyncBaseURL) {
```

The 2nd condition is effectively impossible

mlw: It is as long as someone doesn't change the configurator in the future... If you're arguing to remove the length check, we'd just remove the whole if statement. I think I'd rather keep it in defensively, but agree it shouldn't be possible in the current codebase.

Source/santad/SNTSyncdQueue.mm
```diff
@@ -74,34 +74,44 @@ - (void)reassessSyncServiceConnection {
...
+        newSyncBaseURL = nil;
+      }
+
       // If the SyncBaseURL was added or changed, and a connection already
       // exists, it must be bounced.
-      if (self.previousSyncBaseURL &&
-          ![self.previousSyncBaseURL isEqual:[configurator syncBaseURL]] &&
-          self.syncServiceConnected) {
+      if (self.previousSyncBaseURL.absoluteString.length &&
```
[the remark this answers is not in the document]
mlw: Blah, missed this commit.

Source/santad/SNTSyncdQueue.mm
russellhancox: We typically do `.length` checks on strings because it catches both nil and 0-length on NSString, but because syncBaseURL is already checking for both of those conditions the extra check is unnecessary. If we're worried about someone changing configurator, we should add a comment in configurator to explain why the checks are there.

PR #363, opened 2025-04-15
santactl/rule: Allow deleting all rules

### LINKED ISSUE #362: Provide a better way to clean local rule database when not using a sync server
When not using a sync server, users currently have to import an "empty" rule set with the `--clean` flag set. However, we currently enforce that at least one rule exist when applying a new rule set and otherwise assume something went wrong. This is a failsafe mechanism in case something odd happened with the sync server. However now that we handle more local cases, we should provide a simpler path for users.

russellhancox:
Fixes #362

### DIFF
--- a/Source/santactl/Commands/SNTCommandRule.m
+++ b/Source/santactl/Commands/SNTCommandRule.m
@@ -78,9 +78,10 @@ REGISTER_COMMAND_NAME(@"rule")
 #endif
           @"    --message {message}: custom message to show when binary is blocked\n"
           @"    --comment {comment}: comment to attach to a new rule\n"
-          @"    --clean: when importing rules via JSON clear all non-transitive rules before "
-          @"importing\n"
-          @"    --clean-all: when importing rules via JSON clear all rules before importing\n"
+          @"    --clean: clear all non-transitive rules\n"
+          @"        Can be combined with --import to clear existing rules before importing.\n"
+          @"    --clean-all: clear all rules\n"
+          @"        Can be combined with --import to clear existing rules before importing.\n"
           @"\n"
           @"  Notes:\n"
           @"    The format of `identifier` when adding/checking a `signingid` rule is:\n"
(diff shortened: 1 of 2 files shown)


pmarkowsky (approved):
LGTM.

Source/santad/DataLayer/SNTRuleTable.m
```diff
@@ -423,7 +423,9 @@ - (SNTRule *)ruleForIdentifiers:(struct RuleIdentifiers)identifiers {
 #pragma mark Adding
 
 - (BOOL)addRules:(NSArray *)rules ruleCleanup:(SNTRuleCleanup)cleanupType error:(NSError **)error {
-  if (!rules || rules.count < 1) {
+  // Only accept an empty rules array if the cleanup-type is not none, that means we're
+  // doing a clean-out of the database with no rules added.
+  if ((!rules || rules.count < 1) && cleanupType != SNTRuleCleanupNone) {
```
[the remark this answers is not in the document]
russellhancox: While it doesn't _change_ anything, it does result in the method returning an appropriate error instead of nothing (which also keeps the current tests happy)

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -225,18 +226,16 @@ - (void)runWithArguments:(NSArray *)arguments {
...
-        [self printErrorUsageAndExit:@"--clean-all can only be used with --import"];
-        break;
-      default:
-        // This is a programming error.
-        LOGE(@"Unexpected SNTRuleCleanupType %ld", cleanupType);
-        exit(EXIT_FAILURE);
-    }
+    [[self.daemonConn remoteObjectProxy]
+        databaseRuleAddRules:@[]
+                 ruleCleanup:cleanupType
```
mlw, line 231: With 0 rules and a `cleanupType != SNTRuleCleanupNone` (from if statement) - this will be a no-op when hitting the new check in `addRules:ruleCleanup:error:`

russellhancox: I'd inverted this and forgot to put it back. Fixed

PR #366, opened 2025-04-16
rule download: add support for notification app name

tburgin:
This CL adds support for the new notification_app_name field during rule download.

### DIFF
--- a/Source/santasyncservice/SNTSyncRuleDownload.mm
+++ b/Source/santasyncservice/SNTSyncRuleDownload.mm
@@ -218,7 +218,8 @@ SNTRuleCleanup SyncTypeToRuleCleanup(SNTSyncType syncType) {
   NSString *appName = StringToNSString(protoRule->notification_app_name());
   if (appName.length) {
     // If notification_app_name is set but this is a clean sync, return early. We don't want to
-    // continue processing with deprecated behavior.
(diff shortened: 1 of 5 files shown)


Source/santasyncservice/SNTSyncRuleDownload.mm
```diff
@@ -214,6 +214,24 @@ - (void)announceUnblockingRules:(NSArray<SNTRule *> *)newRules {
 
 - (void)processBundleNotificationsForRule:(SNTRule *)rule
                             fromProtoRule:(const ::pbv1::Rule *)protoRule {
+  // Display a system notification if notification_app_name is set and this is not a clean sync.
+  NSString *appName = StringToNSString(protoRule->notification_app_name());
+  if (appName.length) {
+    // If notification_app_name is set but this is a clean sync, return early. We don't want to
+    // continue processing with deprecated behavior.
```
russellhancox, line 221:
```suggestion
    // If notification_app_name is set but this is a clean sync, return early. We don't want to
    // spam users with notifications for many apps that might be included in a clean sync, and
    // we don't want to fallback to the deprecated behavior
```

russellhancox (dismissed):

russellhancox (approved):

PR #378, opened 2025-04-30
Fixed up background apps plist example

sysophost:
The example background apps plist was syntactically invalid according to `plutil -lint`.

This appeared to be caused by the `PayloadScope` having no value following it. Moving the value `System` from line `23` up to `18` fixes this.

Also added opening `xml` tag to ensure consistency with other documented examples and fixed up the quotes in the `DOCTYPE` tag.


mlw (approved):
LGTM. Thank you!

russellhancox (approved):

PR #388, opened 2025-05-02
Add Timer mixin, adopt in Logger class.

mlw:


Source/common/Timer.h
```diff
@@ -0,0 +1,99 @@
...
+  void SetTimerInterval(NSTimeInterval interval_seconds) {
+    interval_seconds_ = std::min(interval_seconds, minimum_interval_);
+    UpdateTimingParameters();
+  }
+
+  void TimerCallback() { static_cast<T *>(this)->OnTimer(); }
+
+ private:
+  /// Update the timer firing settings. If a timer is currently active, will
+  /// resutl in it immediately firing and then again at the current interval.
```
russellhancox, line 75:
```suggestion
  /// result in it immediately firing and then again at the current interval.
```

Source/santad/Santad.mm
```diff
@@ -547,6 +547,38 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
...
+                                   BOOL newBool = [newValue boolValue];
+
+                                   if (oldBool == newBool) {
+                                     return;
+                                   }
+
+                                   LOGI(@"EnableTelemetryUpload changed: %d -> %d", oldBool,
+                                        newBool);
+
+                                   logger->StartTimer();
```
russellhancox, line 564: What if the key was set to false?

mlw: d'oh

PR #389, opened 2025-05-02
gui: Add 'sync' button to About window

russellhancox:
### Success

https://github.com/user-attachments/assets/4bf0c7d7-c2d6-4c7b-909f-ea0994a5de46

### Failure

https://github.com/user-attachments/assets/1ff9fc13-b6fd-4440-ad24-6c8615f20bcd

The "Copy Logs" button in the failure popup will copy all of the logs that would have printed when running `santactl sync`.

### DIFF
--- a/Source/gui/SNTAboutWindowView.swift
+++ b/Source/gui/SNTAboutWindowView.swift
@@ -1,6 +1,10 @@
 import SwiftUI
 
+import santa_common_MOLXPCConnection
+import santa_common_SNTCommonEnums
 import santa_common_SNTConfigurator
+import santa_common_SNTStoredEvent
+import santa_common_SNTXPCSyncServiceInterface
 import santa_gui_SNTMessageView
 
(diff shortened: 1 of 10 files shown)


Source/gui/SNTAboutWindowView.swift
```diff
@@ -74,6 +81,120 @@ struct SNTAboutWindowView: View {
...
+    proxy?.sync(withLogListener: logListener.endpoint, syncType: .normal) { status in
+      lr = nil
+
+      DispatchQueue.main.sync {
+        inProgress = false
+        syncStatus = status
+      }
+
+      if status == .success {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
```
mlw, line 138: Minor nit - 5 seconds felt a little long looking at the video, but I don't have a strong opinion if it looked fine to you.

mlw (approved):

mlw (approved):

PR #391, opened 2025-05-07
Respect rule comments on import

mlw:


Source/santad/SNTDaemonControlController.mm
```diff
@@ -179,8 +179,8 @@ - (void)staticRuleCount:(void (^)(int64_t count))reply {
 }
 
 - (void)retrieveAllRules:(void (^)(NSArray<SNTRule *> *, NSError *))reply {
+#ifndef DEBUG
   SNTConfigurator *config = [SNTConfigurator configurator];
-
   // Do not return any rules if syncBaseURL is set and return an error.
```
pmarkowsky, line 184: nit: This comment should be updated 

```suggestion
  // Do not return any rules if syncBaseURL or static rules are set and return an error.
```

PR #394, opened 2025-05-07
FAA glob expansion enhancements

mlw:
This PR changes behavior of FAA path glob expansion.

Previously, this worked rather similarly to shell expansion. Matching paths that existed at the time of application would begin to be watched. A special carve out was made for paths that did not contain glob characters in that they could be watched even if the path didn't exist. This means that only rules with non-glob paths could prevent things like file creation. It could also mean that an FAA rule author might encounter unexpected rules to be matched depending on the state of the filesystem when rules were applied since it could've been the case that some subset of rules were ignored if no path existed.

The new implementation here tries to allow for much more flexible rule application when paths don't exist for paths with glob characters so that they work largely similar to non-glob paths. This works by performing glob expansion on FAA rule paths for every subpath component, building up potentially matching paths. As long as each subpath component can match at least one existing path on the filesystem, Santa can now begin watching potentially non-existent files.

For example, given a filesystem state like:
```
/tmp/foo/hi.txt
/tmp/bar/bye.txt
```
And given an FAA rule path like:
```
/tmp/*/hi.txt
```
The current implementation would start watching only `/tmp/foo/hi.txt`. This new implementation will also begin watching for the non-existent `/tmp/bar/hi.txt` such that the authorization of the creation of that file is now gated by the rule covering that path.

### DIFF
--- a/Source/santad/DataLayer/WatchItems.mm
+++ b/Source/santad/DataLayer/WatchItems.mm
@@ -684,37 +684,119 @@ SetPairPathAndType DataWatchItems::operator-(const DataWatchItems &other) const 
   return diff;
 }
 
-bool DataWatchItems::Build(SetSharedDataWatchItemPolicy data_policies) {
+void FindMatches(NSString *base, NSMutableArray<NSString *> *path_components, NSUInteger idx,
+                 std::vector<std::string> &matches) {
+  if (path_components.count == idx) {
+    // Nothing left to match, add the current full base path
+    matches.push_back(base.UTF8String);
+    return;
+  }
+
+  NSString *path = [NSString stringWithFormat:@"%@%@", base, path_components[idx]];
+
   glob_t *g = (glob_t *)alloca(sizeof(glob_t));
-  for (const std::shared_ptr<DataWatchItemPolicy> &item : data_policies) {
-    int err = glob(item->path.c_str(), 0, nullptr, g);
-    if (err != 0 && err != GLOB_NOMATCH) {
-      LOGE(@"Failed to generate path names for watch item: %s", item->name.c_str());
-      return false;
-    }
-
-    // If no paths were returned, check if the search pattern contained any magic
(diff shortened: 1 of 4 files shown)


Source/santad/DataLayer/WatchItems.mm
```diff
@@ -684,37 +684,119 @@ bool ParseConfig(NSDictionary *config, SetSharedDataWatchItemPolicy &data_polici
...
+                 std::vector<std::string> &matches) {
+  if (path_components.count == idx) {
+    // Nothing left to match, add the current full base path
+    matches.push_back(base.UTF8String);
+    return;
+  }
+
+  NSString *path = [NSString stringWithFormat:@"%@%@", base, path_components[idx]];
+
   glob_t *g = (glob_t *)alloca(sizeof(glob_t));
```
tburgin, line 697: Could we not put this on the stack?

mlw: Unfortunately not. This is an Objective-C++ issue with how the struct is defined.

```
....mm:18:10: error: call to implicitly-deleted default constructor of 'glob_t'
   18 |   glob_t g;
      |          ^
In file included from ....mm:9:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/glob.h:63:9:
  note: default constructor of '(anonymous union at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/usr/include/glob.h:59:2)'
  is implicitly deleted because variant field 'gl_errblk' is an ObjC pointer
   63 |                 int (^gl_errblk)(const char *, int);
```

tburgin: Ack

russellhancox (approved):

PR #395, opened 2025-05-09
Collect, open, and send telemetry files to sync service for processing

mlw:
The PR starts many of the foundational pieces for eventual support for exporting telemetry.

The overall flow is:
1. Export is triggered via a timer in `santad`, or a new `santactl telemetry` command
    1.1. If triggered via `santactl`, an XPC message is sent to `santad` to start exporting
2. `santad` will query the configured writer (based on the `EventLogType`) for files to export
    1.1.  For each telemetry file, it is first opened
    1.2. The file descriptor is sent to the sync service to perform the actual export
    * The file descriptor is passed because the sync service doesn't have permissions to read the log files and also ensures logic for handling log files is kept within the daemon
3. When `santad` receives a response from the sync service, it will inform the writer whether or not the action was successful so it can do appropriate cleanup.

This currently can only be enabled in debug builds, including the new `santactl telemetry` command.

### DIFF
--- a/Source/santad/SNTSyncdQueue.mm
+++ b/Source/santad/SNTSyncdQueue.mm
@@ -197,7 +197,6 @@
 
 - (void)exportTelemetryFile:(NSFileHandle *)telemetryFile
           completionHandler:(void (^)(BOOL))completionHandler {
-  LOGE(@"Syncd queue, about to dispatch...");
   [self dispatchBlockOnSyncdQueue:^{
     if (self.syncConnection.remoteObjectProxy) {
       [self.syncConnection.remoteObjectProxy exportTelemetryFile:telemetryFile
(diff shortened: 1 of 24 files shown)


Source/santad/SNTSyncdQueue.mm
```diff
@@ -185,6 +195,20 @@ - (void)dispatchBlockOnSyncdQueue:(void (^)(void))block {
   });
 }
 
+- (void)exportTelemetryFile:(NSFileHandle *)telemetryFile
+          completionHandler:(void (^)(BOOL))completionHandler {
+  LOGE(@"Syncd queue, about to dispatch...");
```
russellhancox, line 200: Leftover? Maybe make LOGD if you want to keep?

mlw: Ope, leftover

russellhancox (approved):

PR #400, opened 2025-05-13
Remove run time and compile time checks for macOS 13

mlw:
Since we bumped our min supported version to macOS 13, we can remove the runtime and compile time checks for macOS 13 from the codebase.

macOS 12 support dropped here: #399


Source/common/SNTFileInfoTest.m
```diff
@@ -96,20 +92,6 @@ - (void)testPageZero {
 
 - (void)testKext {
```
russellhancox, line 93: Think we can probably remove this

russellhancox (approved):

PR #405, opened 2025-05-16
gui: Simplify display of bundle hash progress, animate disappearing

russellhancox:
https://github.com/user-attachments/assets/751ae39c-55f9-4911-a3f3-44a246e9066f

### DIFF
--- a/Source/gui/SNTBinaryMessageWindowView.swift
+++ b/Source/gui/SNTBinaryMessageWindowView.swift
@@ -269,6 +269,10 @@ struct SNTBinaryMessageWindowView: View {
         SNTNotificationSilenceView(silence: $preventFutureNotifications, period: $preventFutureNotificationPeriod)
       }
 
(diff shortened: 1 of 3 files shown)


Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -269,17 +269,15 @@ struct SNTBinaryMessageWindowView: View {
...
-        if bundleProgress.fractionCompleted == 0.0 {
-          ProgressView() {
-            Text(bundleProgress.label)
-          }.progressViewStyle(.linear)
-        } else {
-          ProgressView(value: bundleProgress.fractionCompleted) {
-            Text(bundleProgress.label)
-          }
-        }
+      ProgressView(value: (bundleProgress.fractionCompleted == 0.0 ? nil : bundleProgress.fractionCompleted)) {
```
mlw, line 272: This feels odd with different types - `nil` vs. `fractionCompleted` (double).

PR #410, opened 2025-05-22
santactl: add eventupload command

tburgin:
This CL adds a `santactl eventupload` command. This allows a user to upload metadata about an executable or bundle to a sync server, out of band of Santa's typical operations.

There was a bit of refactoring work to make this possible, mainly around making the bundle service multi-tenant. We don't want `santactl eventupload` invocations interrupting each other or the GUI during hashing events. In short, the bundle service is now safe to call from multiple clients, allowing each request to be separately cancelable.

https://github.com/user-attachments/assets/2eb27e7d-56d1-4e30-bd2e-201d80ae9988


Source/santabundleservice/SNTBundleService.mm
```diff
@@ -139,11 +164,12 @@ - (NSDictionary *)findRelatedBinaries:(SNTStoredEvent *)event progress:(NSProgre
   //
   // Xcode.app has roughly 500k files, 8bytes per pointer is ~4MB for this array. This size to space
   // ratio seems appropriate as Xcode.app is in the upper bounds of bundle size.
-  __block void **fis = calloc(subpaths.count, sizeof(void *));
+  __block void **fis = static_cast<void **>(calloc(subpaths.count, sizeof(void *)));
```
mlw, line 167: Not a blocker, but might be nice to consider an `std::vector` since this is an ObjC++ file now - then we wouldn't have to worry about freeing it manually.

Source/santactl/Commands/SNTCommandEventUpload.mm
```diff
@@ -0,0 +1,218 @@
...
+
+  // Hash the bundle if the server declares support and if the caller has not explicitly disabled
+  // bundle hashing.
+  __block BOOL enableBundles = NO;
+  [self.daemonConn.synchronousRemoteObjectProxy enableBundles:^(BOOL response) {
+    enableBundles = response;
+  }];
+  if (skipBundles) {
+    enableBundles = false;
+  }
```
mlw, line 105: Nit - the NO/false double setting took me a bit to parse. Maybe something like this is cleaner, and also let's us skip asking the daemon if not necessary:
```objc
  __block BOOL enableBundles = !skipBundles;
  if (enableBundles) {
    [self.daemonConn.synchronousRemoteObjectProxy enableBundles:^(BOOL response) {
      enableBundles = response;
    }];
  }
```

mlw (changes requested):

Source/santabundleservice/SNTBundleService.mm
```diff
@@ -111,24 +121,39 @@ - (void)hashBundleBinariesForEvent:(SNTStoredEvent *)event reply:(SNTBundleHashB
...
+    }
   });
 }
 
+#pragma mark Internal Methods
+
 - (void)spindown {
-  LOGI(@"Spinning down");
-  exit(0);
+  if (_current_events == 0) {
```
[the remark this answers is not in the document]
tburgin: We discussed out of band and have decided to just let macOS manage the lifecycle of the process instead of spinning down ourselves. EnablePressuredExit is now enabled to hopefully allow more frequent reaping.

Conf/com.northpolesec.santa.bundleservice.plist
```diff
@@ -22,6 +22,8 @@
 	<string>Interactive</string>
```
russellhancox, line 22: s/Interactive/Adaptive

Source/common/StoredEventHelpers.mm
```diff
@@ -0,0 +1,46 @@
...
+    se.entitlements = cs.entitlements;
+    if (cs.signatureFlags & kSecCodeSignatureAdhoc) {
+      se.signingStatus = SNTSigningStatusAdhoc;
+    } else if (IsDevelopmentCert(cs.leafCertificate)) {
+      se.signingStatus = SNTSigningStatusDevelopment;
+    } else {
+      se.signingStatus = SNTSigningStatusProduction;
+    }
+  }
+  return se;
```
russellhancox, line 45:
```suggestion
  if (csError) {
    se.signingStatus =
        (csError.code == errSecCSUnsigned) ? SNTSigningStatusUnsigned : SNTSigningStatusInvalid;
    return se;
  }
  se.signingChain = cs.certificates;
  se.cdhash = cs.cdhash;
  se.teamID = cs.teamID;
  se.signingID = FormatSigningID(cs);
  se.entitlements = cs.entitlements;
  if (cs.signatureFlags & kSecCodeSignatureAdhoc) {
    se.signingStatus = SNTSigningStatusAdhoc;
  } else if (IsDevelopmentCert(cs.leafCertificate)) {
    se.signingStatus = SNTSigningStatusDevelopment;
  } else {
    se.signingStatus = SNTSigningStatusProduction;
  }
  return se;
```

Conf/com.northpolesec.santa.bundleservice.plist
tburgin: I am going to save that change for a future PR so I can test it a bit more thoroughly.

Source/common/StoredEventHelpers.mm
tburgin: Done

russellhancox (approved):

mlw (approved):

Source/gui/SNTNotificationManager.m
```diff
@@ -76,16 +76,13 @@ - (void)windowDidCloseSilenceHash:(NSString *)hash withInterval:(NSTimeInterval)
...
+  [self.currentWindowController.bundleListener invalidate];
   [self.pendingNotifications removeObject:self.currentWindowController];
   self.currentWindowController = nil;
 
   if (self.pendingNotifications.count) {
     [self showQueuedWindow];
   } else {
-    MOLXPCConnection *bc = [SNTXPCBundleServiceInterface configuredConnection];
-    [bc resume];
-    [[bc remoteObjectProxy] spindown];
```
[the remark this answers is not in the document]
russellhancox: The NSProgress magic still works. Unlike before, `santabundleservice` doesn't disappear, as expected, but after dismissing the window the CPU usage very quickly drops back down to 0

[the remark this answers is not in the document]
mlw: Whoa. Surprised by this!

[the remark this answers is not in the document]
mlw: Resolving as it seems no need to file a followup issue.

PR #414, opened 2025-05-23
Add EnableBundle sync config to sync state plist

mlw:
This config key was previously not set until the first sync. This change will store the previous setting in the sync state plist so that any events captured before the first sync will opportunistically use the same previous behavior.


pmarkowsky (dismissed):
LGTM.

Source/common/SNTConfigurator.m
```diff
@@ -742,6 +748,14 @@ - (void)setEnableTransitiveRules:(BOOL)enabled {
   [self updateSyncStateForKey:kEnableTransitiveRulesKey value:@(enabled)];
 }
 
+- (BOOL)enableBundles {
```
russellhancox, line 751: There's a comment above this property in the header that needs updating

russellhancox (approved):

russellhancox (approved):

PR #415, opened 2025-05-23
bundle service: Adaptive priority

tburgin:
From reading the docs, it seems like this is the correct priority - though in practice I have not noticed a difference in performance or EnablePressuredExit when under load. Doesn't seem to hurt, so let's go for it.


pmarkowsky (approved):
LGTM.

PR #416, opened 2025-05-23
Initial plumbing for export configuration

mlw:
Adopts new export config preflight response setting. The configuration data received is modeled by the new SNTExportConfiguration class. This PR also sets up the initial plumbing to get this data from the sync server over to the daemon.

### DIFF
--- a/Source/common/SNTConfigBundle.h
+++ b/Source/common/SNTConfigBundle.h
@@ -14,7 +14,7 @@
 
 #import <Foundation/Foundation.h>
 
-#include "Source/common/SNTCommonEnums.h"
+#import "Source/common/SNTCommonEnums.h"
 
 @class SNTExportConfiguration;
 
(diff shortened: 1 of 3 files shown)


Source/common/CoderMacros.h
```diff
@@ -15,46 +15,53 @@
 #ifndef SANTA__COMMON__CODERMACROS_H
 #define SANTA__COMMON__CODERMACROS_H
 
-// Encode the property keyed by the property name.
+/// Encode the property keyed by the property name.
```
tburgin, line 18: What is up with the triple `/` for code comments?

mlw: The hope was that they would be picked up as doc strings, but it didn't really work for me. Might be a macro-thing? Left them in cause using doc strings still felt right, but we're not at all consistent across the project...

russellhancox: I originally used triple-slashes for docstrings because Xcode had some rudimentary support for them (which is why there are also `@return` and `@param` comments in a few places). Once we moved away from Xcode I stopped being picky about using them, though I agree that triple-slashes for docstrings to differentiate from "normal" comments is a good idea and we should do continue to do it.

PR #420, opened 2025-05-28
gui: Allow clean syncing by holding Option

russellhancox:
https://github.com/user-attachments/assets/6f04e105-83d7-4104-af32-302a7eaaeebc

### DIFF
--- a/Source/gui/SNTAboutWindowView.swift
+++ b/Source/gui/SNTAboutWindowView.swift
@@ -190,7 +190,7 @@ struct SyncButtonView: View {
       } else if modifierFlags.isOptionPressed() {
         Text("Clean Sync").frame(width: 90.0)
       } else {
-        Text("Sync").frame(width: 90.0)
+        Text("Sync").frame(width: 90.0).help("Hold Option for Clean Sync")
       }
     }
     .disabled(inProgress || syncStatus != .unknown)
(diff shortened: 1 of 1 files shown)


mlw:
Any thoughts on how to have some visual aid that this feature exists? One idea would be to add a hover tooltip over the button. Another thought was to add a small-print label underneath the button like "Hold Option Key To Clean Sync", but this wouldn't fit very well...

russellhancox:
> Any thoughts on how to have some visual aid that this feature exists? One idea would be to add a hover tooltip over the button. Another thought was to add a small-print label underneath the button like "Hold Option Key To Clean Sync", but this wouldn't fit very well...

Honestly I wasn't planning on any visual aid, just documentation, though I could be convinced about a tooltip. Clean syncing should very rarely be required, it's _usually_ papering over a problem or something the server should be dealing with.

pmarkowsky (approved):
LGTM.

pmarkowsky:
Only question I have is do we want another modifier for `SNTSyncTypeCleanAll`?

pmarkowsky:
Also another idea re:how to discover this, is we could make a menu option in the Santa menu

Screenshot 2025-05-31 at 18 47 42

russellhancox:
> Only question I have is do we want another modifier for SNTSyncTypeCleanAll?

No. I'm wary enough of making clean syncs very easy. "Clean Sync All" is potentially very disruptive.

> Also another idea re:how to discover this, is we could make a menu option in the Santa menu

The main menu is created in AppDelegate where there's no access to the function that handles the syncing. We could modify it in the AboutWindow view but then we have to be careful to remove it when the view goes away or any ability to trigger the menu item would cause a crash.

Overall, I don't think it's worth it.

pmarkowsky (approved):
LGTM.

PR #423, opened 2025-05-29
Remove `EnableDebugLogging` config key. Support `santactl sync --debug`

mlw:
This PR changes Santa logging to switch to `os_log` instead of older, deprecated ASL functionality. The primary goal was to eliminate the `SNTConfigurator` dependency from `SNTLogging`, which was preventing the ability to log in both the configurator and in any of the configurator's dependencies.

The `EnableDebugLogging` configuration key is no longer necessary as this filtering can be done by system tools. E.g., to view debug logs, you can use:

```
log stream --level debug --predicate 'sender == "com.northpolesec.santa.daemon"'
```

Additionally, `santactl sync` no longer prints debug logs by default. However, you can now use `santactl sync --debug` to get these logs.

The "sync" command that is triggered via the About window UI will allways include debug logs if an error was encountered.

Finally, new `TEE_LOG*` log macro variants were added. These are primarily for use within `santactl`. There are some technical inconsistencies across the commands in how they log and whether or not the output goes to terminal or the system log. This PR attempts to standardize much of this under the following overarching logic:

* If a command previously used `LOG*` macros, it will now use `TEE_LOG*`
* If a command was writing to `stderr` via `fprintf`, these are now sent to `TEE_LOGE`.
* If a command is going to exit due to some error, it will log via `TEE_LOGE`
* Commands that sent there normal output to terminal via `printf` were left untouched.


Source/common/SNTLogging.h
```diff
@@ -1,59 +1,72 @@
...
+#define SNT_LOG_WITH_TYPE(type, fmt, ...)              \
+  os_log_with_type(OS_LOG_DEFAULT, type, "%{public}s", \
+                   [[NSString stringWithFormat:fmt, ##__VA_ARGS__] UTF8String])
+
+#define SNT_PRINT_LOG(file, fmt, ...) \
+  fprintf(file, "%s\n", [[NSString stringWithFormat:fmt, ##__VA_ARGS__] UTF8String]);
+
+#define LOGD(logFormat, ...) SNT_LOG_WITH_TYPE(OS_LOG_TYPE_DEBUG, logFormat, ##__VA_ARGS__)
+#define LOGI(logFormat, ...) SNT_LOG_WITH_TYPE(OS_LOG_TYPE_INFO, logFormat, ##__VA_ARGS__)
+#define LOG(logFormat, ...) SNT_LOG_WITH_TYPE(OS_LOG_TYPE_DEFAULT, logFormat, ##__VA_ARGS__)
```
russellhancox, line 34: I don't like there being a 'bare' LOG without a level. It won't be clear in context whether that's higher or lower than INFO or WARN, for instance, and the 4 levels we have have been plenty. If this exists just to use the OS_LOG_TYPE_DEFAULT level, why not map LOGW to default?

Source/common/SNTLogging.h
```diff
@@ -1,59 +1,72 @@
...
+    SNT_PRINT_LOG(stdout, logFormat, ##__VA_ARGS__); \
+  } while (0)
+
+#define TEE_LOGI(logFormat, ...)                     \
+  do {                                               \
+    LOGI(logFormat, ##__VA_ARGS__);                  \
+    SNT_PRINT_LOG(stdout, logFormat, ##__VA_ARGS__); \
+  } while (0)
+
+#define TEE_LOG(logFormat, ...)                      \
```
russellhancox, line 52: Same here

russellhancox (changes requested):

Source/common/SNTLogging.h
mlw: Was added because "warn" doesn't map cleanly in the os_log world. While originally I mapped LOGW to OS_LOG_TYPE_DEFAULT, this looked odd when adding TEE_LOGW which would write to stderr. However, thinking on it more, I think this is fine. "Warn" messages are not errors, so no need to use the ERROR log type. And for the TEE variant, we can still write to stderr so warn-level logs don't interfere with stdout.

russellhancox (approved):

PR #424, opened 2025-05-29
common: Add CEL evaluator

russellhancox:
This is the first part of implementing CEL rules (as a means to complete #15, and google/santa#1200).

This PR adds the necessary dependencies and a new CELEvaluator class that holds a configured compiler. The evaluator can compile CEL expressions into a re-usable expression plan and then evaluate it using static per-binary data (e.g. signing timestamp) and a block to retrieve dynamic per-process data (e.g. command-line args). 

This new library is as yet unused and won't be compiled in. The next PR will add the ability to use CEL rules during execution.

### DIFF
--- a/Source/common/CEL.cc
+++ b/Source/common/CEL.cc
@@ -1,0 +1,236 @@
+/// Copyright 2025 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///    http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#include "Source/common/CEL.h"
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "cel/expr/checked.pb.h"
+#include "common/ast_proto.h"
+#include "compiler/compiler_factory.h"
+#include "compiler/standard_library.h"
+#include "eval/public/activation_bind_helper.h"
+#include "eval/public/builtin_func_registrar.h"
+#include "eval/public/cel_expr_builder_factory.h"
+#include "eval/public/cel_function_adapter.h"
+
+namespace cel_runtime = ::google::api::expr::runtime;
+namespace syncv1 = ::santa::sync::v1;
+namespace pbv1 = ::santa::cel::v1;
+
+namespace santa {
+
+// Creates a CEL Activation with the static and dynamic contexts attached
+// appropriately.
+absl::StatusOr<cel_runtime::Activation> CreateActivation(
+    ::pbv1::CELStaticContext *static_context,
+    ::pbv1::CELDynamicContext * (^dynamicBlock)(
+        ::google::protobuf::Arena *arena)) {
+  cel_runtime::Activation activation;
+
+  if (dynamicBlock) {
+    absl::StatusOr<std::unique_ptr<cel_runtime::CelFunction>> getArgsFunction =
+        cel_runtime::FunctionAdapter<::pbv1::CELDynamicContext *>::Create(
+            "getDynamic", false,
+            std::function<::pbv1::CELDynamicContext *(
+                ::google::protobuf::Arena *)>(
+                [dynamicBlock](::google::protobuf::Arena *arena)
+                    -> ::pbv1::CELDynamicContext * {
+                  return dynamicBlock(arena);
+                }));
+    if (!getArgsFunction.ok()) {
+      return getArgsFunction.status();
+    }
+
+    if (auto result =
+            activation.InsertFunction(std::move(getArgsFunction.value()));
+        !result.ok()) {
+      return result;
+    }
+  }
+
+  google::protobuf::Arena arena;
+  if (auto result = BindProtoToActivation(
+          static_context, &arena, &activation,
+          cel_runtime::ProtoUnsetFieldOptions::kBindDefault);
+      !result.ok()) {
+    return result;
+  }
+
+  return activation;
+}
+
+absl::Status CELEvaluator::Initialize() {
+  // Link the reflection for needed messages to that the CEL compiler can
+  // recognize them.
+  google::protobuf::LinkMessageReflection<::pbv1::CELStaticContext>();
+  google::protobuf::LinkMessageReflection<::pbv1::CELDynamicContext>();
+  // We need the Policy enum to be linked in and the library doesn't seem to
+  // have a way to force enum reflection to be linked, so we're using the Rule
+  // message as that contains a field that uses Policy.
+  google::protobuf::LinkMessageReflection<::syncv1::Rule>();
+
+  // Create a compiler builder with the generated descriptor pool for protos.
+  absl::StatusOr<std::unique_ptr<::cel::CompilerBuilder>> builderStatus =
+      ::cel::NewCompilerBuilder(
+          google::protobuf::DescriptorPool::generated_pool());
+  if (!builderStatus.ok()) {
+    return builderStatus.status();
+  }
+  auto builder = std::move(builderStatus.value());
+
+  // Add the standard library.
+  if (auto result = builder->AddLibrary(::cel::StandardCompilerLibrary());
+      !result.ok()) {
+    return result;
+  }
+
+  ::cel::TypeCheckerBuilder &checker_builder = builder->GetCheckerBuilder();
+
+  // Add the static context type. The value will be bound during activation.
+  if (auto result = checker_builder.AddContextDeclaration(
+          ::pbv1::CELStaticContext::descriptor()->full_name());
+      !result.ok()) {
+    return result;
+  }
+
+  // Add the single getDynamic function overload.
+  // The implementation will be bound during activation.
+  absl::StatusOr<::cel::FunctionDecl> decl = ::cel::MakeFunctionDecl(
+      "getDynamic",
+      ::cel::MakeOverloadDecl(
+          "getDynamic",
+          ::cel::MessageType(::pbv1::CELDynamicContext::descriptor())));
+  if (!decl.ok()) {
+    return decl.status();
+  }
+  if (absl::Status result = checker_builder.MergeFunction(decl.value());
+      !result.ok()) {
+    return result;
+  }
+
+  // Build and store the compiler.
+  absl::StatusOr<std::unique_ptr<::cel::Compiler>> compiler = builder->Build();
+  if (!compiler.ok()) {
+    return compiler.status();
+  }
+  compiler_ = std::move(compiler.value());
+  return absl::OkStatus();
+}
+
+absl::StatusOr<std::unique_ptr<::cel_runtime::CelExpression>>
+CELEvaluator::Compile(absl::string_view expr) {
+  if (!compiler_) {
+    return absl::InvalidArgumentError(
+        "CELEvaluator::Initialize() must be called before Compile()");
+  }
+
+  // Compile the expression.
+  absl::StatusOr<::cel::ValidationResult> result = compiler_->Compile(expr);
+  if (!result.ok()) {
+    return absl::Status(std::move(result.status()));
+  }
+  if (!result.value().IsValid() || result.value().GetAst() == nullptr) {
+    return absl::InvalidArgumentError(result.value().FormatError());
+  }
+
+  // Check the AST for correctness.
+  ::cel::expr::CheckedExpr cel_expr;
+  if (absl::Status status =
+          ::cel::AstToCheckedExpr(*result.value().GetAst(), &cel_expr);
+      !status.ok()) {
+    return status;
+  }
+
+  // Setup a default environment for building expressions.
+  cel_runtime::InterpreterOptions options;
+  std::unique_ptr<cel_runtime::CelExpressionBuilder> builder =
+      CreateCelExpressionBuilder(
+          google::protobuf::DescriptorPool::generated_pool(),
+          google::protobuf::MessageFactory::generated_factory(), options);
+
+  // Register the builtin functions.
+  if (auto result = RegisterBuiltinFunctions(builder->GetRegistry(), options);
+      !result.ok()) {
+    return result;
+  }
+
+  // Register the getDynamic function.
+  if (auto result = builder->GetRegistry()->RegisterLazyFunction(
+          ::cel::FunctionDescriptor("getDynamic", false, {}));
+      !result.ok()) {
+    return result;
+  }
+
+  // Create an expression plan with the checked expression.
+  absl::StatusOr<std::unique_ptr<cel_runtime::CelExpression>> expression_plan =
+      builder->CreateExpression(&cel_expr);
+  return expression_plan;
+};
+
+absl::StatusOr<::santa::sync::v1::Policy> CELEvaluator::Evaluate(
+    const ::cel_runtime::CelExpression *expression_plan,
+    ::pbv1::CELStaticContext *static_context,
+    ::pbv1::CELDynamicContext * (^dynamicBlock)(
+        ::google::protobuf::Arena *arena)) {
+  // Create an activation with the static and dynamic contexts.
+  absl::StatusOr<cel_runtime::Activation> activation =
+      CreateActivation(static_context, dynamicBlock);
+  if (!activation.ok()) {
+    return activation.status();
+  }
+  // Evaluate the parsed expression.
+  google::protobuf::Arena arena;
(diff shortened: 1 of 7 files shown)


Source/common/CEL.h
```diff
@@ -0,0 +1,83 @@
...
+#include "eval/public/cel_expression.h"
+#pragma clang diagnostic pop
+
+namespace santa {
+
+class CELEvaluator {
+ public:
+  // Initialize the evaluator. This must be called before any other methods as
+  // it initializes the CEL compiler.
+  absl::Status Initialize();
```
mlw, line 41: It doesn't seem like this class has any useful functionality if not `Initialized`. It would be cleaner API to add a static factory method that creates the compiler, and create a non-default constructor to move the compiler into.

Source/common/CEL.cc
```diff
@@ -0,0 +1,236 @@
...
+  if (dynamicBlock) {
+    absl::StatusOr<std::unique_ptr<cel_runtime::CelFunction>> getArgsFunction =
+        cel_runtime::FunctionAdapter<::pbv1::CELDynamicContext *>::Create(
+            "getDynamic", false,
+            std::function<::pbv1::CELDynamicContext *(
+                ::google::protobuf::Arena *)>(
+                [dynamicBlock](::google::protobuf::Arena *arena)
+                    -> ::pbv1::CELDynamicContext * {
+                  return dynamicBlock(arena);
+                }));
```
mlw, line 51: nit: This can be simplified, the `std::function` should be able to be removed and the type deduced from the lamba which would remove the duplication. You should also be able to just use block syntax to keep things consistent with the rest of the file.
```
^::pbv1::CELDynamicContext *(::google::protobuf::Arena *arena) {
    return dynamicBlock(arena);
}
```

Source/common/CEL.h
```diff
@@ -0,0 +1,83 @@
...
+  // executions of a given binary and can safely be cached.
+  //
+  // If the optional `dynamicBlock` block is provided and the expression uses
+  // uses the `getDynamic` function, the block will be called to retireve the
+  // dynamic context. Evaluations that make use of this dynamic context should
+  // not be cached.
+  absl::StatusOr<::santa::sync::v1::Policy> Evaluate(
+      ::google::api::expr::runtime::CelExpression const *expression_plan,
+      ::santa::cel::v1::CELStaticContext *static_context,
+      ::santa::cel::v1::CELDynamicContext * (^dynamicBlock)(::google::protobuf::Arena *arena));
```
mlw, line 60: We should consider an overloaded interface without the optional dynamic block.

Source/common/CEL.proto
mlw, line 1: Were you wanting to keep this here instead of the protos repo?

mlw (changes requested):

Source/common/CEL.cc
mlw, line 1: Rename to `.mm` - doesn't matter much now, but would reduce some headache if we ever want to do things like add an interface that takes an `id` object, or ever need to add a dispatch queue and want ARC to manage the lifetime.

Source/common/CEL.proto
russellhancox: For now, yes

Source/common/cel/memoizer.h
```diff
@@ -0,0 +1,54 @@
...
+  // Constructor takes the function to be memoized
+  Memoizer(std::function<T()> func) : func_(func) {}
+
+  // Overload the operator() to enable calling the Memoizer like a function
+  // Mark this as const to allow it to be called from const methods. It
+  // technically isn't const given that cache_ is updated but we mark that field
+  // as mutable.
+  T operator()() const {
+    if (cache_.has_value()) {
+      return cache_.value();
```
mlw: super nit: `s/cache_.value();/*cache_/`

Update: this is done in a lot of places. In general, if the `has_value()` check has already done, we should prefer `operator*()` (or `operator->()`).

Source/common/cel/memoizer.h
mlw, line 1: Love this class. we should move it up to `Source/common` since it's not CEL-specific. I think it'd be generally useful.

Source/common/cel/cel.proto
```diff
@@ -0,0 +1,49 @@
...
+  // silently blocked can be very confusing for users.
+  SILENT_BLOCKLIST = 4;
+}
+
+// The fields in this message provided to CEL programs.
+message Context {
+  FileContext file = 1;
+  repeated string args = 2;
+  repeated string envs = 3;
+}
```
mlw: I have some hand wavy concerns about naming. I think we should be more specific for if/when we expand our use of CEL. I guess this is maybe mitigated by us being able to define a separate proto for those future CEL programs? But if there are ever shared message types things might get dicey.

Consider: `message ExecutionContext {}` and the field name `FileContext target_executable`

Source/common/cel/context.h
```diff
@@ -0,0 +1,70 @@
...
+
+namespace cel_runtime = ::google::api::expr::runtime;
+namespace pbv1 = ::santa::cel::v1;
+
+namespace santa {
+namespace cel {
+
+// SantaActivation is a CEL activation that provides lookups of values from the
+// santa.pb.cel.v1.Context message, and easy access to variables for return values.
+class SantaActivation : public ::cel_runtime::BaseActivation {
```
mlw: Rename to `Activation` to remove the `santa::cel::SantaActivation` stutter

Source/common/cel/context.h
```diff
@@ -0,0 +1,70 @@
...
+ public:
+  SantaActivation(const ::pbv1::FileContext *file, std::vector<std::string> (^args)(),
+                  std::vector<std::string> (^envs)())
+      : file_(file), args_(args), envs_(envs) {};
+  ~SantaActivation() = default;
+
+  absl::optional<cel_runtime::CelValue> FindValue(absl::string_view name,
```
mlw: minor nit: I hadn't been good about using `absl::optional`, the rest of the code base uses `std::optional`. Additionally, we'll be moving to the [latest ABSL release](https://github.com/abseil/abseil-cpp/releases/tag/20250512.0) soon which changes `absl::optional` to an alias.

Source/common/cel/context.h
mlw, line 1: Filename of `context` goes against the grain of typically matching the name of the class. Consider renaming unless there's a good reason.

Source/common/cel/context.mm
```diff
@@ -0,0 +1,116 @@
...
+
+absl::optional<cel_runtime::CelValue> SantaActivation::FindValue(
+    absl::string_view name, google::protobuf::Arena *arena) const {
+  // Handle the ReturnValue values.
+  auto retDescriptor = pbv1::ReturnValue_descriptor();
+  for (int i = 1; i < retDescriptor->value_count(); i++) {
+    if (name == retDescriptor->value(i)->name()) {
+      return cel_runtime::CelValue::CreateInt64(retDescriptor->value(i)->number());
+    }
+  }
```
mlw: There looks to be a `FindValueByName` method to replace this loop: https://protobuf.dev/reference/cpp/api-docs/google.protobuf.descriptor/#EnumDescriptor

Source/common/cel/context.h
```diff
@@ -0,0 +1,70 @@
...
+                                                  google::protobuf::Arena *arena) const override;
+
+  // SantaActivation does not support lazy-loaded functions.
+  std::vector<const cel_runtime::CelFunction *> FindFunctionOverloads(
+      absl::string_view) const override {
+    return {};
+  }
+
+  static std::vector<std::pair<absl::string_view, ::cel::Type>> GetVariables(
+      google::protobuf::Arena *arena);
```
mlw: Do we need to worry about duplicate names? Obviously the names within a proto message would need to be unique to make it valid proto. But i' not sure what happens for embedded messages or if we ever accidentally, e.g., add some return enum that has the same name as a field in the Context message?

update: Maybe the checker_builder thing does this automagically?

Source/common/cel/context.mm
```diff
@@ -0,0 +1,116 @@
...
+
+  // Now add all the fields from the CELContext message.
+  auto ctxDescriptor = pbv1::Context::descriptor();
+  for (int i = 0; i < ctxDescriptor->field_count(); i++) {
+    auto field = ctxDescriptor->field(i);
+
+    ::cel::Type type;
+    switch (field->cpp_type()) {
+      case ::google::protobuf::FieldDescriptor::CPPTYPE_STRING: type = ::cel::StringType(); break;
+      case ::google::protobuf::FieldDescriptor::CPPTYPE_MESSAGE:
```
mlw: It's not clear to me what happens for embedded message fields. Do the embedded message fields just get automagically mapped for usage by a cel program? In our case, we have doubly nested messages (Context -> FileContext -> Timestamp). Does this work the same way?

Source/common/cel/context.mm
```diff
@@ -0,0 +1,116 @@
...
+      case ::google::protobuf::FieldDescriptor::CPPTYPE_BOOL: type = ::cel::BoolType(); break;
+      case ::google::protobuf::FieldDescriptor::CPPTYPE_DOUBLE: [[fallthrough]];
+      case ::google::protobuf::FieldDescriptor::CPPTYPE_FLOAT: type = ::cel::DoubleType(); break;
+    }
+
+    if (field->is_repeated()) {
+      type = ::cel::ListType(arena, type);
+    }
+
+    v.push_back({field->name(), type});
```
mlw: I'm a little concerned about lifetimes here. Are we guaranteed the backing memory for this `string_view` doesn't go away? If not, we could just store a `string` instead of `string_view`.

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+#include "eval/public/transform_utility.h"
+#include "parser/parser.h"
+
+namespace cel_runtime = ::google::api::expr::runtime;
+namespace pbv1 = ::santa::cel::v1;
+
+namespace santa {
+namespace cel {
+
+absl::StatusOr<std::unique_ptr<::cel::Compiler>> CreateCompiler(google::protobuf::Arena *arena) {
```
mlw: nit: make `static`

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+  if (!compiler_) {
+    return absl::InvalidArgumentError("Evaluator not properly initialized");
+  }
+
+  // Compile the expression.
+  absl::StatusOr<::cel::ValidationResult> result = compiler_->Compile(expr);
+  if (!result.ok()) {
+    return result.status();
+  }
+  if (!result->IsValid() || result->GetAst() == nullptr) {
```
mlw: the second condition here is duplicative:
https://github.com/google/cel-cpp/blob/8868650b3ec2b83cbf8ed10b798ac6877d605665/checker/validation_result.h#L44-L49

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+    return absl::InvalidArgumentError(result->FormatError());
+  }
+
+  // Check the AST for correctness.
+  ::cel::expr::CheckedExpr cel_expr;
+  if (absl::Status status = ::cel::AstToCheckedExpr(*result->GetAst(), &cel_expr); !status.ok()) {
+    return status;
+  }
+
+  // // Setup a default environment for building expressions.
```
mlw: super nit: double double slash

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+  cel_runtime::InterpreterOptions options;
+  options.constant_folding = true;
+  options.constant_arena = arena_.get();
```
mlw: Seems like we should add `enable_regex_precompilation` when `constant_folding` is set: https://github.com/google/cel-cpp/blob/8868650b3ec2b83cbf8ed10b798ac6877d605665/eval/public/cel_options.h#L137-L151

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+  // Create an expression plan with the checked expression.
+  absl::StatusOr<std::unique_ptr<cel_runtime::CelExpression>> expression_plan =
+      builder->CreateExpression(&cel_expr);
+
+  return expression_plan;
+};
+
+absl::StatusOr<::santa::cel::v1::ReturnValue> Evaluator::Evaluate(
+    const ::cel_runtime::CelExpression *expression_plan, const SantaActivation &activation) {
+  google::protobuf::Arena arena;
```
mlw: Is this leftover? Why not use the member variable?

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+  // A Policy value will be returned as-is
+  // Everything else is an error.
+  if (bool value; result->GetValue(&value)) {
+    if (value) {
+      return pbv1::ReturnValue::ALLOWLIST;
+    }
+    return pbv1::ReturnValue::BLOCKLIST;
+  } else if (int64_t value; result->GetValue(&value) && pbv1::ReturnValue_IsValid((int)value)) {
+    auto policy = static_cast<pbv1::ReturnValue>(value);
+    return policy;
```
mlw: super nit: no need for the variable, just return the value

Source/common/cel/context.h
mlw: We also have broken with the style guide on capitalization. This is largely due to the ObjC/ObjC++ classes using the SNT prefix, and then keeping that initial capitalization for all files (except proto i guess...). We should continue that unless there's good reason to break it.

Source/common/cel/evaluator.mm
```diff
@@ -0,0 +1,162 @@
...
+}
+
+absl::StatusOr<std::unique_ptr<Evaluator>> Evaluator::Create() {
+  std::unique_ptr<google::protobuf::Arena> arena = std::make_unique<google::protobuf::Arena>();
+
+  auto compiler = CreateCompiler(arena.get());
+  if (!compiler.ok()) {
+    return compiler.status();
+  }
+  return std::make_unique<Evaluator>(std::move(compiler.value()), std::move(arena));
```
mlw: Something I noticed when reviewing the tests is that users have to be careful to not use the result of the `Compile` output after the instance of the `Evaluator` goes out of scope because it would be tied to this `Area`. This won't likely occur for how things are setup now. But we should add a comment on the class for future readers so they know this constraint.

mlw (changes requested):

Source/common/cel/evaluator.mm
russellhancox: That's funny, I got this directly from the CEL codelab: https://github.com/google/cel-cpp/blob/8868650b3ec2b83cbf8ed10b798ac6877d605665/codelab/cel_compiler.h#L36

Source/common/cel/evaluator.mm
russellhancox: I think enabling `enable_regex_precompilation` requires (or benefits greatly from) `constant_folding` but having the latter doesn't require the former. But, I'm still going to enable regex precomp as it might benefit anyone using regexes.

Source/common/cel/evaluator.mm
russellhancox: It was

Source/common/cel/cel.proto
russellhancox: After offline discussion, WDYT?

Source/common/cel/context.h
russellhancox: I believe we _would_ need to worry, except that collisions are largely impossible. Variables are case sensitive, so if we added a return value with the same name as a field they wouldn't collide within the expressions. Embedded messages are accessed "through" a field, e.g. the fields inside FileContext can't collide with anything.

Source/common/cel/context.mm
russellhancox: Yeah, it handles embedded message automatically.

Source/common/cel/context.mm
russellhancox: The field name comes from the proto descriptors which are compiled into the binary, so this should be safe.

Source/common/cel/context.h
russellhancox: That's fair, I'll rename to Activation.[h,mm] and Evaluator.[h,mm] once you've finished reviewing (otherwise the context for comments will be lost)

Source/common/cel/context.mm
```diff
@@ -0,0 +1,159 @@
...
+#include "eval/public/structs/cel_proto_wrapper.h"
+#pragma clang diagnostic pop
+
+namespace cel_runtime = ::google::api::expr::runtime;
+
+namespace santa {
+namespace cel {
+
+template <>
+cel_runtime::CelValue Activation::CELValue(const int &v, google::protobuf::Arena *unused_arena) {
```
mlw: Is this specialization necessary? I'd expect widening to `int64_t` would be fine?

Source/common/cel/cel.proto
mlw: lgtm

Source/common/cel/context.mm
russellhancox: It is, or I get linker errors

mlw (approved):

PR #425, opened 2025-05-31
Keychain wrapper implementation

mlw:
This adds objects for Santa to manage store/get/delete operations for keychain items.

The `KeychainManager` is the main class. This will typically be used like a singleton. For Santa's purposes, it will be injected when needed. Consumers of the `KeychainManager` can create `KeychainItem`s for themselves. The `KeychainItem` object then allows users to perform keychain operations for that item.

### DIFF
--- a/Source/common/Keychain.h
+++ b/Source/common/Keychain.h
@@ -1,0 +1,81 @@
+/// Copyright 2025 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     https://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#ifndef SANTA__COMMON__KEYCHAIN_H
+#define SANTA__COMMON__KEYCHAIN_H
+
+#include <Foundation/Foundation.h>
+#include <Security/Security.h>
+
+#include <memory>
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+
+namespace santa {
+
+namespace keychain_utils {
+
+// Validation helpers with semi-arbitrary length checks.
+bool IsValidServiceName(NSString *service);
+bool IsValidAccountName(NSString *account);
+bool IsValidDescription(NSString *description);
+
+absl::Status SecurityOSStatusToAbslStatus(OSStatus status);
+
+}  // namespace keychain_utils
+
+class KeychainItem;
+
+class KeychainManager {
+ public:
+  static std::unique_ptr<KeychainManager> Create(NSString *service, SecPreferencesDomain domain);
+  KeychainManager(NSString *service, SecKeychainRef keychain_ref);
+  ~KeychainManager();
+
+  KeychainManager(KeychainManager &&other);
+  KeychainManager &operator=(KeychainManager &&other);
+
+  // Could be safe to implement, but not currently needed
+  KeychainManager(KeychainManager &other) = delete;
+  KeychainManager &operator=(KeychainManager &other) = delete;
+
+  std::unique_ptr<KeychainItem> CreateItem(NSString *account, NSString *description);
+
+ private:
+  NSString *service_;
+  SecKeychainRef keychain_;
+};
+
+class KeychainItem {
+ public:
+  // Note: The given keychain is retained
+  KeychainItem(NSString *service, NSString *account, NSString *description,
+               SecKeychainRef keychain);
+  ~KeychainItem();
+
+  absl::Status Store(NSData *data);
+  absl::Status Delete();
+  absl::StatusOr<NSData *> Get();
+
+ private:
+  NSString *service_;
+  NSString *account_;
+  NSString *description_;
+  SecKeychainRef keychain_;
+};
+
+}  // namespace santa
+
+#endif  // SANTA__COMMON__KEYCHAIN_H
--- a/Source/common/Keychain.h
+++ b/Source/common/Keychain.h
@@ -15,8 +15,8 @@
 #ifndef SANTA__COMMON__KEYCHAIN_H
 #define SANTA__COMMON__KEYCHAIN_H
 
-#include <Foundation/Foundation.h>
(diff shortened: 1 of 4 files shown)


Source/common/Keychain.h
```diff
@@ -0,0 +1,81 @@
...
+#include <Security/Security.h>
+
+#include <memory>
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+
+namespace santa {
+
+namespace keychain_utils {
```
russellhancox, line 28: Is this our first nested namespace? I can't see any others (after a very quick search on my phone). Is there a reason this is in a new namespace?

Source/common/Keychain.h
```diff
@@ -0,0 +1,81 @@
...
+// Validation helpers with semi-arbitrary length checks.
+bool IsValidServiceName(NSString *service);
+bool IsValidAccountName(NSString *account);
+bool IsValidDescription(NSString *description);
+
+absl::Status SecurityOSStatusToAbslStatus(OSStatus status);
+
+}  // namespace keychain_utils
+
+class KeychainItem;
```
russellhancox, line 39: Why not declare KeychainItem first?

Source/common/Keychain.mm
```diff
@@ -0,0 +1,217 @@
...
+    return absl::ErrnoToStatus(EINVAL, "No data to store");
+  }
+
+  if (auto status = Delete(); !status.ok()) {
+    LOGE(@"Failed to remove previous value. %s", status.message().data());
+    return status;
+  }
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
russellhancox, line 156: Which of these are deprecated? Do we really need the deprecated ones?

Source/common/Keychain.mm
```diff
@@ -0,0 +1,217 @@
...
+
+std::unique_ptr<KeychainManager> KeychainManager::Create(NSString *service,
+                                                         SecPreferencesDomain domain) {
+  if (!keychain_utils::IsValidServiceName(service)) {
+    return nullptr;
+  }
+
+  SecKeychainRef keychain = NULL;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
russellhancox, line 73: Do we need to worry about search domains or specific keychains?

Source/common/Keychain.mm
mlw: `kSecAttrAccessibleAlwaysThisDeviceOnly` is "deprecated", but mainly seems to be in the sense to communicate that it shouldn't be used outside of launch daemons as far as I can tell. We don't have much of an option here.

Source/common/Keychain.mm
mlw: As far as I've been able to find, the keychain you get by default is whatever comes first in the search list that gets automatically assigned. But this is OS version specific, and no docs I found describe the behavior as something that can be relied upon as API. So doing nothing might work just fine? It also might not if some future OS changes default search list parameters? Hard to tell... If things did change, I guess the worst case would be that we wouldn't be able to retrieve an export config and would have to wait to sync.

One important note here - in a test command line tool running as root, the default search list is first the user's login keychain, then the system keychain. Running as a daemon though, it seems like the system keychain is specified twice...? (Not sure why I get two of the same thing in the search list.).

So given the unknown/undocumented nature, and also since programs running as root will seemingly search the user's keychain, I didn't want there to be some case where the non-system keychain may get searched.

Apple wants everyone to move to the fancy new data protection keychain and largely deprecated most APIs to interact with the file based keychains. However the docs explicitly say that launch daemons cannot use the data protection keychain and must use the file based one. So we will hit a lot of deprecated APIs for most all usage.

Source/common/Keychain.h
mlw: It might be the first. I think it makes a lot of sense to add this and moving forward we should do this more often. I don't have a good reason why I haven't done it previously... These methods are fairly generic sounding and felt better having them scoped within a namespace that better attributed their functionality. Style guide seems to encourage the use of nested namespaces when appropriate. Reckon the alternative is to name these differently:
```
bool IsValidKeychainItemServiceName(NSString *service);
bool IsValidKeychainItemAccountName(NSString *account);
bool IsValidKeychainItemDescription(NSString *description);
```

Source/common/Keychain.h
mlw: Personal preference. When reading headers it makes more sense to me to see the "top level" objects you interact with first. Having KeychainItem first feels off to me since it's not something you construct manually.

Source/common/Keychain.mm
mlw: Another option here that might work well for us is creating our own keychain. We could also then add it to our tamper protection. This would keep any of our stuff separate from the system keychain. (note this also would require using deprecated APIs)

Source/common/Keychain.mm
mlw: I moved the pragma to surround only the one line item.

Source/common/Keychain.h
russellhancox: I'm fine with us using namespaces more where it make sense to, my only concern is being able to easily tie namespaces back to files. It might, for example, make sense in this case to have the namespace be `santa::common::keychain`.

Source/common/Keychain.mm
russellhancox: That explanation satisfies me. I'll leave it up to you whether we use our own keychain - it would be kind of nice to keep our items separate from the system stuff but I'm not _sure_ it's worth it.

russellhancox (dismissed):

Source/common/Keychain.h
mlw: I'm a bit torn. I kinda like that, but it gets verbose, fast. It also used to be how Santa did namespaces, but was changed to make things simpler:

https://github.com/google/santa/pull/1384
https://github.com/google/santa/pull/1385
https://github.com/google/santa/pull/1386
https://github.com/google/santa/pull/1387

This also reflected changes to the Google style guide where having nested namespaces match the file path fell out of favor.

I could see `santa::keychain` kinda making sense here, but would be odd to not include the KeychainManager and KeychainItem classes in that scope...

Source/common/Keychain.mm
mlw: Will get this PR in without it and think on it. Creating a new keychain or opening an existing keychain would be pretty easy to do. And once we have the reference, the rest of the code works as-is. Will think on it and might add it when adopting the keychain in the next PR.

Source/common/Keychain.h
mlw: Reworked based on offline discussion.

PR #429, opened 2025-06-04
santactl/fileinfo: Add --utc flag

russellhancox:


pmarkowsky (dismissed):
LGTM.

Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -193,6 +193,7 @@ + (NSString *)longHelpText {
           @"                    -1 down to -n-1 for the root certificate down to the leaf\n"
           @"                  Incompatible with --bundleinfo."
           @"\n"
+          @"    --utc: Use UTC timestamps for all dates.\n"
```
mlw: Could we make UTC default, change the flag to `--local`?

russellhancox: Good call, done, but used `--localtz`

mlw (approved):

PR #434, opened 2025-06-06
santad: Integrate CEL rule processing into execution path

russellhancox:
This change allows Santa to process CEL rules as part of execution authorization. It purposefully does not add any method to get CEL rules _into_ the database, while we validate that everything works both correctly and performantly.


Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
```diff
@@ -190,12 +190,24 @@ - (void)handleMessage:(Message &&)esMsg
...
+      // Do not allow caching when the action is SNTActionRespondHold because Santa
+      // also authorizes EXECs that occur while the current authorization is pending.
+      cacheable = false;
+      OS_FALLTHROUGH;
+    case SNTActionRespondAllowNoCache:
+      // Do not cache when the action is SNTActionRespondAllowNoCache because this
+      // implies a CEL evaluation was performed and used per-process data to make
+      // its decision.
+      cacheable = false;
+      OS_FALLTHROUGH;
```
mlw: The fallthroughs here felt complicated to follow when broken up by the comments and I worry it's prone to accidental regression in the future. Let's just set `cacheable` and `authResult`,  and then `break` out of each case separately.

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
```diff
@@ -190,12 +190,24 @@ - (void)handleMessage:(Message &&)esMsg
...
 
   switch (action) {
     case SNTActionRespondAllowCompiler:
+      // Do not cache compiler processes so future instances get marked appropriately.
       [self.compilerController setProcess:esMsg->event.exec.target->audit_token isCompiler:true];
       OS_FALLTHROUGH;
-    case SNTActionRespondHold: OS_FALLTHROUGH;
+    case SNTActionRespondHold:
+      // Do not allow caching when the action is SNTActionRespondHold because Santa
+      // also authorizes EXECs that occur while the current authorization is pending.
```
mlw: I found the comment a bit confusing... I expanded upon it a bit to give more detail, maybe something instead like:
```
Do not allow caching when the action is SNTActionRespondHold. Santa will
allow the current exec because the process was first suspended so no code
will be executed. However if additional instances of the binary are executed
while authorization is pending, neither Santa nor ES should use the decision
as it is not yet final.
```

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -85,18 +88,35 @@ struct RuleIdentifiers CreateRuleIDs(SNTCachedDecision *cd) {
...
-- (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
+- (instancetype)init {
   self = [super init];
   if (self) {
-    _ruleTable = ruleTable;
     _configurator = [SNTConfigurator configurator];
+
+    auto evaluator = santa::cel::Evaluator::Create();
+    if (evaluator.ok()) {
+      celEvaluator_ = std::move(evaluator.value());
```
mlw: nit: `*evaluator`

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -106,7 +126,32 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
 // It returns YES if the decision was made, NO if the decision was not made.
 - (BOOL)decision:(SNTCachedDecision *)cd
                 forRule:(SNTRule *)rule
-    withTransitiveRules:(BOOL)enableTransitiveRules {
+    withTransitiveRules:(BOOL)enableTransitiveRules
+       andCELActivation:(santa::cel::Activation *)activation {
```
mlw: This method is receiving the result of the call to the `activationCallback()`. Instead, we should pass the block through to here, and then only call it in the `if (rule.state == SNTRuleStateCEL) { ... }` section below. This way the block wouldn't have to be called and create activations for rule states where they never need to be realized.

Source/santad/SNTExecutionController.mm
```diff
@@ -628,4 +638,36 @@ - (void)createRuleForStandaloneModeEvent:(SNTStoredEvent *)se {
...
 }
 
+- (ActivationCallbackBlock)createActivationForMessage:(const santa::Message &)esMsg
+                                            andCSInfo:(nullable MOLCodesignChecker *)csInfo
+                                       preventCaching:(nonnull BOOL *)preventCaching {
+  return ^santa::cel::Activation *() {
+    auto f = std::make_unique<::santa::cel::v1::ExecutableFile>();
+    f->mutable_signing_timestamp()->set_seconds(csInfo.signingTime.timeIntervalSince1970);
+
+    auto activation = new santa::cel::Activation(
```
mlw: nit: no need for intermediate variable, just `return new santa::cel::Activation...`

Source/santad/EventProviders/AuthResultCache.mm
```diff
@@ -158,6 +159,15 @@ static inline uint64_t TimestampFromCachedValue(uint64_t cachedValue) {
     return SNTActionUnset;
   }
 
+  // If the cached value is SNTActionRespondAllowNoCache, remove the cached
+  // value and return. We needed to add the cached value to allow any threads
+  // waiting to see a value but we don't want to let them use the cached value
+  // because the dynamic values might be different.
```
mlw: I'm not sure I follow this comment as to why adding `SNTActionRespondAllowNoCache` to the cache was necessary. Any thread that was waiting on a value was doing so because of `SNTActionRequestBinary`.

Could we instead change `AuthResultCache::AddToCache` to remove the entry when `SNTActionRespondAllowNoCache` is given, instead of adding then removing?

Source/santad/SNTExecutionController.mm
```diff
@@ -628,4 +638,36 @@ - (void)createRuleForStandaloneModeEvent:(SNTStoredEvent *)se {
...
+        f.get(),
+        ^std::vector<std::string>() {
+          *preventCaching = YES;
+          std::vector<std::string> args;
+          for (int i = 0; i < es_exec_arg_count(&esMsg->event.exec); i++) {
```
mlw: We should use the ESAPI wrappers instead of calling the these functions directly, kinda like we do here: https://github.com/northpolesec/santa/blob/de016d6dd9f5153dfc209ad2b8dd1e7a7bea2fb3/Source/santad/EventProviders/SNTEndpointSecurityTamperResistance.mm#L91-L98

Source/santad/SNTExecutionController.mm
```diff
@@ -628,4 +638,36 @@ - (void)createRuleForStandaloneModeEvent:(SNTStoredEvent *)se {
   // TODO: Notify the sync service of the new rule.
 }
 
+- (ActivationCallbackBlock)createActivationForMessage:(const santa::Message &)esMsg
```
mlw: Would like to chat about `esMsg` lifetime here to make sure we're ok.

Source/santad/SNTExecutionController.mm
```diff
@@ -270,10 +272,15 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
   // TODO(markowsky): Maybe add a metric here for how many large executables we're seeing.
   // if (binInfo.fileSize > SomeUpperLimit) ...
 
+  __block BOOL preventCaching = NO;
```
mlw: Would like to chat about options here if there's any way to restructure things to make how this gets set a bit more direct.

Source/santad/SNTExecutionController.mm
```diff
@@ -628,4 +638,36 @@ - (void)createRuleForStandaloneModeEvent:(SNTStoredEvent *)se {
   // TODO: Notify the sync service of the new rule.
 }
 
+- (ActivationCallbackBlock)createActivationForMessage:(const santa::Message &)esMsg
+                                            andCSInfo:(nullable MOLCodesignChecker *)csInfo
+                                       preventCaching:(nonnull BOOL *)preventCaching {
+  return ^santa::cel::Activation *() {
+    auto f = std::make_unique<::santa::cel::v1::ExecutableFile>();
+    f->mutable_signing_timestamp()->set_seconds(csInfo.signingTime.timeIntervalSince1970);
```
mlw: We should talk about how we want to present unsigned code. I reckon right now this will just get set to 0 (?). But I really would like to avoid in band signaling like this.

Source/santad/SNTPolicyProcessor.mm
russellhancox: I meant to do this but got distracted and then forgot about it.

Source/santad/EventProviders/AuthResultCache.mm
russellhancox: This was the piece I was most unhappy about. I've changed the code instead to prevent adding `SNTActionRespondAllowNoCache` to the cache, removed this condition in `CheckCache` and updated the code in SNTEndpointSecurityAuthorizer to remove from the cache at the same time that it responds with the allow.

Source/santad/SNTExecutionController.mm
russellhancox: Done, but to make things easier I've also added wrappers to make a std::vector of args and a std::map of envs.

Source/santad/EventProviders/EndpointSecurity/EndpointSecurityAPI.mm
```diff
@@ -138,6 +147,16 @@
   return es_exec_env(event, index);
 }
 
+std::map<std::string, std::string> EndpointSecurityAPI::ExecEnvs(const es_event_exec_t *event) {
+  std::map<std::string, std::string> envs;
+  for (uint32_t i = 0; i < es_exec_env_count(event); i++) {
+    auto s = santa::StringTokenToStringView(es_exec_env(event, i));
+    auto npos = s.find("=");
```
mlw, line 154: While I'm fairly sure this will always work, I think it'd be good to check `npos != std::string::npos` to prevent throwing an exception. Then I guess we'd either toss the value, or put some SANT_ENV_VAL_MISSING_PLACEHOLDER value or something...

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -125,10 +125,11 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
 // It returns YES if the decision was made, NO if the decision was not made.
 - (BOOL)decision:(SNTCachedDecision *)cd
-                forRule:(SNTRule *)rule
-    withTransitiveRules:(BOOL)enableTransitiveRules
-       andCELActivation:(santa::cel::Activation *)activation {
+                     forRule:(SNTRule *)rule
+         withTransitiveRules:(BOOL)enableTransitiveRules
+    andCELActivationCallback:(ActivationCallbackBlock)activationCallback {
   if (rule.state == SNTRuleStateCEL) {
+    auto activation = activationCallback ? activationCallback() : nullptr;
```
mlw: Can a null activation ever be valid? Could we skip calling `CompileAndEvaluate` if null?

Source/common/SNTCachedDecision.h
```diff
@@ -54,4 +54,6 @@
 @property NSString *customURL;
 @property BOOL silentBlock;
 
+@property BOOL cacheable;
```
mlw, line 57: nit: Would prefer a name more specific to this use case since this doesn't mean generally cacheable

Source/common/cel/cel.proto
```diff
@@ -59,5 +59,16 @@ message ExecutionContext {
...
+  // This timestamp is provided by the developer when this binary was signed.
+  // It is possible for developers to set this to whatever they wish when signing
+  // unlike the secure_signing_timestamp.
+  optional google.protobuf.Timestamp signing_timestamp = 1;
+
+  // The timestamp of when the binary was signed.
+  //
+  // This timestamp is the secure timestamp that was certified by Apple's timestamp
+  // authority service and can be trusted.
+  optional google.protobuf.Timestamp secure_signing_timestamp = 2;
```
mlw: Could we change to `signing_time` and `secure_signing_time` to match the sync protocol names?

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
```diff
@@ -190,12 +190,32 @@ - (void)handleMessage:(Message &&)esMsg
...
+      cacheable = false;
+      authResult = ES_AUTH_RESULT_ALLOW;
+      break;
+    case SNTActionRespondAllowNoCache:
+      // Do not cache when the action is SNTActionRespondAllowNoCache because this
+      // implies a CEL evaluation was performed and used per-process data to make
+      // its decision.
+      cacheable = false;
+      authResult = ES_AUTH_RESULT_ALLOW;
+      self->_authResultCache->RemoveFromCache(esMsg->event.exec.target->executable);
```
mlw, line 217: Does this work? We call `AddToCache` down below. The default case there should throw an exception.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -418,31 +473,34 @@ - (nonnull SNTCachedDecision *)decisionForFilePath:(nonnull NSString *)filePath
...
+          }
+        } else {
+          if (error.code == errSecCSUnsigned) {
+            return SNTSigningStatusUnsigned;
+          } else {
+            return SNTSigningStatusInvalid;
+          }
+        }
+      }
+      activationCallback:^santa::cel::Activation *() {
```
mlw: Can this just be `nil` now

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -105,8 +125,38 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable {
...
+    auto activation = activationCallback ? activationCallback() : nullptr;
+    auto evalResult = self->celEvaluator_->CompileAndEvaluate(
+        santa::NSStringToUTF8StringView(rule.celExpr), *activation);
+    if (!evalResult.ok()) {
+      LOGE(@"Failed to evaluate CEL rule (%@): %s", rule.celExpr,
+           std::string(evalResult.status().message()).c_str());
+      return NO;
+    }
+
+    ::santa::cel::v1::ReturnValue returnValue = (*evalResult).first;
```
mlw: nit: `evalResult->first` and `evalResult->second`

Source/common/SNTCachedDecision.h
russellhancox: It's difficult to come up with a better name; while CEL rules be the only ones currently setting this to NO (with the default as YES), it's possible others could use this in the future, and the place where the value is read is not CEL specific

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
russellhancox: It seemed to when I tested, but I'm not sure why. I've updated anyway

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -128,8 +128,8 @@ - (BOOL)decision:(SNTCachedDecision *)cd
                      forRule:(SNTRule *)rule
          withTransitiveRules:(BOOL)enableTransitiveRules
     andCELActivationCallback:(ActivationCallbackBlock)activationCallback {
-  if (rule.state == SNTRuleStateCEL) {
-    auto activation = activationCallback ? activationCallback() : nullptr;
+  if (rule.state == SNTRuleStateCEL && activationCallback) {
```
mlw, line 131: Just confirming - the plan is to better handle the `santactl fileinfo` path in a future PR, yeah?

mlw (approved):
This is awesome, excited to get this in.

Source/santad/SNTPolicyProcessor.mm
russellhancox: Yeah. I need to think about how to make it most useful, the simple option would be to just show that a CEL rule is in effect and not to try and show a definitive response. Another option would be to have a way for fileinfo to provide the dynamic data for an Activation but that's a lot to maintain.

PR #435, opened 2025-06-11
Adopt latest AWS export config

mlw:


Source/common/SNTExportConfiguration.h
```diff
@@ -26,8 +26,16 @@ typedef NS_ENUM(NSInteger, SNTExportConfigurationType) {
 
 /// Configuration required for exporting to AWS
 @interface SNTExportConfigurationAWS : NSObject <SNTExportConfigurationProtocol>
-@property(readonly) NSData *token;
-- (instancetype)initWithToken:(NSData *)token;
+@property(readonly) NSString *accessKey;
+@property(readonly) NSString *secretAccessKey;
+@property(readonly) NSString *sessionToken;
+@property(readonly) NSString *bucketName;
+@property(readonly) NSString *objectKeyPrefix;
```
russellhancox, line 33: I'm mildly surprised this works. It's fine to declare a property readonly in the header but it should be re-declared as `readwrite, copy` in a class extension in the implementation file so that the backing ivar is created.

mlw: This only affects the dot notation. Keeping it readonly, no setter is generated. The backing ivar still exists and can be set within the implementation. I really want these to be "const" as there shouldn't be any way to change after initialization, but that's obviously not really supported in ObjC.

PR #436, opened 2025-06-11
santactl/fileinfo: Change how Rule field is populated

russellhancox:
Instead of showing the _decision_ that would (probably) be made by santad, the `Rule` field now shows which rule matches. This more accurately reflects the field name and allows us to show that a CEL rule exists (which largely precludes being able to actually give a decision). If we wanted, we could later expand on this to also 

This involved removing the (now unneeded) `decisionForFilePath:identifiers:reply:` method from the XPC control interface, and moving the `databaseRuleForIdentifiers:reply:` method from the privileged to the unprivileged interface. I also moved a `stringifyRule:withColor:` method that had been in SNTCommandRule into SNTRule itself so it can easily be used by both commands.

### DIFF
--- a/Source/common/SNTRule.m
+++ b/Source/common/SNTRule.m
@@ -408,6 +408,85 @@ static const NSUInteger kExpectedTeamIDLength = 10;
                        self.identifier, self.state, self.type, (unsigned long)self.timestamp];
 }
 
+- (NSString *)stringifyWithColor:(BOOL)colorize {
+  NSMutableString *output;
+  // Rule state is saved as eventState for output colorization down below
+  SNTEventState eventState = SNTEventStateUnknown;
+
+  switch (self.state) {
+    case SNTRuleStateUnknown:
+      output = [@"None" mutableCopy];
+      break;
+    case SNTRuleStateAllow: OS_FALLTHROUGH;
+    case SNTRuleStateAllowCompiler: OS_FALLTHROUGH;
+    case SNTRuleStateAllowTransitive:
+      output = [@"Allowed" mutableCopy];
+      eventState = SNTEventStateAllow;
+      break;
+    case SNTRuleStateBlock: OS_FALLTHROUGH;
+    case SNTRuleStateSilentBlock:
+      output = [@"Blocked" mutableCopy];
+      eventState = SNTEventStateBlock;
+      break;
+    case SNTRuleStateCEL:
+      output = [@"CEL" mutableCopy];
+      break;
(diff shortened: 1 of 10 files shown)


Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -383,7 +383,6 @@ - (SNTAttributeBlock)rule {
     dispatch_once(&token, ^{
       [cmd.daemonConn resume];
     });
-    __block SNTEventState state;
     dispatch_semaphore_t sema = dispatch_semaphore_create(0);
     NSError *err;
```
tburgin, line 390: If there is a cs error, we shouldn't bother getting attributes from csc.

tburgin (dismissed):
This is a welcome refactor.

Source/santactl/Commands/SNTCommandFileInfo.m
russellhancox: Done, sort of

Source/common/SNTRule.m
```diff
@@ -408,6 +408,81 @@ - (NSString *)description {
...
+      eventState = SNTEventStateBlock;
+      break;
+    case SNTRuleStateCEL: output = [@"CEL" mutableCopy]; break;
+    case SNTRuleStateRemove: OS_FALLTHROUGH;
+    default:
+      output = [NSMutableString stringWithFormat:@"Unexpected rule state: %ld", self.state];
+      break;
+  }
+
+  if (self.state == SNTRuleStateUnknown) {
```
mlw: We should also return on the `default` case as well to match the `SNTRuleStateUnknown` case.

Source/santactl/Commands/SNTCommandFileInfo.m
```diff
@@ -383,77 +385,44 @@ - (SNTAttributeBlock)rule {
...
-    __block SNTEventState state;
     dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+
     NSError *err;
     MOLCodesignChecker *csc = [fileInfo codesignCheckerWithError:&err];
-
-    NSString *cdhash = csc.cdhash;
-    NSString *teamID = csc.teamID;
-    NSString *signingID = FormatSigningID(csc);
+    SNTSigningStatus signingStatus = SNTSigningStatusInvalid;
```
mlw: Can we move this to CertificateHelpers and handle all possible statuses?

mlw (changes requested):

PR #441, opened 2025-06-14
Add Defer implementation

mlw:
Adds a Defer object to run code on destruction. The primary purpose is to better ensure cleanup operations (closing files, freeing memory, etc.) are performed in functions that have many return paths.


mlw:
TIL: `absl::Cleanup`. Does most of what we need so going to use that instead of adding more code to our repo

PR #444, opened 2025-06-16
Add glob support to paths for Proc FAA rules

### LINKED ISSUE #442: Clarification on how FAA works
I tried inserting the following section as the main part of Santa’s .mobileconfig file. The configuration loads successfully (confirmed via santactl status), but /bin/ls can still list directories in the restricted paths without being blocked. All of these are in MONITOR mode.

Since it is difficult to figure this out from the documentation: 

Is this expected?
How can I enforce actual file access blocking without switching to LOCKDOWN mode?

```xml
<key>mcx_preference_settings</key>
<dict>
    <key>FileAccessPolicy</key>
    <dict>
        <!-- Edited -->
        <key>Version</key>
        <string>v0.1</string>
        <key>WatchItems</key>
        <dict>
            <key>LS</key>
            <dict>
                <key>Paths</key>
                <array>
                    <dict>
                        <key>Path</key>
                        <string>/Users/*</string>
                        <key>IsPrefix</key>
                        <true/>
                    </dict>
                </array>
                <key>Options</key>
                <dict>
                    <key>RuleType</key>
                    <string>ProcessesWithDeniedPaths</string>
                    <key>AuditOnly</key>
                    <false/>
                    <key>InvertProcessExceptions</key>
                    <false/>
                </dict>
                <key>Processes</key>
                <array>
                    <dict>
                        <key>BinaryPath</key>
                        <string>/bin/ls</string>
                    </dict>
                </array>
            </dict>
        </dict>
    </dict>
    <key>ClientMode</key>
    <integer>1</integer>
    <key>EnableSilentMode</key>
    <false/>
</dict>
```

mlw:
Fixes #442

### DIFF
--- a/Source/common/Glob.mm
+++ b/Source/common/Glob.mm
@@ -42,7 +42,7 @@ void FindMatches(NSString *base, NSMutableArray<NSString *> *path_components, NS
 
   int err = glob(path.UTF8String, GLOB_NOSORT, nullptr, g);
   if (err != 0 && err != GLOB_NOMATCH) {
-    LOGE(@"Failed to generate path names for watch item: %@", path);
+    LOGE(@"Failed to generate path names from glob: %@", path);
     return;
   }
 
@@ -95,7 +95,7 @@ std::vector<std::string> FindMatches(NSString *path) {
 
   int err = glob(path.UTF8String, 0, nullptr, g);
   if (err != 0 && err != GLOB_NOMATCH) {
-    LOGE(@"Failed to generate path names for watch item: %@", path);
+    LOGE(@"Failed to generate path names from glob: %@", path);
     return {};
   }
 
@@ -111,7 +111,7 @@ std::vector<std::string> FindMatches(NSString *path) {
   // Semi-arbitrary to prevent run away recursion. We could consider increasing this if
   // anyone ever has a good use case.
   if (path_components.count > 40) {
-    LOGW(@"Watch path contained too many components, skipping: %@", path);
+    LOGW(@"Glob path contained too many components, skipping: %@", path);
     return {};
   }
 


Source/common/Glob.mm
```diff
@@ -0,0 +1,137 @@
...
+  g->gl_pathv = NULL;
+
+  // Ensure globfree is always called
+  absl::Cleanup glob_cleaup = ^{
+    globfree(g);
+  };
+
+  int err = glob(path.UTF8String, GLOB_NOSORT, nullptr, g);
+  if (err != 0 && err != GLOB_NOMATCH) {
+    LOGE(@"Failed to generate path names for watch item: %@", path);
```
russellhancox: "watch item" isn't as generic as this file seems to be. Perhaps "Failed to generate path names from glob"

Source/common/Glob.mm
```diff
@@ -0,0 +1,137 @@
...
+  g->gl_pathv = NULL;
+
+  // Ensure globfree is always called
+  absl::Cleanup glob_cleaup = ^{
+    globfree(g);
+  };
+
+  int err = glob(path.UTF8String, 0, nullptr, g);
+  if (err != 0 && err != GLOB_NOMATCH) {
+    LOGE(@"Failed to generate path names for watch item: %@", path);
```
russellhancox: Sam here wrt watch item

Source/common/Glob.mm
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  NSArray<NSString *> *path_components = [path pathComponents];
+  // Code above enforces the given path starts with a /, so must be at least two entries
+  assert(path_components.count > 1);
+
+  // Semi-arbitrary to prevent run away recursion. We could consider increasing this if
+  // anyone ever has a good use case.
+  if (path_components.count > 40) {
+    LOGW(@"Watch path contained too many components, skipping: %@", path);
```
russellhancox: And here

PR #448, opened 2025-06-18
santactl/doctor: Add checks for syncing, update MOLAuth logs

russellhancox:
Most of the useful changes are really in MOLAuthenticatingURLSession to print more useful information. Now that `santactl sync` doesn't print these logs unless `--debug` is provided I think the extra logging is well worth it to make diagnosing issues easier.

`santactl doctor` will check a few basic things and make an empty preflight with an invalid UUID. If this request is successfully made, even if it returns a bad request error etc., will tell us that the authentication for syncing should work.

### DIFF
--- a/Source/common/MOLAuthenticatingURLSession.m
+++ b/Source/common/MOLAuthenticatingURLSession.m
@@ -218,7 +218,8 @@
     allCerts = [MOLCertificate certificatesFromArray:results];
 
     if (self.clientCertCommonName) {
-      [self log:@"[Client Trust] Looking for certificate with common name: %@", self.clientCertCommonName];
+      [self log:@"[Client Trust] Looking for certificate with common name: %@",
+                self.clientCertCommonName];
       foundIdentity = [self identityByFilteringArray:allCerts
                                           commonName:self.clientCertCommonName
                                     issuerCommonName:nil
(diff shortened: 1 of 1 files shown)


Source/santactl/Commands/SNTCommandDoctor.mm
```diff
@@ -137,4 +142,80 @@ - (BOOL)validateConfiguration {
...
+  authURLSession.userAgent = @"santactl-sync/";
+  NSString *santactlVersion = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"];
+  if (santactlVersion) {
+    authURLSession.userAgent = [authURLSession.userAgent stringByAppendingString:santactlVersion];
+  }
+  authURLSession.refusesRedirects = YES;
+  authURLSession.serverHostname = syncBaseURL.host;
+  authURLSession.loggingBlock = ^(NSString *line) {
+    print(@"%s", [line UTF8String]);
+  };
```
mlw, line 179: Consider abstracting this setup. Being duplicated here is a little brittle, not that it's ever changed... Not a blocker.

russellhancox: Ack, will look into fixing this after the June release is out

PR #453, opened 2025-06-20
santactl/sync: Add ability to ingest CEL rules.

russellhancox:
This covers `santactl rule`, syncing, and StaticRules.

### DIFF
--- a/Source/santactl/Commands/SNTCommandRule.m
+++ b/Source/santactl/Commands/SNTCommandRule.m
@@ -54,6 +54,7 @@ REGISTER_COMMAND_NAME(@"rule")
           @"    --block: add to block\n"
           @"    --silent-block: add to silent block\n"
           @"    --compiler: allow and mark as a compiler\n"
+          @"    --cel {cel_expr}: add a CEL rule\n"
           @"    --remove: remove existing rule\n"
           @"    --check: check for an existing rule\n"
           @"    --import {path}: import rules from a JSON file\n"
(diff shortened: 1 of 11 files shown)


Source/santasyncservice/SNTSyncRuleDownload.mm
```diff
@@ -42,7 +43,17 @@ SNTRuleCleanup SyncTypeToRuleCleanup(SNTSyncType syncType) {
...
 
-@implementation SNTSyncRuleDownload
+@implementation SNTSyncRuleDownload {
+  std::unique_ptr<santa::cel::Evaluator> _celEvaluator;
+}
+
+- (nullable instancetype)initWithState:(nonnull SNTSyncState *)syncState {
+  self = [super initWithState:syncState];
+  if (self) {
+    _celEvaluator = santa::cel::Evaluator::Create().value();
```
mlw: This feels a bit scary since it will throw if no value. But it seems like maybe there isn't much else for us to do. It looks like failures in `Create` would either be transitive, or always fail. Biggest concern is a crash loop, but that feels like something that would be caught quickly since it would mean the sync service never completes rule download...

Do we need an evaluator per rule download instance? Or could we instead create one for the lifetime of the sync service? If we create one for the lifetime, we could make that a requirement for loading which would move any failures forward and be more easily discovered via our testing.

Source/santactl/Commands/SNTCommandRule.m
```diff
@@ -54,6 +54,7 @@ + (NSString *)longHelpText {
           @"    --block: add to block\n"
           @"    --silent-block: add to silent block\n"
           @"    --compiler: allow and mark as a compiler\n"
+          @"    --cel {cel_expr}: add a CEL rule\n"
```
pmarkowsky, line 57: nit: we probably need to add an explanation or link to CEL documentation to the usage string so people know that how to use args, target and CEL operators.

Source/santasyncservice/SNTSyncRuleDownload.mm
pmarkowsky: Shouldn't this just be the normal StatusOr check? e.g.

```C++
auto result = santa::cel::Evaluator::Create();
if (!result.ok()) {
  // Treat this as if initWithState failed, explicitly abort or assign it to NULL 
  // and check later (which we already do).
  _celEvaluator = nullptr;
} else {
  _celEvaluator = result.value();
}
```

or the inverse assuming we're initializing to NULL/nullptr already.

```C++
auto result = santa::cel::Evaluator::Create();
if (result.ok()) {
  _celEvaluator = result.value();
} 
```

russellhancox: > Do we need an evaluator per rule download instance? Or could we instead create one for the lifetime of the sync service?

No, creating at process start and passing it into stages via SNTSyncState is reasonable and what I've gone for. That did involve moving most of the remaining ObjC classes to ObjC++.

> If we create one for the lifetime, we could make that a requirement for loading which would move any failures forward and be more easily discovered via our testing.

I'm leaving the evaluator as a nullable pointer with a check around it, I don't think we should break syncing if we can't create an evaluator.

Source/santactl/Commands/SNTCommandRule.m
russellhancox: Done

PR #454, opened 2025-06-20
sync: add support for primary user groups

tburgin:
If defined, primary user groups are uploaded during preflight.

### DIFF
--- a/Source/common/SNTConfigurator.m
+++ b/Source/common/SNTConfigurator.m
@@ -95,9 +95,11 @@ static NSString *const kEnableStatsCollectionKey = @"EnableStatsCollection";
 static NSString *const kStatsOrganizationID = @"StatsOrganizationID";
 
 static NSString *const kMachineOwnerKey = @"MachineOwner";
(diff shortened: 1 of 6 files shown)


Source/common/SNTConfigurator.m
```diff
@@ -1039,6 +1047,26 @@ - (NSString *)machineOwner {
   return machineOwner ?: @"";
 }
 
+- (NSArray<NSString *> *)machineOwnerGroups {
+  NSArray<NSString *> *machineOwnerGroups = self.configState[kMachineOwnerGroupsKey];
+  if (machineOwnerGroups.count) return machineOwnerGroups;
+
+  NSString *plistPath = self.configState[kMachineOwnerPlistFileKey];
+  NSString *plistKey = self.configState[kMachineOwnerGroupsPlistKeyKey];
+  if (plistPath && plistKey) {
```
russellhancox:
```suggestion
  if (plistPath.length && plistKey.length) {
```

tburgin: Done, and also in -machineOwner

mlw (approved):

PR #456, opened 2025-06-23
docs: Add basic docs for CEL policy

russellhancox:


docs/docs/features/binary-authorization.md
```diff
@@ -277,6 +277,43 @@ informing the user that it was Santa that did it can be a _very_ confusing
 experience for users and lead to wasted time trying to determine the underlying
 cause.
 
+#### CEL
```
pmarkowsky, line 280: We should expand the acronym here. 

```suggestion
#### Common Expression Language (CEL)
```

docs/docs/features/binary-authorization.md
```diff
@@ -277,6 +277,43 @@ informing the user that it was Santa that did it can be a _very_ confusing
...
+[CEL expression](https://cel.dev/) which will be evaluated as part of making a
+decision.
+
+The input to the expression will be an
+[santa.cel.v1.ExecutionContext](https://github.com/northpolesec/protos/blob/704246489aa55e6e2b60b47133a8668bc3656105/cel/v1.proto#L42)
+and the return value must either be a
+[santa.cel.v1.ReturnValue](https://github.com/northpolesec/protos/blob/704246489aa55e6e2b60b47133a8668bc3656105/cel/v1.proto#L20)
+or a bool. If the return value is a bool, true will be treated as a
+ReturnValue.ALLOWLIST and false will be treated as ReturnValue.BLOCKLIST.
+
```
pmarkowsky, line 295: We should also call out `target` and it's attributes along with the the helper functions.

docs/docs/features/binary-authorization.md
pmarkowsky, line 1: Do we need to update the diagram to have something that's show the policy evaluation order?

mlw (dismissed):

docs/docs/features/binary-authorization.md
russellhancox: Done, except not in the title, because it ruins the sidebar TOC.

docs/docs/features/binary-authorization.md
russellhancox: There is no such thing as policy evaluation order

docs/docs/features/binary-authorization.md
russellhancox: I linked to the ExecutionContext proto, target is one of the fields in the proto. We don't have any helper functions

pmarkowsky: The args and env in the context/environment are not part of the CEL standard. So just saying we should call those out.

russellhancox: There is no context in the CEL standard. Our input to the expression is the content of the ExecutionContext proto (which contains target, args and envs).

mlw (approved):

docs/docs/features/binary-authorization.md
pmarkowsky: It was admittedly a bad comment. The point I was trying to make was that there's an order of match first on the identifier then checking the CEL policy.

PR #469, opened 2025-06-26
santad: Check CEL expression validity before adding to DB

russellhancox:
Move CEL expression validation from SNTRuleDownload -> SNTRuleTable so that it applies to both `santactl rule` and rule download. This doesn't cover static rules but that will be coming in a future PR.

Output from rule download failures:
```
Failed to add rule(s) to database: Rule array contained rule with invalid CEL expression
Failure reason: ERROR: <input>:1:1: undeclared reference to 'undefined_field' (in container '')
 | undefined_field
 | ^
Rule download failed, aborting run
```

Output from santactl sync failures:
```
» sudo santactl rule --force --signingid --identifier platform:com.apple.say --cel 'test'
Failed to modify rules: ERROR: <input>:1:1: undeclared reference to 'test' (in container '')
 | test
 | ^
```

### DIFF
--- a/Source/santad/DataLayer/SNTRuleTable.mm
+++ b/Source/santad/DataLayer/SNTRuleTable.mm
@@ -348,8 +348,8 @@ static void addPathsFromDefaultMuteSet(NSMutableSet *criticalPaths) {
 
 - (SNTRule *)ruleFromResultSet:(FMResultSet *)rs {
   return [[SNTRule alloc] initWithIdentifier:[rs stringForColumn:@"identifier"]
-                                       state:(SNTRuleState)[rs intForColumn:@"state"]
-                                        type:(SNTRuleType)[rs intForColumn:@"type"]
+                                       state:static_cast<SNTRuleState>([rs intForColumn:@"state"])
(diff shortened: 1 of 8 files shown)


Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -337,8 +348,8 @@ - (int64_t)cdhashRuleCount {
 
 - (SNTRule *)ruleFromResultSet:(FMResultSet *)rs {
   return [[SNTRule alloc] initWithIdentifier:[rs stringForColumn:@"identifier"]
-                                       state:[rs intForColumn:@"state"]
-                                        type:[rs intForColumn:@"type"]
+                                       state:(SNTRuleState)[rs intForColumn:@"state"]
+                                        type:(SNTRuleType)[rs intForColumn:@"type"]
```
mlw:
```
static_cast<SNTRuleState>
```

mlw (changes requested):

mlw (dismissed):

PR #472, opened 2025-06-27
docs: add owner groups

tburgin:
Also add support in the config generator for user input for repeated fields.

https://github.com/user-attachments/assets/811d0be3-a41b-45c4-87fe-486da64e2a4d

### DIFF
--- a/docs/src/lib/santaconfig.ts
+++ b/docs/src/lib/santaconfig.ts
@@ -35,9 +35,6 @@ export type SantaConfigKey = {
   // A fixed set of possible values.
   possibleValues?: SantaPossibleValue[];
 
-  // Set to true enable user input for repeated fields.
-  freeform?: boolean;
-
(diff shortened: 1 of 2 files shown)


docs/src/lib/santaconfig.ts
```diff
@@ -35,6 +35,9 @@ export type SantaConfigKey = {
   // A fixed set of possible values.
   possibleValues?: SantaPossibleValue[];
 
+  // Set to true enable user input for repeated fields.
+  freeform?: boolean;
```
russellhancox: Can't this be inferred from repeated=true and possibleValues=[]

russellhancox (dismissed):

docs/src/lib/santaconfig.ts
tburgin: It can, but I needed to add an additional check of enableIf, so we can still show `This key is not yet supported by the generator` for fields we don't yet support.

PR #473, opened 2025-06-27
GitHub issue templates

mlw:


.github/ISSUE_TEMPLATE/01-feature-request.yml
```diff
@@ -0,0 +1,33 @@
+name: 🚀 Feature Request / Enhancement
+description: Request a new feature or or enhancement
+labels: ["enhancement"]
```
russellhancox, line 3: I don't think we need this label, given that we're using the Feature type. Same for the other templates, except 004

.github/ISSUE_TEMPLATE/01-feature-request.yml
```diff
@@ -0,0 +1,33 @@
...
+type: "Feature"
+assignees: []
```
russellhancox: I think you can leave this out, it's optional. Same for the other templates

.github/ISSUE_TEMPLATE/02-question.yml
```diff
@@ -0,0 +1,34 @@
...
+    id: detailed_question
+    attributes:
+      label: Detailed Question
+      description: Provide more details about your question
+      placeholder: |
+        Please include:
+        - What you're trying to accomplish
+        - What you've already tried (if applicable)
+        - Any specific context or constraints
+        - Why this information would be helpful to you
```
russellhancox: I don't really get what benefit we get from asking this

.github/ISSUE_TEMPLATE/04-faa-cel-use-case.yml
```diff
@@ -0,0 +1,55 @@
+name: 💡 FAA/CEL Cookbook Idea
+description: Submit a use case idea to add to the CEL and FAA cookbook
+labels: ["cookbook", "enhancement"]
```
russellhancox, line 3: Does cookbook already exist?

mlw: No not yet, but reckon we all seem comfortable with it so will go ahead and do that now.

.github/ISSUE_TEMPLATE/01-feature-request.yml
mlw: I like having it, it feels like a nice separation. To me, a bug type or task type can be an enhancement. So if the label were to exist, why wouldn't it apply to features? It's how we've used it before adding types. It also means simpler queries for all enhancements.

.github/ISSUE_TEMPLATE/01-feature-request.yml
```diff
@@ -0,0 +1,32 @@
+name: 🚀 Feature Request / Enhancement
+description: Request a new feature or or enhancement
```
russellhancox: Or or

.github/ISSUE_TEMPLATE/01-feature-request.yml
russellhancox: I'm not sure I follow the logic. Features are getting an enhancement label, Bugs are getting a bug label and Questions are getting a question label, but the type is already signifying this so why also have the label duplicating it? Types can be queried just like labels can, so I don't see what benefit there is to having the label also attached when it's saying the exact same thing

We did use labels this way before we had types, but that was out of necessity

mlw: I guess personally I think types are silly and we shouldn't use them. Labels are more expressive. If we say tasks can be enhancements then surely features can be enhancements too? I guess the other way is kinda acceptable - get rid of the enhancement label and no tasks can be an enhancement, they must be reclassified as feature. But this feels worse than just using labels, and requires going thru open issues to assign types. I don't quite understand the point of types or why they're useful when labels are a superset of functionality.

russellhancox: > I don't quite understand the point of types or why they're useful when labels are a superset of functionality.

Types are not quite a subset of labels. An issue can have many labels, it can only ever have one type. I think of types as categorizing and labels as extra "notes".

I would much prefer that we remove the "bug", "enhancement", and "question" labels and use types for those. An issue can only ever be a feature, a bug, or a question, so being able to attach both "bug" and "enhancement" labels to an issue doesn't make sense.

Labels then enhance an issue, we can use them for prioritization (e.g P0, P1, P2 labels) or scoping (we already have size/X labels).

mlw: Resolved offline.

PR #474, opened 2025-06-28
gui: Fix minor UI issue when an app sets empty bundle name

mlw:
This is a minor UI issue only that is rare and caused by an app developer intentionally setting their bundle name to an empty string. This issue only affects what data Santa chooses to show in the UI, and could result in labels having missing values. There is no missing data in sync events or telemetry.

### DIFF
--- a/Source/gui/SNTBinaryMessageWindowView.swift
+++ b/Source/gui/SNTBinaryMessageWindowView.swift
@@ -57,7 +57,7 @@ import santa_gui_SNTMessageView
 }
 
 func copyDetailsToClipboard(e: SNTStoredEvent?, customURL: String?) {
-  var s = "Santa blocked \(e?.fileBundleName ?? "an application:")"
+  var s = "Santa blocked \((e?.fileBundleName?.isEmpty == false) ? e!.fileBundleName! : "an application")"
   if let publisher = e?.publisherInfo {
     s += "\nPublisher  : \(publisher)"
   }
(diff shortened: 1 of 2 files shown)


Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -57,7 +57,7 @@ import santa_gui_SNTMessageView
 }
 
 func copyDetailsToClipboard(e: SNTStoredEvent?, customURL: String?) {
-  var s = "Santa blocked \(e?.fileBundleName ?? "an application:")"
+  var s = "Santa blocked \((e?.fileBundleName?.isEmpty == false) ? e!.fileBundleName! : "an application")"
```
russellhancox, line 60: With the change to SNTFileInfo, the previous behavior or just nil-checking fileBundleName should be enough, no?

Source/gui/SNTBinaryMessageWindowView.swift
```diff
@@ -203,7 +203,7 @@ struct SNTBinaryMessageEventView: View {
       Divider()
 
       VStack(alignment: .leading, spacing: 10.0) {
-        if let bundleName = e?.fileBundleName {
+        if let bundleName = e?.fileBundleName, !bundleName.isEmpty {
```
russellhancox, line 206: Same; with the change to SNTFileInfo the second case is essentially impossible.

Source/gui/SNTBinaryMessageWindowView.swift
mlw: Yeah, it would be enough today. I liked the idea of matching the logic that was used to print the "Application"/"File" label. It felt odd to make that check less robust given this code is reliant on data from another component, so I made these checks more robust so that things are more consistent.

russellhancox (approved):

PR #483, opened 2025-07-02
Adopt scoped type to fix leak

mlw:
This builds out more of ScopedCFTypeRef to make it more generally useful and then adopts it in MOLCodesignChecker to fix a minor memory leak.

### DIFF
--- a/Source/common/MOLCodesignChecker.mm
+++ b/Source/common/MOLCodesignChecker.mm
@@ -95,8 +95,8 @@ static NSString *const kErrorDomain = @"com.google.molcodesignchecker";
       }
     });
 
-    // For static code checks, if it was valid, perform additional checks
-    if (CFGetTypeID(codeRef) == SecStaticCodeGetTypeID() && !scopedError) {
+    // For static code checks perform additional checks across all slices
+    if (CFGetTypeID(codeRef) == SecStaticCodeGetTypeID()) {
       // Ensure signing is consistent for all architectures.
       // Any issues found here take precedence over already found issues.
       if (!_binaryPath) _binaryPath = [self binaryPathForCodeRef:self.codeRef];
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,6 +3,7 @@ build --define=apple.propagate_embedded_extra_outputs=yes
 
 build --copt=-Werror
 build --copt=-Wall
+build --copt=-Wformat-security
 build --copt=-Wno-error=deprecated-declarations
 build --copt=-Wno-error=nullability-completeness
 build --copt=-Wreorder-init-list


Source/common/MOLCodesignChecker.mm
```diff
@@ -77,12 +82,18 @@ - (instancetype)initWithSecStaticCodeRef:(SecStaticCodeRef)codeRef error:(NSErro
...
   if (self) {
-    OSStatus status = errSecSuccess;
-    CFErrorRef cfError = NULL;
+    auto [status, scopedError] = ScopedCFError::AssumeFrom(^OSStatus(CFErrorRef *out) {
+      if (CFGetTypeID(codeRef) == SecStaticCodeGetTypeID()) {
+        return SecStaticCodeCheckValidityWithErrors(codeRef, kStaticSigningFlags, NULL, out);
+      } else if (CFGetTypeID(codeRef) == SecCodeGetTypeID()) {
+        return SecCodeCheckValidityWithErrors((SecCodeRef)codeRef, kSigningFlags, NULL, out);
+      } else {
+        return errSecSuccess;
```
russellhancox: Not sure success is the right response here, the codeRef should not be empty and so it should only be one of those two.

Perhaps this if/else if/else should be a switch on the CFGetTypeID response and the default should be to return errSecUnimplemented or errSecBadReq

Source/common/MOLCodesignChecker.mm
```diff
@@ -92,14 +103,13 @@ - (instancetype)initWithSecStaticCodeRef:(SecStaticCodeRef)codeRef error:(NSErro
...
+        scopedError = ScopedCFError::Retain(
+            (__bridge CFErrorRef)[self errorWithCode:status description:description]);
       }
-    } else if (CFGetTypeID(codeRef) == SecCodeGetTypeID()) {
-      status = SecCodeCheckValidityWithErrors((SecCodeRef)codeRef, kSigningFlags, NULL, &cfError);
     }
 
     // Do not set _signingInformation or _certificates for universal binaries with signing issues.
-    NSError *err = CFBridgingRelease(cfError);
+    NSError *err = scopedError.Bridge<NSError *>();
```
russellhancox: I'm concerned about lifetime here. This err is ultimately stored in the *error out-parameter. As an "owned" NSError that's fine, but here the underlying error is owned by the scopedError so when that goes out of scope won't it prematurely free?

Can we add a BridgeRelease method that uses CFBridgingRelease and then marks itself invalid?

Source/common/MOLCodesignChecker.mm
mlw: Done.

Source/common/MOLCodesignChecker.mm
mlw: Done

PR #486, opened 2025-07-08
santactl/doctor: Stop complaining about standard profile keys

russellhancox:


Source/common/SNTConfigurator.m
```diff
@@ -1609,6 +1609,13 @@ - (nullable NSArray *)validateConfiguration {
     CFStringRef keyRef = (__bridge CFStringRef)key;
     if (!CFPreferencesAppValueIsForced(keyRef, kMobileConfigDomain)) return;
 
+    // If the key is a 'standard' configuration profile key, skip it.
```
mlw, line 1612: Why are these nested in the Santa config?

russellhancox: See the changes in https://github.com/northpolesec/santa/pull/487.

mlw: I reckon I thought we were just inspecting the payload content section, but doctor is looking at the whole mobileconfig? If so, should there be other ignored keys as well?

mlw (approved):

Source/common/SNTConfigurator.m
russellhancox: It's not as simple as that; the PayloadContent in the profile ends up in the user defaults and we're validating all forced values in the user defaults. In my testing the only two "profile" keys that end up passed through are `PayloadUUID` and `_manualProfile` (which only applies to manually installed mobileconfig files, installs via MDM don't have this).

PR #488, opened 2025-07-09
santactl/fileinfo: Fix mismatched signature error matching

russellhancox:


Source/common/MOLCodesignChecker.mm
```diff
@@ -61,7 +61,7 @@ up validation (see below) but does mean images, plists, etc will not be checked
 */
 static const SecCSFlags kSigningFlags = kSecCSDefaultFlags;
 
-static NSString *const kErrorDomain = @"com.google.molcodesignchecker";
+static NSString *const kErrorDomain = @"com.northpolesec.santa.molcodesignchecker";
```
tburgin: We should export this so the caller doesn't have the redefine it. It will prevent this type of mismatch from happening again.

russellhancox: Good call

PR #490, opened 2025-07-10
ObjC++ all the things

mlw:


Source/common/SNTMetricSetTest.mm
```diff
@@ -27,7 +27,7 @@ @interface SNTMetricSetHelperFunctionsTest : XCTestCase
 @implementation NSDate (custom)
 
 + (instancetype)date {
-  NSDateFormatter *formatter = NSDateFormatter.new;
+  NSDateFormatter *formatter = [NSDateFormatter new];
```
russellhancox:
```suggestion
  NSDateFormatter *formatter = [[NSDateFormatter alloc] init];
```

https://google.github.io/styleguide/objcguide.html#do-not-use-new

PR #497, opened 2025-07-22
debug: fix SANTA_STORE_SYNC_JSON

tburgin:
Santa fails to build with `--define=SANTA_STORE_SYNC_JSON=1` - `santa::StringToNSString` doesn't support taking a `std::string_view`.

I added a variant that takes a string_view. I noticed we are assuming UTF8 input, so I updated the names to reflect this. This matches better with the `santa::NSStringToUTF8StringView` naming.

### DIFF
--- a/Source/common/String.h
+++ b/Source/common/String.h
@@ -33,20 +33,25 @@ static inline std::string NSStringToUTF8String(NSString *str) {
   return std::string(str.UTF8String, [str lengthOfBytesUsingEncoding:NSUTF8StringEncoding]);
 }
 
-static inline NSString *StringToNSString(const std::string &str) {
+static inline NSString *UTF8StringToNSString(const std::string &str) {
   return [NSString stringWithUTF8String:str.c_str()];
 }
 
-static inline NSString *StringToNSString(const char *str) {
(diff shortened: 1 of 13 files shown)


Source/common/String.h
```diff
@@ -33,20 +33,26 @@ static inline std::string NSStringToUTF8String(NSString *str) {
...
+}
+
+static inline NSString *OptionalUTF8StringToNSString(
+    const std::optional<std::string> &optional_str) {
   std::string str = optional_str.value_or("");
   if (str.length() == 0) {
     return nil;
   } else {
-    return StringToNSString(str);
+    return UTF8StringToNSString(str);
```
russellhancox, line 55:
```suggestion
if (!optional_str.has_value()) return nil;
return UTF8StringToNSString(optional_str.value());
```

mlw: Need to be careful here, this changes behavior - Maybe for the better? But not sure the ramifications.

Previously, passing an empty string would result in a `nil` result being returned.  The suggested edit will return a valid object with an empty string.

tburgin: To more thoroughly test the suggestion, I'll punt on this for now.

PR #498, opened 2025-07-23
Rework SNTStoredEvent

mlw:
This PR changes `SNTStoredEvent` to be a proper base class for a new `SNTStoredExecutionEvent` and `SNTStoredFileAccessEvent` (previously `SNTFIleAccessEvent`).

This is prep work for being able to store all types conforming to SNTStoredEvent in the events database.


Source/common/SNTStoredEvent.mm
```diff
@@ -127,29 +50,20 @@ - (BOOL)isEqual:(id)other {
...
-}
-
-- (NSString *)publisherInfo {
-  return Publisher(self.signingChain, self.teamID);
-}
-
-- (NSArray *)signingChainCertRefs {
-  return CertificateChain(self.signingChain);
+- (NSString *)hashForEvent;
+{
```
russellhancox:
```suggestion
- (NSString *)hashForEvent {
```

Source/common/SNTStoredEventTest.mm
```diff
@@ -12,17 +12,19 @@
 /// See the License for the specific language governing permissions and
```
russellhancox, line 12: It's a bit odd that this is SNTStoredEventTest but is testing both SNTStoredExecutionEvent and SNTStoredFileAccessEvent. Perhaps split into multiple test files?

Source/santad/EventProviders/EndpointSecurity/MessageTest.mm
```diff
@@ -51,6 +54,17 @@ pid_t AttemptToFindUnusedPID() {
   return 0;
 }
 
+std::string GetProcessPath(pid_t pid) {
+  char pathbuf[PROC_PIDPATHINFO_MAXSIZE] = {};
+  int ret = proc_pidpath(pid, pathbuf, sizeof(pathbuf));
+  NSLog(@"got ret: %d, path buf: %s", ret, pathbuf);
```
russellhancox: Debugging?

Source/common/SNTStoredEvent.mm
mlw: Added new clang copt to catch this.

PR #502, opened 2025-07-25
sync: Send a hash of all database rules at beginning and end of sync

russellhancox:


Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -664,4 +668,39 @@ - (void)updateStaticRules:(NSArray<NSDictionary *> *)staticRules {
...
+    FMResultSet *rs =
+        [db executeQuery:@"SELECT identifier, state, type, cel_expr FROM rules WHERE state != ?",
+                         @(SNTRuleStateAllowTransitive)];
+    while ([rs next]) {
+      NSString *identifier = [rs stringForColumn:@"identifier"];
+      NSString *cel = [rs stringForColumn:@"cel_expr"];
+      int state = [rs intForColumn:@"state"];
+      int type = [rs intForColumn:@"type"];
+
+      hash.Update(identifier.UTF8String, identifier.length);
```
tburgin:
```suggestion
      hash.Update(identifier.UTF8String, [identifier lengthOfBytesUsingEncoding:NSUTF8StringEncoding]);
```

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -664,4 +668,39 @@ - (void)updateStaticRules:(NSArray<NSDictionary *> *)staticRules {
...
+      hash.Update(cel.UTF8String, cel.length);
```
tburgin:
```suggestion
      hash.Update(cel.UTF8String, [cel lengthOfBytesUsingEncoding:NSUTF8StringEncoding]);
```

Source/common/String.h
```diff
@@ -41,6 +41,10 @@ static inline NSString *StringToNSString(const char *str) {
   return [NSString stringWithUTF8String:str];
 }
 
+static inline NSString *StringViewToNSString(const std::string_view &str) {
```
mlw: nit: Should be named the same as the other overloads `StringToNSString`

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -516,6 +517,9 @@ - (BOOL)addRules:(NSArray *)rules ruleCleanup:(SNTRuleCleanup)cleanupType error:
     *error = blockErr;
   }
 
+  // Clear the rules hash
+  self.rulesHash = nil;
```
mlw: This should probably be done within the transaction so it's atomic with the operation

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -664,4 +668,40 @@ - (void)updateStaticRules:(NSArray<NSDictionary *> *)staticRules {
...
+      NSString *identifier = [rs stringForColumn:@"identifier"];
+      NSString *cel = [rs stringForColumn:@"cel_expr"];
+      int state = [rs intForColumn:@"state"];
+      int type = [rs intForColumn:@"type"];
+
+      hash.Update(identifier.UTF8String,
+                  [identifier lengthOfBytesUsingEncoding:NSUTF8StringEncoding]);
+      hash.Update(cel.UTF8String, [cel lengthOfBytesUsingEncoding:NSUTF8StringEncoding]);
+      hash.Update((const void *)&state, sizeof(state));
+      hash.Update((const void *)&type, sizeof(type));
```
mlw: nit: `static_cast<void*>(...)` (shouldn't need to worry about const here at the call site)

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -664,4 +668,40 @@ - (void)updateStaticRules:(NSArray<NSDictionary *> *)staticRules {
   self.cachedStaticRules = [rules copy];
 }
 
+- (NSString *)hashOfHashes {
+  // If santad has previously computed the hash and stored it in memory, return it.
+  // When a rule is added or removed the hash will be cleared so that the next
+  // request for the hash will recompute it.
+  if (self.rulesHash.length) {
+    return self.rulesHash;
```
mlw: we should prolly move `ruleHash` into the queue - same with when set down below.

mlw (changes requested):

Source/santad/DataLayer/SNTRuleTable.mm
russellhancox: I wasn't totally sure what this meant, but I _think_ I got it.

mlw: Yep, apologies for the confusion.

mlw (approved):

PR #503, opened 2025-07-26
Cookbook: Add CEL rule to prevent users from disabling gatekeeper

pmarkowsky:
This adds a rule for preventing users from disabling gatekeeper via spctl.

### DIFF
--- a/docs/docs/cookbook/cel.md
+++ b/docs/docs/cookbook/cel.md
@@ -17,5 +17,5 @@ target.signing_time >= timestamp('2025-05-31T00:00:00Z')
 Create a signing ID rule for `platform:com.apple.spctl` and attach the following CEL program
 
 ```clike
-['--global-disable', '--disable', '--remove'].exists(flag, flag in args) ? BLOCKLIST : ALLOWLIST
+['--global-disable', '--master-disable','--disable', '--add', '--remove'].exists(flag, flag in args) ? BLOCKLIST : ALLOWLIST
 ```
--- a/docs/docs/cookbook/cel.md
+++ b/docs/docs/cookbook/cel.md
@@ -17,5 +17,5 @@ target.signing_time >= timestamp('2025-05-31T00:00:00Z')
 Create a signing ID rule for `platform:com.apple.spctl` and attach the following CEL program
 
 ```clike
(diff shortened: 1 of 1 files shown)


docs/docs/cookbook/cel.md
```diff
@@ -12,3 +12,10 @@ before the provided date. This is particularly useful when attached to a
 target.signing_time >= timestamp('2025-05-31T00:00:00Z')
 ```
 
+## Prevent users from disabling gatekeeper
+
+Create a signing ID rule for `platform:com.apple.spctl` and attach the following CEL program
+
+```clike
+['--global-disable', '--disable', '--remove'].exists(flag, flag in args) : BLOCKLIST ? ALLOWLIST
```
russellhancox:
```suggestion
['--global-disable', '--disable', '--remove'].exists(flag, flag in args) ? BLOCKLIST : ALLOWLIST
```

docs/docs/cookbook/cel.md
```diff
@@ -12,3 +12,10 @@ before the provided date. This is particularly useful when attached to a
 target.signing_time >= timestamp('2025-05-31T00:00:00Z')
 ```
 
+## Prevent users from disabling gatekeeper
+
+Create a signing ID rule for `platform:com.apple.spctl` and attach the following CEL program
+
+```clike
+['--global-disable', '--disable', '--remove'].exists(flag, flag in args) ? BLOCKLIST : ALLOWLIST
```
russellhancox:
```suggestion
['--master-disable', '--disable', '--remove'].exists(flag, flag in args) ? BLOCKLIST : ALLOWLIST
```

russellhancox: If we're handling --remove you may also want to add --add, otherwise a user could add a rule for a program on disk that would otherwise be denied.

pmarkowsky: I did't see --master-disable in the man page. But trying it seems to work seems like we need both.

```
NAME
     spctl – SecAssessment system policy security

SYNOPSIS
     spctl --assess [-t type] [-] file ...
     spctl --status
     spctl --global-enable
     spctl --global-disable | --disable-status
```

pmarkowsky: Makes sense, added.

russellhancox (approved):

PR #512, opened 2025-08-01
docs: Add Slack Cookies FAA policy

russellhancox:


docs/docs/cookbook/faa.md
```diff
@@ -48,6 +48,51 @@ process.
 </dict>
 ```
 
+## Slack Cookies
+
+This policy will prevent reads of cookies from the Slack app, except to Slack
+itself and the Spotlight indexing process. This is almost identical to the
+Chrome Browser rule above because Slack is built with Electron.
```
pmarkowsky, line 55: Probably worth calling out that this rule was highlighted by SpecterOps in their talk [Modern macOS Red Teaming Tactics](https://www.youtube.com/watch?v=t_L2bdbXkp0&t=2863s)

PR #513, opened 2025-08-04
Make Spool a template class on type of batcher member

mlw:
This restructures the spool-related classes to prep for a new spool-based writer.

The max batch size support was remove from `FsSpoolLogBatchWriter` was removed as it isn't used and complicated the design.

### DIFF
--- a/Source/santad/Logs/EndpointSecurity/Writers/Spool.h
+++ b/Source/santad/Logs/EndpointSecurity/Writers/Spool.h
@@ -113,11 +113,12 @@ class Spool : public Writer, public std::enable_shared_from_this<Spool<T>> {
       // This will account for Flush failing above.
       // Use the more lenient threshold here in case the Flush failures are transitory.
       if (shared_this->accumulated_bytes_ < shared_this->spool_file_size_threshold_leniency_) {
-        shared_this->accumulated_bytes_ += moved_bytes.size();
-        // auto status = shared_this->log_batch_writer_.WriteMessage(any);
+        size_t bytes_written = moved_bytes.size();
         auto status = shared_this->log_batch_writer_.WriteMessage(std::move(moved_bytes));
         if (!status.ok()) {
           LOGE(@"ProtoEventLogger::LogProto failed with: %s", status.ToString().c_str());
+        } else {
+          shared_this->accumulated_bytes_ += bytes_written;
         }
       }
 


Source/santad/Logs/EndpointSecurity/Writers/Spool.h
```diff
@@ -25,56 +25,190 @@
...
+      if (shared_this->accumulated_bytes_ >= shared_this->spool_file_size_threshold_) {
+        shared_this->FlushSerialized();
+      }
+
+      // Only write the new message if we have room left.
+      // This will account for Flush failing above.
+      // Use the more lenient threshold here in case the Flush failures are transitory.
+      if (shared_this->accumulated_bytes_ < shared_this->spool_file_size_threshold_leniency_) {
+        shared_this->accumulated_bytes_ += moved_bytes.size();
+        // auto status = shared_this->log_batch_writer_.WriteMessage(any);
```
tburgin: Leftover?

Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool_log_batch_writer.cc
```diff
@@ -61,19 +69,22 @@ absl::Status FsSpoolLogBatchWriter::FlushNoLock() {
...
   // long time (forever?).
   cache_ = santa::fsspool::binaryproto::LogBatch();
-  cache_.mutable_records()->Reserve(max_batch_size_);
+  cache_.mutable_records()->Reserve(kReservedBatchSize);
   return absl::OkStatus();
 }
 
-absl::Status FsSpoolLogBatchWriter::WriteMessage(
-    const ::google::protobuf::Any& msg) {
+absl::Status FsSpoolLogBatchWriter::WriteMessage(std::vector<uint8_t> bytes) {
```
tburgin, line 76: This now always returns `absl::OkStatus()`, do you still need the return status? I don't see any other functions overriding this signature with other implementations.

Source/santad/Logs/EndpointSecurity/Writers/Spool.h
```diff
@@ -25,56 +25,190 @@
...
+
+      if (shared_this->accumulated_bytes_ >= shared_this->spool_file_size_threshold_) {
+        shared_this->FlushSerialized();
+      }
+
+      // Only write the new message if we have room left.
+      // This will account for Flush failing above.
+      // Use the more lenient threshold here in case the Flush failures are transitory.
+      if (shared_this->accumulated_bytes_ < shared_this->spool_file_size_threshold_leniency_) {
+        shared_this->accumulated_bytes_ += moved_bytes.size();
```
tburgin: `accumulated_bytes_` is incremented before the write has succeeded. Technically this should be okay because `WriteMessage` never returns an error, and the branch on L119 doesn't do anything meaningful - but it was confusing the read. I suggest clearing up if `WriteMessage` needs to return a status, and if so remove the branch on L119.

Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool_log_batch_writer.cc
mlw: I thought about this, but I'm suspecting the upcoming batcher will have more conditions. Since this was the existing type, I opted to keep as-is for now instead of changing, and then having to change back. If the upcoming batcher doesn't need it I'm happy to change it then.

Source/santad/Logs/EndpointSecurity/Writers/Spool.h
mlw: Fixed.

tburgin (approved):

Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool_log_batch_writer.cc
tburgin: SGTM

PR #521, opened 2025-08-11
telemetry export: prepare for signed URL export

tburgin:
This change removes the rednose uploader and gets ready to use signed URLs to upload.

### DIFF
--- a/Source/common/SNTExportConfiguration.h
+++ b/Source/common/SNTExportConfiguration.h
@@ -21,4 +21,7 @@
 
 - (instancetype)initWithURL:(NSURL *)url formValues:(NSDictionary *)formValues;
 
+- (NSData *)serialize;
++ (instancetype)deserialize:(NSData *)data;
+
 @end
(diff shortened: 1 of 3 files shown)


Source/common/SNTExportConfiguration.h
```diff
@@ -14,59 +14,11 @@
...
-
-// GCP Initializer
-- (instancetype)initWithGCPBearerToken:(NSString *)bearerToken
-                            bucketName:(NSString *)bucketName
-                       objectKeyPrefix:(NSString *)objectKeyPrefix;
+@property(readonly) NSURL *url;
+@property(readonly) NSDictionary *formValues;
 
-- (NSData *)serialize;
-+ (instancetype)deserialize:(NSData *)data;
```
russellhancox: I think it would be nice to keep these

tburgin: ack - done

mlw (approved):

PR #524, opened 2025-08-12
Add string extensions to CEL evaluator

pmarkowsky:
We weren't registering the string extensions in the CEL operator which made matching on sequences more difficult than it should be.

This allows us to do things like `args.join(' ').contains('-setremotelogin on')`.


Source/common/cel/Evaluator.mm
```diff
@@ -51,6 +52,12 @@
     return result;
   }
 
+  // Add the extensions for string operations list join
```
russellhancox:
```suggestion
  // Add the extensions for string operations like list join
```

russellhancox (approved):

PR #528, opened 2025-08-12
telemetry export: post to cloud bucket

tburgin:


Source/santasyncservice/SNTStreamingMultipartFormData.mm
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  // A partial write to stream has occurred, adjust the fd offset so the next
+  // write operation picks up where this partial write left off.
+  uint64_t offset = 0;
+  if (![self.fd getOffset:&offset error:&error]) {
+    [self closeWithError:error];
+    return;
+  }
+  if (![self.fd seekToOffset:offset - (data.length - bytesWritten) error:&error]) {
```
russellhancox: It's annoying NSFileHandle doesn't support relative seek. 

Perhaps a future optimization but we could avoid the extra get call by either doing seek() ourselves on the underlying fd or tracking the offset as we read

Source/santasyncservice/SNTSyncService.mm
```diff
@@ -102,8 +103,40 @@ - (void)exportTelemetryFile:(NSFileHandle *)fd
...
+
+  NSURLSessionConfiguration *sessionConfig =
+      [NSURLSessionConfiguration ephemeralSessionConfiguration];
+  NSURLSession *session = [NSURLSession sessionWithConfiguration:sessionConfig];
+  NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:config.url];
+  request.HTTPBodyStream = stream.stream;
+  [request setValue:stream.contentType forHTTPHeaderField:@"Content-Type"];
+  [request setValue:[NSString stringWithFormat:@"%lu", stream.contentLength]
+      forHTTPHeaderField:@"Content-Length"];
+  [request setValue:@"santactl-telemetry/" forHTTPHeaderField:@"User-Agent"];
```
russellhancox: This should use the same one as normal syncs to include the version or not bother

Source/santasyncservice/SNTSyncService.mm
```diff
@@ -102,8 +103,40 @@ - (void)exportTelemetryFile:(NSFileHandle *)fd
...
-  // TODO: Export telemetry.
-  reply(NO);
+  SNTStreamingMultipartFormData *stream =
+      [[SNTStreamingMultipartFormData alloc] initWithFormParts:config.formValues
+                                                          file:fd
+                                                      fileName:fileName];
+
+  NSURLSessionConfiguration *sessionConfig =
+      [NSURLSessionConfiguration ephemeralSessionConfiguration];
+  NSURLSession *session = [NSURLSession sessionWithConfiguration:sessionConfig];
```
russellhancox, line 115: As you're not customizing the session or its configuration, any reason not to use `[NSURLSession sharedSession]`?

Source/santasyncservice/SNTStreamingMultipartFormData.mm
```diff
@@ -0,0 +1,209 @@
...
+    // about using CFWriteStreamSetDispatchQueue.
+    // https://stackoverflow.com/questions/31306642/how-to-deal-with-concurrency-issues-brought-by-nsstream-run-loop-scheduling-usin
+    [output setDelegate:self];
+    CFWriteStreamSetDispatchQueue((__bridge CFWriteStreamRef)output, _streamQueue);
+    [output open];
+    _output = output;
+
+    // Load up the stream with form data. This should never block since it is
+    // smaller than the buffer.
+    if ([output write:(const uint8_t *)formData.bytes
```
mlw: Once the streams are open, is there a race here if the data being written here is done at the same time as data written from a thread running on `_streamQueue`?

tburgin: Yes, nice catch. I have not been able to have have the delegate callback to win the race, but in theory it could be an issue. I have moved writing of the form data to be handled in the delegate callback.

Source/santasyncservice/SNTStreamingMultipartFormData.mm
tburgin: Good idea. Done.

Source/santasyncservice/SNTSyncService.mm
tburgin: The ephemeral session seems more appropriate since we don't need any caches or cookies.

Source/santasyncservice/SNTSyncService.mm
tburgin: Removed

tburgin:
I have also added the desired XPC interface for handling multiple files. There are TODOs for each side of the request to support it.

Source/santasyncservice/SNTStreamingMultipartFormData.mm
```diff
@@ -0,0 +1,262 @@
...
+    // 8   SNTStreamingMultipartFormDataTest     0x10768ce4c -[SNTStreamingMultipartFormData initWithFormParts:file:fileName:] + 2180
+    // ...
+    // Thread 1 Crashed::  Dispatch queue: com.northpolesec.santa.syncservice.multipartstream
+    // 0   libsystem_pthread.dylib               0x185ab54f8 pthread_mutex_lock + 12
+    // 1   CoreFoundation                        0x185c0f484 _CFStreamSetDispatchQueue + 244
+    // 2   SNTStreamingMultipartFormDataTest     0x10768d9cc -[SNTStreamingMultipartFormData closeWithError:] + 776
+    // 3   SNTStreamingMultipartFormDataTest     0x10768e82c -[SNTStreamingMultipartFormData stream:handleEvent:] + 1244
+    // 4   CoreFoundation                        0x185bc5ec8 _outputStreamCallbackFunc + 76
+    //
+    // clang-format on
```
russellhancox, line 115: The comment is very useful but I'm not sure the stack trace is really needed

Source/common/SNTXPCSyncServiceInterface.h
```diff
@@ -34,10 +31,14 @@
 - (void)postBundleEventToSyncServer:(SNTStoredExecutionEvent *)event
                               reply:(void (^)(SNTBundleEventAction))reply;
 - (void)pushNotificationStatus:(void (^)(SNTPushNotificationStatus))reply;
-- (void)exportTelemetryFile:(NSFileHandle *)fd
-                   fileName:(NSString *)fileName
-                     config:(SNTExportConfiguration *)config
-                      reply:(void (^)(BOOL))reply;
+
+// Array of telemetry files to export. Reply contains an array of indicies (corresponding to the
```
russellhancox:
```suggestion
// Array of telemetry files to export. Reply contains an array of indices (corresponding to the
```

russellhancox (dismissed):

Source/common/SNTXPCSyncServiceInterface.h
tburgin: Removed. I chatted with Matt, we don't need this partial success reply protocol.

Source/santasyncservice/SNTStreamingMultipartFormData.mm
tburgin: I would like to keep it. It was hard to reproduce and I want a record of exactly what happened for our future selves.

mlw (approved):

PR #529, opened 2025-08-13
Add digest to protostream encoding

mlw:
Also updates the printlog command to support verifying the digest.


Source/santactl/Commands/SNTCommandPrintLog.mm
```diff
@@ -133,23 +133,36 @@
       return absl::InternalError("Failed to parse message length");
     }
 
-    // Use a ScopedLimit to ensure we don't read past the end of the message.
-    google::protobuf::io::CodedInputStream::Limit limit = coded_input_->PushLimit(message_length);
-
     // Read the raw message data
+    std::vector<uint8_t> msg_buf(message_length);
+    if (!coded_input_->ReadRaw(msg_buf.data(), message_length)) {
+      return absl::InternalError("Failed to read message buffer");
```
russellhancox:
```suggestion
      return absl::InternalError("Failed to read message into buffer");
```

Source/santactl/Commands/SNTCommandPrintLog.mm
```diff
@@ -133,23 +133,36 @@
...
+    }
+
+    if (expected_hash != 0) {
+      santa::Xxhash64 xxhash;
+      xxhash.Update(msg_buf.data(), msg_buf.size());
+      __block uint64_t got_hash;
+      xxhash.Digest(^(const uint8_t *buf, size_t size) {
+        memcpy(&got_hash, buf, sizeof(got_hash));
```
russellhancox: Instead of copying this out, could we not do the comparison inside the block and copy the bool result out?

mlw: Changed this up to assign directly, the memcpy isn't really necessary. I think the new code is more readable and probably no real different than copying out a bool.

russellhancox (approved):

PR #531, opened 2025-08-13
docs: Add llms.txt

statico:
This change adds the [docusaurus-plugin-llms](https://www.npmjs.com/package/docusaurus-plugin-llms) plugin which adds `/llms.txt` and `/llms-full.txt` to the docs. See https://llmstxt.org/

The output of the plugin is not perfect but should be satisfactory for now.

CleanShot 2025-08-13 at 12 38 00


pmarkowsky (approved):
LGTM.

PR #541, opened 2025-08-17
Compressed protostream support

mlw:
Adds ability to compress protostream output with either Gzip ro Zstd, and adds new `EventLogType`s to support them.


Source/common/SNTConfigurator.h
```diff
@@ -188,8 +188,12 @@
 ///      format. Use spoolDirectory to specify a path. Use spoolDirectoryFileSizeThresholdKB,
 ///      spoolDirectorySizeThresholdMB and spoolDirectoryEventMaxFlushTimeSec to configure
 ///      additional settings.
-///    SNTEventLogTypeProtobufStream "protobufstream": Similar to "protobuf", but formatted as a
+///    SNTEventLogTypeProtobufStream "ProtobufStream": Similar to "protobuf", but formatted as a
 ///      coded output stream.
+///    SNTEventLogTypeProtobufStreamGzip "ProtobufStreamGzip": Similar to "ProtobufStream", but
+///      output is compressed as gzip.
+///    SNTEventLogTypeProtobufStreamZstd "ProtobufStreamZstd": Similar to "ProtobufStream", but
+///      output is compressed as zstd.
```
russellhancox: It's a little weird that the config strings are capitalized here when none of the other formats are. Given that we ignore capitalization, is it necessary to change?

mlw: No real reason to change. Thought it was a bit easier to read, but nbd. Changed it back.

PR #542, opened 2025-08-17
NSData Gzip Decompression, stream compression verification

mlw:
Adds support for zlib/gzip decompression in the NSData Zlib class category.

This also updates the StreamBatcherTest to verify gzip stream compression.

This is dependent on #541

### DIFF
--- a/Source/common/NSDataZlibTest.mm
+++ b/Source/common/NSDataZlibTest.mm
@@ -1,16 +1,17 @@
 /// Copyright 2022 Google Inc. All rights reserved.
+/// Copyright 2025 North Pole Security, Inc.
 ///
 /// Licensed under the Apache License, Version 2.0 (the "License");
 /// you may not use this file except in compliance with the License.
 /// You may obtain a copy of the License at
 ///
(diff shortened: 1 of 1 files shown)


Source/common/NSDataZlibTest.mm
```diff
@@ -32,19 +32,33 @@ - (NSData *)dataFromFixture:(NSString *)file {
...
 }
 
+- (void)testZlibDecompressed {
+  NSData *sut = [self dataFromFixture:@"compression_test_zlib.z"];
+  NSData *want = [self dataFromFixture:@"compression_test_uncompressed.json"];
+
+  XCTAssertEqualObjects([sut zlibDecompressed], want);
+}
+
+- (void)testGzipDecommpressed {
```
russellhancox:
```suggestion
- (void)testGzipDecompressed {
```

Source/common/NSDataZlibTest.mm
```diff
@@ -14,7 +14,7 @@
```
russellhancox, line 15: Consider updating the license

PR #543, opened 2025-08-18
Move TemporaryFile class to a standalone ScopedFile class

mlw:
Moves the TemporaryFile class out from `SNTStreamingMultipartFormDataTest.mm` as this functionality will be needed elsewhere

### DIFF
--- a/Source/common/ScopedFile.h
+++ b/Source/common/ScopedFile.h
@@ -92,13 +92,10 @@ class ScopedFile {
   }
 
   ScopedFile &operator=(ScopedFile &&rhs) {
-    if (this == &rhs) {
-      return *this;
+    if (this != &rhs) {
+      fd_ = rhs.fd_;
+      rhs.fd_ = -1;
     }
-
-    fd_ = rhs.fd_;
-    rhs.fd_ = -1;
-
     return *this;
   }
 


Source/common/ScopedFile.h
```diff
@@ -0,0 +1,97 @@
...
+
+    path = [path stringByAppendingFormat:@"/%@", filename_template];
+
+    char *mutable_path = strdup(path.UTF8String);
+    if (!mutable_path) {
+      LOGE(@"Failed to allocate memory for temp file path");
+      return nullptr;
+    }
+
+    NSLog(@"About to create temporary: %s", mutable_path);
```
russellhancox: LOGD or remove

Source/common/ScopedFile.h
```diff
@@ -0,0 +1,97 @@
...
+#include <cstring>
+#include <memory>
+
+#import "Source/common/SNTLogging.h"
+
+namespace santa {
+
+class ScopedFile {
+ public:
+  static std::unique_ptr<ScopedFile> CreateTemporary(
```
russellhancox: Could we add move constructors and avoid using unique_ptr here? Then this could be a regular constructor.

mlw: Because of so many different failure cases, I like that factory approach so we don't have to throw or have an otherwise invalid object state. I could instead return a StatusOr - could see that being preferred

Source/common/ScopedFile.h
```diff
@@ -70,16 +72,36 @@ class ScopedFile {
...
+
+  ScopedFile &operator=(ScopedFile &&rhs) {
+    if (this == &rhs) {
+      return *this;
+    }
+
+    fd_ = rhs.fd_;
+    rhs.fd_ = -1;
+
+    return *this;
```
russellhancox:
```suggestion
    if (this != &rhs) {
      fd_ = rhs.fd_;
      rhs.fd_ = -1;
    }
    return *this;
```

PR #544, opened 2025-08-18
Support compressed stream in santactl printlog

mlw:
This PR extends `santactl printlog` to first determine if any of the given files were compressed, and decompress them before deserializing.

### DIFF
--- a/Source/common/ScopedFile.h
+++ b/Source/common/ScopedFile.h
@@ -24,11 +24,6 @@
 #import "Source/common/SNTLogging.h"
 #include "absl/status/statusor.h"
 
-// Forward declarations
-namespace santa {
-class ScopedFilePeer;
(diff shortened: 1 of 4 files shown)


Source/common/ScopedFile.h
```diff
@@ -107,7 +102,10 @@ class ScopedFile {
     return [[NSFileHandle alloc] initWithFileDescriptor:dup(fd_) closeOnDealloc:YES];
   }
 
-  friend class ScopedFilePeer;
+  // Some consumers need access to the raw file descriptor. But usaage must be
```
russellhancox:
```suggestion
  // Some consumers need access to the raw file descriptor. But usage must be
```

russellhancox (approved):

PR #550, opened 2025-08-21
Support export batches

mlw:
Supports exporting batches of telemetry files as a single operation as well as new supporting configuration keys. Also supports handling spools that switch between event log types and will independently batch files of the different types.

### DIFF
--- a/Source/santad/Logs/EndpointSecurity/Logger.h
+++ b/Source/santad/Logs/EndpointSecurity/Logger.h
@@ -18,6 +18,7 @@
 
 #import <Foundation/Foundation.h>
 
+#include <atomic>
 #include <memory>
 #include <string_view>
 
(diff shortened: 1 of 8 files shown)


Source/santad/Logs/EndpointSecurity/Logger.h
```diff
@@ -147,6 +166,9 @@ class Logger : public Timer<Logger> {
   std::shared_ptr<santa::Serializer> serializer_;
   std::shared_ptr<santa::Writer> writer_;
   ExportTracker tracker_;
+  std::unique_ptr<std::atomic_uint64_t> export_batch_threshold_size_bytes_;
+  std::unique_ptr<std::atomic_uint32_t> export_max_files_per_batch_;
+  std::unique_ptr<std::atomic_uint32_t> export_timeout_secs_;
```
tburgin, line 171: Do these need to be a pointer?

mlw: Yes because the class is movable. This is really only a requirement for how the tests are currently written. I *think* It'd be possible to rewrite tests to remove this requirement, but I haven't looked into it.

PR #568, opened 2025-09-05
sync: Add logging when private key is inaccessible

russellhancox:


Source/common/MOLAuthenticatingURLSession.mm
```diff
@@ -264,6 +270,38 @@ - (NSURLCredential *)clientCredentialForProtectionSpace:(NSURLProtectionSpace *)
...
+    // identity and attempt to sign some random data with it. This is replicating what will happen
+    // during the mTLS handshake.
+    OSStatus status;
+    ScopedSecKeyRef scopedPrivateKey;
+    std::tie(status, scopedPrivateKey) = ScopedSecKeyRef::AssumeFrom(^OSStatus(SecKeyRef *out) {
+      return SecIdentityCopyPrivateKey(foundIdentity, out);
+    });
+    if (status != errSecSuccess) {
+      // This should never really happen if we've managed to find an identity.
+      [self log:@"[Client Trust] Failed to copy private key, authentication will likely fail: %d",
```
tburgin: Nit: "copy" could give the impression Santa is doing something unsafe with a private key.

```suggestion
      [self log:@"[Client Trust] Failed to access private key, authentication will likely fail: %d",
```

Source/common/MOLAuthenticatingURLSession.mm
```diff
@@ -264,6 +270,36 @@ - (NSURLCredential *)clientCredentialForProtectionSpace:(NSURLProtectionSpace *)
...
+    });
+    if (status != errSecSuccess) {
+      // This should never really happen if we've managed to find an identity.
+      [self log:@"[Client Trust] Failed to access private key, authentication will likely fail: %d",
+                status];
+    } else {
+      NSData *dataToSign = [@"test" dataUsingEncoding:NSUTF8StringEncoding];
+      auto [scopedDataRef, scopedErrorRef] = ScopedCFError::AssumeFrom([&](CFErrorRef *out) {
+        return SecKeyCreateSignature(scopedPrivateKey.Unsafe(), kSecKeyAlgorithmRSASignatureRaw,
+                                     (__bridge CFDataRef)dataToSign, out);
```
mlw: Do we know we're always dealing with RSA keys?

russellhancox: Good catch; no, ECDSA keys should work too. I've updated to default to RSA but handle ECDSA and if neither signature algorithm is supported we'll log that instead of a random error.

mlw (approved):

PR #574, opened 2025-09-10
Add secure signing timestamp and signing timestamp

pmarkowsky:
Add secure signing timestamp and signing timestamp to the telemetry feed.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -333,6 +333,18 @@ message Execution {
 
   // Entitlement information about the target executbale
   optional EntitlementInfo entitlement_info = 16;
+
+  // The secure timestamp of when this binary was signed, as provided by the
+  // timestamp service. Not populated for unsigned or ad-hoc signed binaries,
+  // or if the secure timestamp is missing.
+  // Number of seconds since UNIX epoch.
+  uint32 secure_signing_time = 32;
+
+  // The timestamp provided by the developer when this binary was signed.
+  // It is possible for developers to set this to whatever they wish when signing
+  // unlike the secure_signing_time.
+  // Number of seconds since UNIX epoch.
+  uint32 signing_time = 33;
 }
 
 // Information about a fork event
--- a/Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
+++ b/Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
@@ -568,6 +568,14 @@ std::vector<uint8_t> Protobuf::SerializeMessage(const EnrichedExec &msg, SNTCach
   pb_exec->set_reason(GetReasonEnum(cd.decision));
   pb_exec->set_mode(GetModeEnum(cd.decisionClientMode));
 
+  if (cd.secureSigningTime) {
+    pb_exec->set_secure_signing_time([cd.secureSigningTime timeIntervalSince1970]);
+  }
+
+  if (cd.signingTime) {
+    pb_exec->set_signing_time([cd.signingTime timeIntervalSince1970]);
+  }
+
   if (cd.certSHA256 || cd.certCommonName) {
     EncodeCertificateInfo(pb_exec->mutable_certificate_info(), cd.certSHA256, cd.certCommonName);
   }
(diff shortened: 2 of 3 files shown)


Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
```diff
@@ -70,6 +70,14 @@ static inline void EncodeTimestamp(Timestamp *timestamp, struct timeval tv) {
   EncodeTimestamp(timestamp, (struct timespec){tv.tv_sec, tv.tv_usec * 1000});
 }
 
+static inline void EncodeTimestamp(Timestamp *timestamp, NSDate *date) {
+  NSTimeInterval interval = [date timeIntervalSince1970];
+  double seconds;
+  double fractional = modf(interval, &seconds);
+  timestamp->set_seconds((int64_t)seconds);
+  timestamp->set_nanos((int32_t)(fractional * 1e9));
```
mlw, line 78: We should bubble the call up to the "main" variant to set the values
```suggestion
  EncodeTimestamp(timestamp, {.tv_sec = (int64_t)seconds,
                              .tv_nsec = (int32_t)(fractional * NSEC_PER_SEC)});
```

Source/common/santa.proto
```diff
@@ -333,6 +333,16 @@ message Execution {
...
+
+  // The secure timestamp of when this binary was signed, as provided by the
+  // timestamp service. Not populated for unsigned or ad-hoc signed binaries,
+  // or if the secure timestamp is missing.
+  optional google.protobuf.Timestamp secure_signing_time = 17;
+
+  // The timestamp provided by the developer when this binary was signed.
+  // It is possible for developers to set this to whatever they wish when signing
+  // unlike the secure_signing_time.
+  optional google.protobuf.Timestamp signing_time = 18;
```
mlw, line 345: These fields are part of the code signature. Let's move it to the `CodeSignature` message.

Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
```diff
@@ -568,6 +576,14 @@ void EncodeEntitlements(::pbv1::Execution *pb_exec, SNTCachedDecision *cd) {
   pb_exec->set_reason(GetReasonEnum(cd.decision));
   pb_exec->set_mode(GetModeEnum(cd.decisionClientMode));
 
+  if (cd.secureSigningTime) {
+    EncodeTimestamp(pb_exec->mutable_secure_signing_time(), cd.secureSigningTime);
+  }
+
+  if (cd.signingTime) {
+    EncodeTimestamp(pb_exec->mutable_signing_time(), cd.signingTime);
+  }
```
mlw, line 585: Likewise, this functionality should be moved to `EncodeCodeSignature`. FWIW, my new variant looked like:

```
static inline void EncodeCodeSignature(::pbv1::CodeSignature *pb_code_sig, const es_cdhash_t cdhash,
                                       es_string_token_t sid, es_string_token_t tid,
                                       NSDate *secure_signing_time = nil,
                                       NSDate *signing_time = nil) {
  pb_code_sig->set_cdhash(cdhash, sizeof(es_cdhash_t));
  EncodeStringToken([pb_code_sig] { return pb_code_sig->mutable_signing_id(); }, sid);
  EncodeStringToken([pb_code_sig] { return pb_code_sig->mutable_team_id(); }, tid);
  if (secure_signing_time) {
    EncodeTimestamp(pb_code_sig->mutable_secure_signing_time(), secure_signing_time);
  }
  if (signing_time) {
    EncodeTimestamp(pb_code_sig->mutable_signing_time(), signing_time);
  }
}
```

Source/santad/Logs/EndpointSecurity/Serializers/ProtobufTest.mm
```diff
@@ -697,6 +697,154 @@ - (void)testSerializeMessageExecJSON {
                           json:YES];
 }
 
+- (void)testSigningTimesSerialization {
```
mlw, line 700: In the `setUp` method of `ProtobufTest`, there is an `SNTCachedDecision` that gets populated. We should add these lines - something like:

```
  self.testCachedDecision.secureSigningTime = [NSDate dateWithTimeIntervalSince1970:1751111111];
  self.testCachedDecision.signingTime = [NSDate dateWithTimeIntervalSince1970:1752222222];
```

This will then use the current testing machinery to make sure values are encoded properly. You'll just need to update the json test goldens. Much of what is here is duplicated (e.g. message setup and serialization). Testing the different variants as you have would be good, but an easier way would be to make a new `testEncodeCodeSignature` method (similar to something like `testEncodeEntitlements`). This will simplify input/output testing.

PR #581, opened 2025-09-15
Parse FAA rules in rule download phase

mlw:
This PR extracts FAA rules from the rule download sync stage into a format that will be consumed by the rest of Santa. 

Nothing is done yet with the new rules in this PR, they are merely parsed. Future PRs will do the work of verifying valid rule configuration, setting up appropriate rules DB, XPCing, and handling add/remove to the DB.

### DIFF
--- a/Source/common/SNTFileAccessRule.h
+++ b/Source/common/SNTFileAccessRule.h
@@ -23,8 +23,8 @@ typedef NS_ENUM(NSInteger, SNTFileAccessRuleState) {
 @interface SNTFileAccessRule : NSObject
 
 @property SNTFileAccessRuleState state;
-@property NSString *name;
-@property NSDictionary *details;
+@property(copy) NSString *name;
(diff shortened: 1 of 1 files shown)


Source/common/SNTFileAccessRule.h
```diff
@@ -0,0 +1,31 @@
...
+  SNTFileAccessRuleStateUnknown,
+  SNTFileAccessRuleStateAdd,
+  SNTFileAccessRuleStateRemove,
+};
+
+@interface SNTFileAccessRule : NSObject
+
+@property SNTFileAccessRuleState state;
+@property NSString *name;
+@property NSDictionary *details;
```
tburgin: nit:

```suggestion
@property(copy) NSString *name;
@property(copy) NSDictionary *details;
```

https://google.github.io/styleguide/objcguide.html#copy-potentially-mutable-objects

tburgin (approved):

PR #586, opened 2025-09-19
Add new table to the rules database for file access rules

mlw:
This also modifies the `details` property of `SNTFileAccessRule` to be `NSData` which allows it to be encoded via `NSKeyedArchiver` in the sync service before XPC'ing to the daemon and, importantly, outside of the DB transaction to minimize DB lock time.


Source/santad/DataLayer/SNTRuleTable.h
```diff
@@ -28,10 +30,15 @@
 @interface SNTRuleTable : SNTDatabaseTable
 
 ///
-///  @return Number of rules in the database
+///  @return Number of execution rules in the database
 ///
 - (int64_t)ruleCount;
```
russellhancox: Any reason not to rename this for clarity?

Source/santad/DataLayer/SNTRuleTable.h
```diff
@@ -106,10 +125,15 @@
 - (void)removeOutdatedTransitiveRules;
 
 ///
-///  Retrieve all rules from the database for export.
+///  Retrieve all execution rules from the database for export.
 ///
 - (NSArray<SNTRule *> *)retrieveAllRules;
```
russellhancox: Again, feel like this should be renamed for clarity

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -446,74 +489,163 @@ - (SNTRule *)ruleForIdentifiers:(struct RuleIdentifiers)identifiers {
...
+                      toDB:(FMDatabase *)db
+                     error:(NSError **)error {
+  for (SNTFileAccessRule *rule in fileAccessRules) {
+    // FAA rules must:
+    // 1. Be the right type
+    // 2. Have a name
+    // 3. If an add rule, must have details, otherwise must be a remove rule
+    if (![rule isKindOfClass:[SNTFileAccessRule class]] || rule.name.length == 0 ||
+        ((rule.state == SNTFileAccessRuleStateAdd && rule.details.length == 0) &&
+         (rule.state != SNTFileAccessRuleStateRemove))) {
```
russellhancox, line 502: What's the need for this last condition?

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -446,74 +489,163 @@ - (SNTRule *)ruleForIdentifiers:(struct RuleIdentifiers)identifiers {
...
+              ruleCleanup:(SNTRuleCleanup)cleanupType
+                    error:(NSError **)error {
   // Only accept an empty rules array if the cleanup-type is not none.
-  if ((!rules || rules.count < 1) && cleanupType == SNTRuleCleanupNone) {
-    [SNTError populateError:error withCode:SNTErrorCodeEmptyRuleArray format:@"Empty rule array"];
+  if (executionRules.count == 0 && fileAccessRules.count == 0 &&
+      cleanupType == SNTRuleCleanupNone) {
+    [SNTError populateError:error
+                   withCode:SNTErrorCodeEmptyRuleArray
+                     format:@"Empty execution  and file access rule arrays"];
```
russellhancox:
```suggestion
                     format:@"Empty execution and file access rule arrays"];
```

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -80,7 +81,8 @@ @interface SNTRuleTable () {
 @property NSDictionary *criticalSystemBinaries;
 @property(readonly) NSArray *criticalSystemBinaryPaths;
 @property(readwrite) NSDictionary<NSString *, SNTRule *> *cachedStaticRules;
-@property(atomic) NSString *rulesHash;
+@property(atomic) NSString *executionRulesHash;
+@property(atomic) NSString *fileAccessRulesHash;
```
russellhancox, line 85: I don't see this being set anywhere?

mlw: I added the property but was saving the work of populating it for a future PR when I hook up more values to the sync protocol (like rule counts as well)

Ok to remove now if you'd rather. Was just meant to be a placeholder.

russellhancox: Saving this for later is fine, was just checking

Source/santad/DataLayer/SNTRuleTable.mm
mlw: The line split maybe makes it a bit harder to see, but i'm trying to capture that the state must be either `Add` or `Remove`, and if it's `Add` there must be `details`.

Source/santad/DataLayer/SNTRuleTable.h
mlw: I changed a couple others as well. I hesitated for "static rules" since that matches config name. I also didn't change "transitive rules". Happy to chat if you have strong opinions here.

Source/santad/DataLayer/SNTRuleTable.mm
mlw: Done.

russellhancox (approved):

PR #588, opened 2025-09-22
Validate FAA rules on when received via sync server

mlw:
Updates the Rule Download phase to discard any received FAA rule that fails to validate.

### DIFF
--- a/Source/common/SNTFileAccessRule.h
+++ b/Source/common/SNTFileAccessRule.h
@@ -23,8 +23,8 @@ typedef NS_ENUM(NSInteger, SNTFileAccessRuleState) {
 @interface SNTFileAccessRule : NSObject <NSSecureCoding>
 
 @property(readonly) SNTFileAccessRuleState state;
-@property(readonly) NSString *name;
-@property(copy, readonly) NSData *details;
+@property(readonly, copy) NSString *name;
+@property(readonly) NSData *details;
 
 - (instancetype)initAddRuleWithName:(NSString *)name details:(NSDictionary *)details;
 - (instancetype)initRemoveRuleWithName:(NSString *)name;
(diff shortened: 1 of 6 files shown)


Source/common/faa/WatchItems.mm
```diff
@@ -519,11 +537,16 @@ bool ParseConfigSingleWatchItem(NSString *name, std::string_view policy_version,
     }
   }
 
+  // Passing nill sets means this is a verification only.
```
russellhancox:
```suggestion
  // Passing nil sets means this is a verification only.
```

Source/common/SNTFileAccessRule.h
```diff
@@ -22,10 +22,11 @@ typedef NS_ENUM(NSInteger, SNTFileAccessRuleState) {
 
 @interface SNTFileAccessRule : NSObject <NSSecureCoding>
 
-@property SNTFileAccessRuleState state;
-@property(copy) NSString *name;
-@property(copy) NSData *details;
+@property(readonly) SNTFileAccessRuleState state;
+@property(readonly) NSString *name;
```
russellhancox: This should still be a copy, otherwise there's nothing stopping a caller passing in an NSMutableString to one of the init methods and then changing the name later. And a copy is basically free.

Source/santasyncservice/SNTSyncRuleDownload.mm
```diff
@@ -224,24 +224,21 @@ - (NSDictionary *)optionsFromProtoFAARuleAdd:(const ::pbv1::FileAccessRule::Add
   switch (pbAddRule.rule_type()) {
     case ::pbv1::FileAccessRule::RULE_TYPE_UNSPECIFIED:
     case ::pbv1::FileAccessRule::RULE_TYPE_PATHS_WITH_ALLOWED_PROCESSES:
-      optionsDict[kWatchItemConfigKeyOptionsRuleType] =
-          @(static_cast<int>(santa::WatchItemRuleType::kPathsWithAllowedProcesses));
+      optionsDict[kWatchItemConfigKeyOptionsRuleType] = @"PathsWithAllowedProcesses";
```
russellhancox: Are there not constants for these strings somewhere we could import?

mlw: Added.

Source/common/SNTFileAccessRule.h
mlw: Done

PR #598, opened 2025-10-07
Add FAA rule counts and rule hashes to sync protocol

mlw:


Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -335,6 +339,7 @@ - (void)runWithArguments:(NSArray *)arguments {
     if (watchItemsEnabled) {
       printf("  %-25s | %s\n", "Policy Version", watchItemsPolicyVersion.UTF8String);
       printf("  %-25s | %llu\n", "Rule Count", watchItemsRuleCount);
+      printf("  %-25s | %s\n", "Rules Hash", [fileAccessRulesHash UTF8String]);
```
russellhancox, line 342: I'm not sure about the placement of this, given that it's not a hash of the watch items config. I think this should be down in the sync section and "Rules Hash" should be renamed to "Exec Rules Hash"

russellhancox: As discussed on Meet, we can punt on this until the general WatchItems output is massaged.

Resolving.

PR #602, opened 2025-10-11
Allow loading FAA policies that are partially invalid

mlw:
With this change, invalid rules within a policy will be ignored and valid rules loaded, instead of discarding the entire new policy.

### DIFF
--- a/Source/common/faa/WatchItemsTest.mm
+++ b/Source/common/faa/WatchItemsTest.mm
@@ -966,7 +966,7 @@ BlockGenResult CreatePolicyBlockGen() {
 
   data_policies.clear();
 
-  // Test a large config with serval valid and some invalid rules
+  // Test a large config with several valid and some invalid rules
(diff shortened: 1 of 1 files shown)


Source/common/faa/WatchItemsTest.mm
```diff
@@ -915,26 +932,81 @@ - (void)testParseConfig {
...
     kWatchItemConfigKeyVersion : @"1",
     kWatchItemConfigKeyWatchItems : @{@"a" : @{kWatchItemConfigKeyPaths : @[ @"asdf" ]}}
   },
                             &data_policies, &proc_policies, &err));
+  XCTAssertEqual(data_policies.size(), 1);
+  XCTAssertEqual(proc_policies.size(), 0);
+
+  data_policies.clear();
+
+  // Test a large config with serval valid and some invalid rules
```
russellhancox:
```suggestion
  // Test a large config with several valid and some invalid rules
```

PR #611, opened 2025-10-19
Support timer restarts, change how startup delay works

mlw:
This removes the "startup delay" variable. This must now be done externally. This variable is instead replaced with an option to dictate if the timer fires immediately on start and each subsequent interval, or if a full interval must be done first before firing.

### DIFF
--- a/Source/common/Timer.h
+++ b/Source/common/Timer.h
@@ -127,7 +127,7 @@ class Timer : public std::enable_shared_from_this<Timer<T>> {
   }
 
   /// Update the timer firing settings.
-  /// In SingleShot Mode, If the update is from a restart, will wait a full interval cycle.
+  /// In SingleShot Mode, if the update is from a restart, will wait a full interval cycle.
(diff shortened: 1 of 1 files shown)


Source/common/Timer.h
```diff
@@ -115,25 +127,20 @@ class Timer : public std::enable_shared_from_this<Timer<T>> {
   }
 
   /// Update the timer firing settings.
-  /// In SingleShot Mode:
-  ///   If the update is from a restart, will wait a full interval cycle.
-  ///   Otherwise, will use a 10 second "startup" value to allow changes to settle.
-  /// In Continuous Mode:
-  ///   An initial 10 second "startup" value is used to allow the system to settle
-  ///   and will then fire continuously on every interval cycle.
+  /// In SingleShot Mode, If the update is from a restart, will wait a full interval cycle.
```
russellhancox:
```suggestion
  /// In SingleShot Mode, if the update is from a restart, will wait a full interval cycle.
```

russellhancox (dismissed):

PR #613, opened 2025-10-21
Pin NPS domains and cert PEMs

mlw:


Source/santasyncservice/Pinning.mm
```diff
@@ -0,0 +1,209 @@
...
+    if ([host isEqualToString:domain] || [host hasSuffix:[@"." stringByAppendingString:domain]]) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+NSString *PinnedCertPEMs() {
+  return @"-----BEGIN CERTIFICATE-----\n"
```
russellhancox: I think it would be good to either have these in an array (that's merged together in this function) or to at least have a comment above each one describing it (Common name, fingerprint, expiry date)

Source/santasyncservice/Pinning.mm
```diff
@@ -0,0 +1,209 @@
...
+  @"workshop.cloud",
+  @"north-pole.tech",
+];
+
+bool IsDomainPinned(NSURL *url) {
+  NSString *host = url.host;
+
+  // Check if the given host or host suffix matches a pinned domain
+  for (NSString *domain in pinnedDomains) {
+    if ([host isEqualToString:domain] || [host hasSuffix:[@"." stringByAppendingString:domain]]) {
```
russellhancox: The first check is needless, Santa is never going to connect to either of these domains directly.

PR #620, opened 2025-10-27
NATS based push notifications for Sync V2

pmarkowsky:
This adds NATS push notifications for Sync v2 users.

Fixes SNT-143

### DIFF
--- a/Source/santasyncservice/SNTSyncManagerNATSTest.mm
+++ b/Source/santasyncservice/SNTSyncManagerNATSTest.mm
@@ -55,7 +55,8 @@
   OCMStub([self.mockConfigurator fcmEnabled]).andReturn(NO);
   OCMStub([self.mockConfigurator enableNATS]).andReturn(YES);
   OCMStub([self.mockConfigurator enableAPNS]).andReturn(NO);
-  OCMStub([self.mockConfigurator syncBaseURL]).andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
+  OCMStub([self.mockConfigurator syncBaseURL])
+      .andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
 
   // When: Sync manager is initialized
   self.syncManager = [[SNTSyncManager alloc] initWithDaemonConnection:self.mockDaemonConn];
@@ -70,7 +71,8 @@
   OCMStub([self.mockConfigurator fcmEnabled]).andReturn(YES);
   OCMStub([self.mockConfigurator enableNATS]).andReturn(YES);
   OCMStub([self.mockConfigurator enableAPNS]).andReturn(NO);
-  OCMStub([self.mockConfigurator syncBaseURL]).andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
+  OCMStub([self.mockConfigurator syncBaseURL])
+      .andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
 
   // When: Sync manager is initialized
   self.syncManager = [[SNTSyncManager alloc] initWithDaemonConnection:self.mockDaemonConn];
@@ -85,7 +87,8 @@
   OCMStub([self.mockConfigurator fcmEnabled]).andReturn(NO);
   OCMStub([self.mockConfigurator enableNATS]).andReturn(NO);
   OCMStub([self.mockConfigurator enableAPNS]).andReturn(YES);
-  OCMStub([self.mockConfigurator syncBaseURL]).andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
+  OCMStub([self.mockConfigurator syncBaseURL])
+      .andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
 
   // When: Sync manager is initialized
   self.syncManager = [[SNTSyncManager alloc] initWithDaemonConnection:self.mockDaemonConn];
@@ -100,7 +103,8 @@
   OCMStub([self.mockConfigurator fcmEnabled]).andReturn(NO);
   OCMStub([self.mockConfigurator enableNATS]).andReturn(NO);
   OCMStub([self.mockConfigurator enableAPNS]).andReturn(NO);
-  OCMStub([self.mockConfigurator syncBaseURL]).andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
+  OCMStub([self.mockConfigurator syncBaseURL])
+      .andReturn([NSURL URLWithString:@"https://example.workshop.cloud"]);
 
   // When: Sync manager is initialized
   self.syncManager = [[SNTSyncManager alloc] initWithDaemonConnection:self.mockDaemonConn];


deps/BUILD.nats
```diff
@@ -0,0 +1,41 @@
...
+    ],
+    copts = [
+        "-Wno-unused-function",
+        "-Wno-unused-variable",
+    ],
+    linkopts = [
+        "-framework", "Security",
+        "-lpthread",
+    ],
+)
```
mlw:
```suggestion
)

```

Source/santasyncservice/SNTPushClientNATSConnectionTest.mm
```diff
@@ -0,0 +1,252 @@
...
+#import <OCMock/OCMock.h>
+#import <XCTest/XCTest.h>
+
+#import "Source/common/SNTConfigurator.h"
+#import "Source/santasyncservice/SNTPushClientNATS.h"
+
+// Include NATS C client header
+extern "C" {
+#import "src/nats.h"
+}
```
mlw:
```suggestion
#import <OCMock/OCMock.h>
#import <XCTest/XCTest.h>
#include <sys/cdefs.h>

#import "Source/common/SNTConfigurator.h"
#import "Source/santasyncservice/SNTPushClientNATS.h"

__BEGIN_DECLS

#import "src/nats.h"

__END_DECLS
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+#import "Source/common/SNTConfigurator.h"
+#import "Source/common/SNTLogging.h"
+#import "Source/common/SNTSyncConstants.h"
+#import "Source/common/SNTSystemInfo.h"
+#import "Source/santasyncservice/SNTSyncState.h"
+
+// Include NATS C client header
+extern "C" {
+#import "src/nats.h"
+}
```
mlw:
```suggestion
#import <dispatch/dispatch.h>
#include <string.h>
#include <sys/cdefs.h>

#import "Source/common/SNTConfigurator.h"
#import "Source/common/SNTLogging.h"
#import "Source/common/SNTSyncConstants.h"
#import "Source/common/SNTSystemInfo.h"
#import "Source/santasyncservice/SNTSyncState.h"

__BEGIN_DECLS

#import "src/nats.h"

__END_DECLS
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+    if (status != NATS_OK) {
+      LOGE(@"NATS: Failed to subscribe to global tag topic: %s", natsStatus_GetText(status));
+    } else {
+      LOGI(@"NATS: Subscribed to global tag topic: %@", globalTagTopic);
+    }
+  }
+
+  // Subscribe to all tags from preflight: santa.tag.<tag>
+  if (self.tags && self.tags.count > 0) {
+    LOGD(@"NATS: Processing %lu tags from preflight", (unsigned long)self.tags.count);
```
russellhancox: Workshop is currently sending _all_ the topics to subscribe to, including the per-host tag and the global tag, with sanitization. I think it might be better to keep this in Workshop and keep the code in Santa simpler.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+    _fullSyncInterval = kDefaultPushNotificationsFullSyncInterval;
+    _connectionQueue =
+        dispatch_queue_create("com.northpolesec.santa.nats.connection", DISPATCH_QUEUE_SERIAL);
+    _messageQueue =
+        dispatch_queue_create("com.northpolesec.santa.nats.message", DISPATCH_QUEUE_SERIAL);
+    _isShuttingDown = NO;
+    _hasSyncedWithServer = NO;
+    _tagSubscriptions = [NSMutableArray array];
+    _currentRetryDelay = 1.0;  // Start with 1 second
+    _retryAttempt = 0;
```
mlw: Don't need to set defaults.
```suggestion
    _tagSubscriptions = [NSMutableArray array];
    _currentRetryDelay = 1.0;  // Start with 1 second
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+    if (self.isShuttingDown) return;
+
+    NSString *fullServer;
+#ifdef SANTA_FORCE_SYNC_V2
+    // In debug builds, allow overriding the domain suffix
+    fullServer = server;
+    LOGW(@"NATS: Domain suffix disabled for debugging - using server as-is: %@", fullServer);
+#else
+    // Always append domain suffix in release builds 
+    fullServer = [NSString stringWithFormat:@"%@.push.northpole.security", server];
```
russellhancox: The push_server field in preflight has the hostname, this should check that it contains push.northpole.security instead of always adding it.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+- (void)configureWithPushServer:(NSString *)server
+                      pushToken:(NSString *)token
+                            jwt:(NSString *)jwt
+                   pushDeviceID:(NSString *)deviceID
+                           tags:(NSArray<NSString *> *)tags {
+  dispatch_async(self.connectionQueue, ^{
+    if (self.isShuttingDown) return;
+
+    NSString *fullServer;
+#ifdef SANTA_FORCE_SYNC_V2
```
russellhancox: Why a new define instead of `#ifdef DEBUG`?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+- (void)connect {
+  dispatch_async(self.connectionQueue, ^{
+    if (self.isShuttingDown) return;
+
+    if (self.conn) {
+      LOGD(@"NATS already connected");
+      return;
+    }
+
+    // Check if we have teh necessary configuration to connect to the push service.
```
russellhancox:
```suggestion
    // Check if we have the necessary configuration to connect to the push service.
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+
+    // Set server URL with TLS unless debug mode is enabled
+    NSString *serverURL;
+
+#if defined(SANTA_FORCE_SYNC_V2) && defined(SANTA_NATS_DISABLE_TLS)
+    serverURL = [NSString stringWithFormat:@"nats://%@", self.pushServer];
+    LOGW(@"NATS: TLS disabled for debugging - using insecure connection on %@", serverURL);
+#else 
+      serverURL = [NSString stringWithFormat:@"tls://%@", self.pushServer];
+#endif
```
russellhancox: Right now, Workshop is sending the push server with the scheme. We can change it, but that does make testing rather easy, our test instances can send nats:// and a prod instance can default to tls://, then on the Santa side we'd just need to enforce that it's using tls:// in a prod build.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+    // Set error callback to catch subscription violations and other errors
+    natsOptions_SetErrorHandler(opts, &errorHandler, (__bridge void *)self);
+
+    // Set connection callbacks for better monitoring
+    natsOptions_SetDisconnectedCB(opts, &disconnectedCallback, (__bridge void *)self);
+    natsOptions_SetReconnectedCB(opts, &reconnectedCallback, (__bridge void *)self);
+    natsOptions_SetClosedCB(opts, &closedCallback, (__bridge void *)self);
+
+    // Create connection
+    status = natsConnection_Connect(&self->_conn, opts);
```
russellhancox:
```suggestion
    status = natsConnection_Connect(self.conn, opts);
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+    return;
+  }
+
+  dispatch_source_set_timer(
+      self.connectionRetryTimer,
+      dispatch_time(DISPATCH_TIME_NOW, (int64_t)(self.currentRetryDelay * NSEC_PER_SEC)),
+      DISPATCH_TIME_FOREVER,
+      (int64_t)(0.1 * NSEC_PER_SEC));  // 100ms leeway
+
+  __weak SNTPushClientNATS *weakSelf = self;
```
russellhancox: We have WEAKIFY/STRONGIFY macros for this in common/SNTStrengthify.h. Those shadow the variable so you can avoid having to refer directly to weakSelf or strongSelf

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I was following the design doc that said we'd only send the first part. Will change and update the doc.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I wanted to test production releases, but force V2.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Again this was left over from the design doc that said we weren't doing that. 

I'll make this change.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I'd forgotten about those. Will change.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+
+@interface SNTPushClientNATS ()
+@property(weak) id<SNTPushNotificationsSyncDelegate> syncDelegate;
+@property(nonatomic) natsConnection *conn;
+@property(nonatomic) natsSubscription *deviceSub;
+@property(nonatomic) natsSubscription *globalSub;
+@property(nonatomic)
+    NSMutableArray<NSValue *> *tagSubscriptions;        // Array of natsSubscription pointers
+@property(nonatomic) dispatch_queue_t connectionQueue;  // Single queue for connection management
+@property(nonatomic) dispatch_queue_t messageQueue;     // Queue for processing messages
```
mlw: style nit: some comments are on the line above, and some, like these, are inline. Would be nice to standardize

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I'd rather keep the sanitization in the client as well. If we make changes to the server. I still want to client to catch incorrect values.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+      dispatch_source_cancel(self.connectionRetryTimer);
+      self.connectionRetryTimer = nil;
+    }
+
+    // Subscribe to topics
+    [self subscribe];
+  });
+}
+
+- (void)disconnectWithCompletion:(void (^)(void))completion {
```
mlw, line 356: This doesn't appear to be called outside of tests. Is there an expectation it will get used?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,765 @@
...
+  SNTPushClientNATS *self = (__bridge SNTPushClientNATS *)closure;
+  LOGI(@"NATS: Reconnected to server");
+  dispatch_async(self.connectionQueue, ^{
+    self.isConnected = YES;
+
+    // Trigger sync with jitter to avoid thundering herd
+    // We might have missed push notifications while disconnected
+    if (!self.isShuttingDown) {
+      // Calculate jitter: random delay between 0 and 30 seconds
+      uint32_t jitterSeconds = arc4random_uniform(30);
```
mlw: nit: man page states that the result returned will be less than the upper bound provided.
```suggestion
      // Calculate jitter: random delay between 0 and 30 seconds
      uint32_t jitterSeconds = arc4random_uniform(31);
```

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -72,12 +73,8 @@ - (instancetype)initWithDaemonConnection:(MOLXPCConnection *)daemonConn {
...
     _daemonConn = daemonConn;
 
-    SNTConfigurator *config = [SNTConfigurator configurator];
-    if (config.enableAPNS) {
-      _pushNotifications = [[SNTPushClientAPNS alloc] initWithSyncDelegate:self];
-    } else if (config.fcmEnabled) {
-      _pushNotifications = [[SNTPushClientFCM alloc] initWithSyncDelegate:self];
-    }
+    // Always use NATS for push notifications
+    _pushNotifications = [[SNTPushClientNATS alloc] initWithSyncDelegate:self];
```
mlw: I didn't think we wanted to remove FCM support, did we?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,767 @@
...
+
+  LOGD(@"NATS: Received message on subject '%@': %@", msgSubject, msgData ?: @"<no data>");
+
+  // Process on message queue to serialize handling
+  dispatch_async(self.messageQueue, ^{
+    if (!self.isShuttingDown) {
+      LOGI(@"NATS: Triggering immediate sync due to message on %@", msgSubject);
+
+      // Queue sync to main thread
+      dispatch_async(dispatch_get_main_queue(), ^{
```
mlw, line 535: I'm not sure I follow the purpose/need for the nested serial queues - if necessary, could you expand the comment as to why this is being done?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,767 @@
...
+                                                                       withString:@""];
+
+      if (![sanitizedDeviceID isEqualToString:self.pushDeviceID]) {
+        LOGW(@"NATS: Sanitized device ID from '%@' to '%@' (removed periods/hyphens)",
+             self.pushDeviceID, sanitizedDeviceID);
+      }
+
+      if (sanitizedDeviceID.length == 0) {
+        LOGE(@"NATS: Device ID became empty after sanitization");
+        return;
```
mlw: Up above, if `self.pushDeviceID` isn't set in the if-condition check, there is a comment `// Don't return - continue with other subscriptions`. Also just below, if the `isValidNATSTopic` fails, the code also continues.

Here, if there is an error "sanitizing" the device ID, we return. Wanted to confirm this was intentional to treat an error differently and bail on other subscriptions.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I've updated this.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Changed this back because the signature is a pointer to a pointer. 

See https://github.com/nats-io/nats.c/blob/ae50b1cca2dd02955d4aa34b99b165bcaa695035/src/conn.c#L3359

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Done.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+  }
+
+  // Subscribe to all tags from preflight: santa.tag.<tag>
+  if (self.tags && self.tags.count > 0) {
+    LOGD(@"NATS: Processing %lu tags from preflight", (unsigned long)self.tags.count);
+
+    // Keep track of already subscribed topics to avoid duplicates
+    NSMutableSet *subscribedTopics = [NSMutableSet set];
+    if (deviceTopic) {
+      [subscribedTopics addObject:deviceTopic];
```
mlw: There are some code paths above where `deviceTopic` gets set, but then some error occurs such that it doesn't successfully subscribe. However the `deviceTopic` is not cleared, which would lead to it being wrongly tracked here as part of `subscribedTopics`

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+    }
+    [subscribedTopics addObject:globalTagTopic];
+
+    for (NSString *tag in self.tags) {
+      LOGD(@"NATS: Processing tag: '%@'", tag);
+
+      // Skip santa.host.* topics in tags - these should not be in the tags array
+      if ([tag hasPrefix:@"santa.host."]) {
```
mlw: The first check in `buildTopicFromTag:` does this same check. Can likely remove this one.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+        LOGW(@"NATS: Skipping host topic in tags array: %@ (host topics should not be in tags)",
+             tag);
+        continue;
+      }
+
+      // Build the topic from the tag
+      NSString *tagTopic = [self buildTopicFromTag:tag];
+
+      // Double-check the built topic doesn't start with santa.host.
+      if ([tagTopic hasPrefix:@"santa.host."]) {
```
mlw: This doesn't seem necessary Maybe `buildTopicFromTag:` could just return `nil` on error and check `tagTopic` against `nil`?

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: One queue is handling the connection lifecycle. The other message processing.  Given that there are two dimensions here, it seemed better to treat them separately.

I'll update the comment.

Source/santasyncservice/SNTPushClientNATSConnectionTest.mm
pmarkowsky: Happy to do this for convention. 

I originally put extern "C" because this would always be in a objective-C++/C++ context so figured there's no reason to hide the macro since it will always be true.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+      return;
+ }
+
+ // Make sure this starts with either nats:// or tls://
+ if (![server hasPrefix:@"tls://"]) {
+      LOGE(@"NATS: Invalid push server domain. Must start with 'tls://', got: %@", server);
+      return;
+ }
+
+  fullServer = server;
```
tburgin: This block seems to have lost its indentation.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+  // Validate suffix: must exist and cannot contain periods or hyphens
+  return suffix.length > 0 && [suffix rangeOfString:@"."].location == NSNotFound &&
+         [suffix rangeOfString:@"-"].location == NSNotFound;
+}
+
+- (void)subscribe {
+  if (self.isShuttingDown) return;
+
+  natsStatus status;
+  NSString *deviceTopic = nil;
```
tburgin:
```suggestion
  NSString *deviceTopic;
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+  // If tag already has santa.tag. prefix, extract just the tag part
+  if ([tag hasPrefix:@"santa.tag."]) {
+    processedTag = [tag substringFromIndex:10];  // Length of "santa.tag."
+    LOGD(@"NATS: Tag already has prefix, extracted: '%@'", processedTag);
+  }
+
+  // Strip hyphens and periods from tag for NATS compatibility
+  NSString *sanitizedTag = [processedTag stringByReplacingOccurrencesOfString:@"-" withString:@""];
+  sanitizedTag = [sanitizedTag stringByReplacingOccurrencesOfString:@"." withString:@""];
+  LOGD(@"NATS: Sanitized tag: '%@'", sanitizedTag);
```
tburgin: Consider doing the sanitizing in Workshop and just validating in Santa.

Source/santasyncservice/SNTPushClientNATS.mm
tburgin: Before getting to this comment I also added a similar comment about doing the sanitization in Workshop. The client can still validate to catch errors, but it doesn't need to be doing transformations on its own. Otherwise it opens up opportunity for Workshop and the client to make different transformation causing subtle issues. Workshop should be the source of truth here.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+    natsMsg_Destroy(msg);
+    return;
+  }
+
+  const char *subject = natsMsg_GetSubject(msg);
+  const char *data = natsMsg_GetData(msg);
+  int dataLen = natsMsg_GetDataLength(msg);
+
+  NSString *msgSubject = subject ? @(subject) : @"<unknown>";
+  NSString *msgData = nil;
```
tburgin:
```suggestion
  NSString *msgData;
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+  if (data && dataLen > 0) {
+    msgData = [[NSString alloc] initWithBytes:data length:dataLen encoding:NSUTF8StringEncoding];
```
tburgin, line 516: Do we know `natsMsg_GetDataLength` returns length using UTF8 encoding?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+  SNTPushClientNATS *self = (__bridge SNTPushClientNATS *)closure;
+  LOGI(@"NATS: Reconnected to server");
+  dispatch_async(self.connectionQueue, ^{
+    self.isConnected = YES;
+
+    // Trigger sync with jitter to avoid thundering herd
+    // We might have missed push notifications while disconnected
+    if (!self.isShuttingDown) {
+      // Calculate jitter: random delay between 0 and 30 seconds
+      uint32_t jitterSeconds = arc4random_uniform(30);
```
tburgin: We probably want more than 30 seconds here. In the event the cluster is unavailable and then comes back online, we will have all clients syncing within 30 seconds. I suggest spreading this out to about 10 min or so.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,785 @@
...
+    // Connection options
+    natsOptions_SetAllowReconnect(opts, true);
+    natsOptions_SetMaxReconnect(opts, -1);  // Infinite reconnects
+
+    // Set error callback to catch subscription violations and other errors
+    natsOptions_SetErrorHandler(opts, &errorHandler, (__bridge void *)self);
+
+    // Set connection callbacks for better monitoring
+    natsOptions_SetDisconnectedCB(opts, &disconnectedCallback, (__bridge void *)self);
+    natsOptions_SetReconnectedCB(opts, &reconnectedCallback, (__bridge void *)self);
```
tburgin, line 325: Can you explain the differences between the nats reconnect handler and our connect retry handler?

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Not really. I put this here for tests. If we handled sleep / graceful shutdown we could do this.  

Alternatively we could add this to dealloc.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Technically this is a byte count of the payload. Which is a collection of bytes. It's up to us what we put in here. 

I'd assumed initially we'd use UTF8 since that's easy for JSON etc.  and we're planning on sending empty messages.

tburgin: Perfect, just wanted to make sure it wasn't doing any encoding.

Source/santasyncservice/SNTSyncManager.mm
pmarkowsky: Fixed.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: The reconnect handler manages automatic recovery of lost connections and issues a jittered sync (0 - 90 seconds) just in case. It's invoked by the NATS library directly. 

The connect retry handler manages initial connection establishment failures with exponential backoff logic.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I've bumped it to 10 minutes.

Source/santasyncservice/SNTSyncManagerNATSTest.mm
```diff
@@ -0,0 +1,63 @@
...
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import <OCMock/OCMock.h>
+#import <XCTest/XCTest.h>
+
+#import "Source/common/MOLXPCConnection.h"
+#import "Source/common/SNTConfigurator.h"
+#import "Source/santasyncservice/SNTPushClientAPNS.h"
+#import "Source/santasyncservice/SNTPushClientFCM.h"
```
russellhancox: These aren't used

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+                           tags:(NSArray<NSString *> *)tags {
+  dispatch_async(self.connectionQueue, ^{
+    if (self.isShuttingDown) return;
+
+    NSString *fullServer;
+#ifdef DEBUG
+    // In debug builds, allow overriding the domain suffix
+    fullServer = server;
+    LOGW(@"NATS: Domain check disabled - using server as-is: %@", fullServer);
+#else
```
mlw: nit: Got a bit confused reading through this because `fullServer = server;` is also set outside of this `ifdef` down below. Based on the message, I was expecting some mutation to happen in the `#else` case, but that isn't what happens. Could probably just ditch this branch, and use `#ifndef` only (with a comment). But if the log message is desired, it might be good to remove the assignment to reduce confusion.
```suggestion
#ifdef DEBUG
    // In debug builds, allow overriding the domain suffix
    LOGW(@"NATS: Domain check disabled - using server as-is: %@", fullServer);
#else
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+    if (![server hasSuffix:@".push.northpole.security:443"]) {
+        LOGE(@"NATS: Invalid push server domain. Must end with '.push.northpole.security', got: %@", server);
+        return;
+    }
+
+    // Make sure this starts with either nats:// or tls://
+    if (![server hasPrefix:@"tls://"]) {
+        LOGE(@"NATS: Invalid push server domain. Must start with 'tls://', got: %@", server);
+        return;
+    }
```
mlw: The comment doesn't match the code - should the `nats://` scheme be here or is the comment incorrect?

```suggestion
    // Make sure this starts with either nats:// or tls://
    if (![server hasPrefix:@"tls://"] && ![server hasPrefix:@"nats://"]) {
      LOGE(@"NATS: Invalid push server domain. Must start with 'tls://' or 'nats://', got: %@",
           server);
      return;
    }
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+
+    // Now attempt to connect
+    [self connectIfConfigured];
+  } else {
+    NSMutableArray *missing = [NSMutableArray array];
+    if (!syncState.pushServer) [missing addObject:@"server"];
+    if (!syncState.pushNKey) [missing addObject:@"nkey"];
+    if (!syncState.pushJWT) [missing addObject:@"JWT"];
+    if (!syncState.pushDeviceID) [missing addObject:@"device ID"];
+    LOGW(@"NATS: Missing required push configuration from preflight: %@",
```
mlw, line 739: With the default value of `enableNATS` set to return `YES` from the configurator, it seems like this will result in this log warning message getting displayed all the time for non-WS deployments?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+  self.isRetrying = YES;
+  self.retryAttempt++;
+
+  // Calculate exponential backoff with jitter
+  // Base delay doubles each attempt: 1s, 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s
+  NSTimeInterval baseDelay = pow(2.0, MIN(self.retryAttempt - 1, 8));  // Cap at 2^8 = 256 seconds
+
+  // Add jitter: ±25% randomization to prevent thundering herd
+  double jitterFactor = 0.75 + (0.5 * ((double)arc4random_uniform(UINT32_MAX) / UINT32_MAX));
+  self.currentRetryDelay = baseDelay * jitterFactor;
```
mlw: This property isn't read outside of this method, and though the value changes, the value isn't used as input into calculations. This can instead be a local variable, would help simplify understanding this class a bit.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+  dispatch_async(self.connectionQueue, ^{
+    self.isConnected = NO;
+  });
+}
+
+// Schedule a connection retry with exponential backoff and jitter
+- (void)scheduleConnectionRetry {
+  if (self.isShuttingDown || self.isRetrying) return;
+
+  self.isRetrying = YES;
```
mlw, line 651: ~This if-condition check for `self.isRetrying` doesn't look correct. After the first retry, if we continue to fail to connect, it seems like the timer backoff won't increase.~

Update: Looks like this is toggled on event firing, so it does seem correct. This was difficult to follow.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+    self.isRetrying = NO;
+    return;
+  }
+
+  // Set a one shot timer to retry the connection with the calculated delay.
+  dispatch_source_set_timer(
+      self.connectionRetryTimer,
+      dispatch_time(DISPATCH_TIME_NOW, (int64_t)(self.currentRetryDelay * NSEC_PER_SEC)),
+      DISPATCH_TIME_FOREVER,
+      (int64_t)(0.1 * NSEC_PER_SEC));  // 100ms leeway
```
mlw: nit: might be more clear to do:

```suggestion
      (int64_t)(100 * NSEC_PER_MSEC));  // 100ms leeway
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+    // Set server URL with TLS unless debug mode is enabled
+    NSString *serverURL;
+
+#if defined(SANTA_FORCE_SYNC_V2) && defined(SANTA_NATS_DISABLE_TLS)
+    if (![self.pushServer hasPrefix:@"tls://"]) {
+      LOGE(@"NATS: Invalid push server domain. Must start with 'tls://', got: %@", self.pushServer);
+    }
+#else 
+   if (![self.pushServer hasPrefix:@"tls://"]) {
+      LOGE(@"NATS: Invalid push server domain. Must start with 'tls://', got: %@", self.pushServer);
```
mlw: Another section where `tls://` scheme is checked, but not `nats://`. Is this desired?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+  });
+}
+
+// Check if we have the necessary configuration to connect to the push service.
+- (BOOL)hasRequiredConfiguration {
+  return self.hasSyncedWithServer && self.pushServer && self.pushToken && self.jwt &&
+         self.pushDeviceID;
+}
+
+- (void)connectIfConfigured {
```
mlw: This method doesn't appear to buy much. The `hasRequiredConfiguration` is also checked in `connect`, which this method calls

Source/santasyncservice/SNTPushClientNATS.h
```diff
@@ -0,0 +1,26 @@
...
+
+@interface SNTPushClientNATS : NSObject <SNTPushNotificationsClientDelegate>
+- (instancetype)initWithSyncDelegate:(id<SNTPushNotificationsSyncDelegate>)syncDelegate;
+- (void)disconnectWithCompletion:(void (^)(void))completion;
+- (void)configureWithPushServer:(NSString *)server
+                      pushToken:(NSString *)token
+                            jwt:(NSString *)jwt
+                   pushDeviceID:(NSString *)deviceID
+                           tags:(NSArray<NSString *> *)tags;
+- (void)connectIfConfigured;
```
mlw: These don't seem to be part of the public API for this class. Can they be removed?
```suggestion
```

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+      [self unsubscribeAll];
+    }
+
+    // Store configuration
+    self.pushServer = fullServer;
+    self.pushToken = token;
+    self.jwt = jwt;
+    self.pushDeviceID = deviceID;
+    self.tags = tags;
+    self.hasSyncedWithServer = YES;
```
mlw: I don't think `hasSyncedWithServer` buys much. This only gets set with the rest of the configuration, and is only checked in `hasRequiredConfiguration`, along with the rest of the configuration.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+    LOGE(@"NATS: Permission/Subscription violation on subject: %s", subSubject);
+
+    // Check if this is a device subscription error
+    if (subSubject && strstr(subSubject, "santa.host")) {
+      LOGE(@"NATS: Device-specific subscription not permitted by server JWT");
+      // Don't mark connection as failed - other subscriptions might work
+    } else {
+      // Mark connection as failed if we get permission violations on other subjects
+      dispatch_async(self.connectionQueue, ^{
+        self.isConnected = NO;
```
mlw, line 574: Is this true? Are there cases where the error handler callback fires and the disconnected callback doesn't?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+  });
+}
+
+// NATS closed callback
+static void closedCallback(natsConnection *nc, void *closure) {
+  if (!closure) return;
+  SNTPushClientNATS *self = (__bridge SNTPushClientNATS *)closure;
+  LOGI(@"NATS: Connection closed");
+  dispatch_async(self.connectionQueue, ^{
+    self.isConnected = NO;
```
mlw, line 629: On closed, it seems like reconnects won't be attempted. Does the existing `self.conn` or anything about a previous connection state need to be cleaned up?

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+
+// Include NATS C client header
+#import "src/nats.h"
+
+__END_DECLS
+
+@interface SNTPushClientNATS ()
+@property(weak) id<SNTPushNotificationsSyncDelegate> syncDelegate;
+@property(nonatomic) natsConnection *conn;
+@property(nonatomic) natsSubscription *deviceSub;
```
mlw: This doesn't appear to be used anymore.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -0,0 +1,669 @@
...
+@property(nonatomic) natsSubscription *globalSub;
```
mlw: This doesn't appear to be used anymore either.

mlw (changes requested):

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: The check is correct as `nats://` is not encrypted. We only want to support encrypted sessions for non-debug builds.

mlw: That's fine, the comment should be updated then.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Yes, see previous comment.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Yes. If our permission check fails, we can connect but not have permission to publish to a given topic.

mlw: I think I'm confused still. Expanding the comment here might be very valuable. The piece I don't understand still is how we get into a state where we do get connected, and then we become disconnected without the disconnected callback firing. You mention "permission violations", but it's not clear to me what permissions, or what's lacking permissions. It's also not clear, from what I could find in the NATS C docs, under which conditions the connection is no longer considered connected when the error handler fires.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: I've updated the PR so that NATS is only initialized when we're using sync v2

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Comment updated.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: Good catch. This callback is only called when Nat's reconnect handler is exhausted. But if we get here it's not great.

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: An easy example is that Santa gets a JWT that's signed for NN-days.

That can expire on you.

pmarkowsky: I refactored this. Closing this comment.

Source/common/BUILD
```diff
@@ -425,6 +425,7 @@ objc_library(
         ":SNTRule",
         ":SNTStrengthify",
         ":SNTSystemInfo",
+        "//Source/santasyncservice:Pinning",
```
mlw: We shouldn't bring anything outside of common into common.

We shouldn't need to modify the configurator for this change. What if we just changed the `IsDomainPinned` function to be named something more straightforward - `DomainSupportsSyncV2` - and then in `SNTSyncManager`, we can just check:

```
} else if (config.enableNATS == YES && santa::DomainSupportsSyncV2(config.syncBaseURL)) {
```

mlw (approved):

PR #625, opened 2025-11-03
Make sync interval changes apply immediately

pmarkowsky:
This PR makes the changes to sync intervals apply immediately when using push notifications.

It also adds some basic unit tests for the changes.

### DIFF
--- a/Source/santasyncservice/BUILD
+++ b/Source/santasyncservice/BUILD
@@ -338,6 +338,7 @@ santa_unit_test(
     sdk_dylibs = ["libz"],
     deps = [
         ":FCM_lib",
+        ":NATS_lib",
         ":SNTPushNotificationsTracker",
         ":SNTSyncConfigBundle",
         ":SNTSyncLogging",


Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -304,7 +304,16 @@ - (SNTSyncStatusType)preflightWithSyncState:(SNTSyncState *)syncState {
...
+      NSUInteger oldInterval = self.pushNotifications.fullSyncInterval;
       [self.pushNotifications handlePreflightSyncState:syncState];
+
+      // If push interval changed, reschedule the timer immediately
+      if (oldInterval != self.pushNotifications.fullSyncInterval) {
+        LOGD(@"Push notification sync interval changed from %lu to %lu seconds. Rescheduling timer.",
+             oldInterval, self.pushNotifications.fullSyncInterval);
+      }
+      [self rescheduleTimerQueue:self.fullSyncTimer
+                  secondsFromNow:self.pushNotifications.fullSyncInterval];
```
tburgin: Did you want to always reschedule, or only when the interval changes? The current implementation always reschedules, but it is at odds with the comment and log.

Seeing as the non-push variation always reschedules, I suggest doing that to keep the behaviors aligned.

```suggestion
      [self rescheduleTimerQueue:self.fullSyncTimer
                  secondsFromNow:self.pushNotifications.fullSyncInterval];
```

mlw: An extra benefit of always rescheduling is that the sync time is tied to the trailing edge of sync. This means that if a sync takes a long time (e.g., because doing a clean sync and needing to download a lot of rules, or upload a bunch of events), the syncs don't stack up on each other. The reschedule will mean that a full sync interval will occur regardless of the time it took to sync.

pmarkowsky: I was aiming to rescheduling because it would add some natural jitter to syncs.

Source/santasyncservice/SNTSyncManagerTest.mm
```diff
@@ -0,0 +1,192 @@
...
+
+- (void)testPushClientAPNSDefaultInterval {
+  // Verify default initialization
+  SNTPushClientAPNS *pushClient = [[SNTPushClientAPNS alloc] initWithSyncDelegate:nil];
+
+  XCTAssertEqual(pushClient.fullSyncInterval, kDefaultPushNotificationsFullSyncInterval,
+                 @"Push client should initialize with default interval");
+}
+
+- (void)testPushClientAPNSIntervalUpdate {
```
mlw, line 53: It might be more useful to have the test pulled up one layer in the stack. All the tests below essentially check that `handlePreflightSyncState` modifies a value, but doesn't attempt to ensure timers are accounted for.

pmarkowsky: I added more tests to check for the timer changes.

Source/santasyncservice/BUILD
```diff
@@ -312,6 +312,65 @@ santa_unit_test(
...
+        "SNTSyncPostflight.h",
+        "SNTSyncPostflight.mm",
+        "SNTSyncPreflight.h",
+        "SNTSyncPreflight.mm",
+        "SNTSyncRuleDownload.h",
+        "SNTSyncRuleDownload.mm",
+    ],
+    sdk_dylibs = ["libz"],
+    deps = [
+        ":FCM_lib",
```
russellhancox, line 340:
```suggestion
        ":FCM_lib",
        ":NATS_lib",
```

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -312,13 +312,24 @@ - (SNTSyncStatusType)preflightWithSyncState:(SNTSyncState *)syncState {
     // Start listening for push notifications with a full sync every
     // pushNotificationsFullSyncInterval.
     if (self.pushNotifications) {
+      NSUInteger oldInterval = self.pushNotifications.fullSyncInterval;
       [self.pushNotifications handlePreflightSyncState:syncState];
+
+      // If push interval changed, mark log the difference.
```
russellhancox:
```suggestion
      // If push interval changed, log the difference.
```

PR #626, opened 2025-11-04
Adopt mode transition sync protocol changes

mlw:
This PR handles parsing out the mode transition message and getting the config over to the daemon to save in the sync state.

### DIFF
--- a/Source/santasyncservice/SNTSyncPreflight.mm
+++ b/Source/santasyncservice/SNTSyncPreflight.mm
@@ -20,6 +20,7 @@
 #import "Source/common/SNTConfigurator.h"
 #import "Source/common/SNTExportConfiguration.h"
 #import "Source/common/SNTLogging.h"
+#import "Source/common/SNTModeTransition.h"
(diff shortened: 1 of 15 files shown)


Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -281,6 +285,32 @@ BOOL Preflight(SNTSyncPreflight *self, google::protobuf::Arena *arena,
     }
   }
 
+  if constexpr (IsV2) {
```
russellhancox, line 288: This method is very long. I'm OK if it's a TODO for now, but we should break it up a bit.

It might also be nice to unify the multiple "if V2" sections.

russellhancox (approved):

Source/santasyncservice/SNTSyncPreflight.mm
mlw: Sounds good, will noodle on some ideas to split up.

PR #632, opened 2025-11-10
Rename EnableNATS to EnablePushNotifications

pmarkowsky:
This renames the configuration variable while also maintaining backwards compatibility.

### DIFF
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -444,8 +444,8 @@ objc_library(
         ":SNTRule",
         ":SNTStrengthify",
         ":SNTSystemInfo",
-        "//Source/santasyncservice:Pinning",
         ":SystemResources",
+        "//Source/santasyncservice:Pinning",
     ],
 )
 


Source/common/SNTConfigurator.mm
```diff
@@ -1668,6 +1686,12 @@ - (NSMutableDictionary *)readForcedConfig {
 
   [self applyOverrides:forcedConfig];
 
+  // Backward compatibility: if EnableNATS is set but EnablePushNotifications is not,
+  // copy the value from EnableNATS to EnablePushNotifications
+  if (forcedConfig[kEnableNATS] != nil && forcedConfig[kEnablePushNotifications] == nil) {
+    forcedConfig[kEnablePushNotifications] = forcedConfig[kEnableNATS];
+  }
```
mlw, line 1815: We should apply overrides last
```suggestion
  // Backward compatibility: if EnableNATS is set but EnablePushNotifications is not,
  // copy the value from EnableNATS to EnablePushNotifications
  if (forcedConfig[kEnableNATS] != nil && forcedConfig[kEnablePushNotifications] == nil) {
    forcedConfig[kEnablePushNotifications] = forcedConfig[kEnableNATS];
  }
  
  [self applyOverrides:forcedConfig];
```

Source/common/SNTConfigurator.h
```diff
@@ -777,7 +777,17 @@
...
+@property(readonly, nonatomic) BOOL enablePushNotifications;
+
+///
+///  Deprecated: Use enablePushNotifications instead.
+///  Set to true to use NATS push notifications. Defaults to true.
+///  NATS takes precedence over APNS if enabled, but not FCM.
+///
+//   This can only used with Workshop and SyncV2.
+///
+@property(readonly, nonatomic) BOOL enableNATS
```
mlw: Does this property still need to exist? It isn't used anymore and the config key for it read from `enablePushNotifications`.

mlw (changes requested):

Source/common/SNTConfigurator.mm
```diff
@@ -1326,9 +1330,19 @@ - (BOOL)enableAPNS {
...
+  NSNumber *value = self.configState[kEnablePushNotifications];
+  if (value != nil) {
+    return [value boolValue];
+  }
+  NSNumber *deprecatedValue = self.configState[kEnableNATS];
+  if (deprecatedValue != nil) {
+    return [deprecatedValue boolValue];
+  }
+  // Default to true when neither key is set, as documented
+  return YES;
```
mlw: This seems to be a change from how `EnableNATS` worked. Old code would default to `NO` if not set:
```
- (BOOL)enableNATS {
  // TODO: Consider supporting enablement from the sync server.
  return [self.configState[kEnableNATS] boolValue];
}
```
I recall this was originally enabled by default, but we had decided to change that logic and require an active config option.

pmarkowsky: >I recall this was originally enabled by default, but we had decided to change that logic and require an active config option

This was because of the interval handling. It would have prevented changes from applying. Since that's fixed I was making that change now, for 2025.12 / 2026.1.

mlw: Thinking about this more, I don't think we can enable by default. This will change all the default sync interval timings, even for people that will never have push notifications.

pmarkowsky: Good point. I'll change it back to default to false.

pmarkowsky: In the latest commit, I've split the difference so that if you're using Workshop and haven't set anything this will default to true as of the next release, if you're not using Workshop then it will default to false. 

If we're good with this I'll update the docs to reflect this.

mlw (dismissed):

PR #634, opened 2025-11-11
ci: Stop running on macos-13, fix lint.sh

russellhancox:
macOS 13 builders are being deprecated and having increasing brownouts (to discourage further use) so we need to disable them.

The latest commits of buildifier moved the actual command into a `cmd` subfolder, so we're pinning back to the latest stable release.

### DIFF
--- a/Testing/lint.sh
+++ b/Testing/lint.sh
@@ -16,7 +16,6 @@ swift format lint -s -r ${GIT_ROOT}
 ! git grep -EIn $'[ \t]+$' -- ':(exclude)*.patch'
 
 GBIN=${GOBIN:-${GOPATH:-${HOME}/go}/bin}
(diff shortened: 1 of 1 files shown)


Testing/lint.sh
```diff
@@ -15,6 +15,6 @@ swift format lint -s -r ${GIT_ROOT}
 
 ! git grep -EIn $'[ \t]+$' -- ':(exclude)*.patch'
 
-go install github.com/bazelbuild/buildtools/buildifier@latest
+go install github.com/bazelbuild/buildtools/buildifier/cmd@latest
```
tburgin: Can we pin to a specific version?

russellhancox: Sure, we can pin to the latest release (v8.2.1) which also negates the need to use the cmd subdir and rename.

PR #638, opened 2025-11-14
santactl command to temporarily enter Monitor Mode if eligible

mlw:
Adds a new `santactl monitormode` command that will enable previously-authorized hosts to enter Monitor Mode, on-demand, for some configured duration.

### DIFF
--- a/Source/santactl/Commands/SNTCommandMonitorMode.mm
+++ b/Source/santactl/Commands/SNTCommandMonitorMode.mm
@@ -1,0 +1,121 @@
+/// Copyright 2025 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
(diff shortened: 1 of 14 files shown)


Source/santactl/Commands/SNTCommandMonitorMode.mm
```diff
@@ -0,0 +1,121 @@
...
+                                              return;
+                                            }
+
+                                            TEE_LOGI(
+                                                @"Temporarily switching to Monitor Mode for %u %@",
+                                                minutes, minutes > 1 ? @"minutes" : @"minute");
+                                          }];
+  }
+
+  exit(success ? EXIT_SUCCESS : EXIT_FAILURE);
```
russellhancox, line 118: I don't know if it's _guaranteed_ that the reply block has finished executing when a method call on a synchronous remote object proxy returns. It might be safer to move the exit() call into the reply block to ensure the log lines are always printed.

Source/santad/SNTDaemonControlController.mm
```diff
@@ -57,10 +58,16 @@
 double watchdogCPUPeak = 0;
 double watchdogRAMPeak = 0;
 
+// Semi-arbitrary limits for temporary Monitor Mode.
+// Min: 1 minute. Max: 30 days.
+static constexpr uint32_t kMinTempMonitorModeMinutes = 1;
+static constexpr uint32_t kMaxTempMonitorModeMinutes = 1 * 60 * 24 * 30;
+
```
russellhancox, line 65: These already exist in SNTModeTransition.mm, don't they?

Source/santactl/Commands/SNTCommandMonitorMode.mm
mlw: Spoke offline, tested this works as expected.

Source/santad/SNTDaemonControlController.mm
mlw: Good catch. Will consolidate in the next PR.

PR #640, opened 2025-11-15
santactl status now displays temporary Monitor Mode time remaining

mlw:
Some sample outputs:
```
  Mode                      | Temporary Monitor Mode (About 57 seconds remaining)
```

```
  Mode                      | Temporary Monitor Mode (About 1 hour, 21 minutes remaining)
```

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.mm
+++ b/Source/santactl/Commands/SNTCommandStatus.mm
@@ -33,36 +33,17 @@ NSString *StartupOptionToString(SNTDeviceManagerStartupPreferences pref) {
   }
 }
 
-NSString *Plural(uint64_t val, NSString *base) {
-  if (val == 1) {
-    return base;
-  } else {
-    return [base stringByAppendingString:@"s"];
-  }
-}
+NSString *FormatTimeRemaining(NSTimeInterval seconds) {
+  NSDateComponentsFormatter *formatter = [[NSDateComponentsFormatter alloc] init];
+  formatter.unitsStyle = NSDateComponentsFormatterUnitsStyleFull;
+  formatter.allowedUnits =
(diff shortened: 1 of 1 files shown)


Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -33,6 +33,38 @@
...
+                                        Plural(hours, @"hour"), remainingMinutes,
+                                        Plural(remainingMinutes, @"minute")];
+    } else {
+      return [NSString stringWithFormat:@"About %llu %@ remaining", hours, Plural(hours, @"hour")];
+    }
+  } else {
+    return
+        [NSString stringWithFormat:@"About %llu %@ remaining", minutes, Plural(minutes, @"minute")];
+  }
+}
```
pmarkowsky: Can't we just use  [localizedStringFromTimeInterval]( https://developer.apple.com/documentation/foundation/relativedatetimeformatter/localizedstring(fromtimeinterval:)?language=objc)

e.g. Something like

```C
- (NSString *)FormatTimeRemaining:(NSTimeInterval)seconds {
    NSDateComponentsFormatter *formatter = [[NSDateComponentsFormatter alloc] init];
    formatter.unitsStyle = NSDateComponentsFormatterUnitsStyleFull;
    formatter.allowedUnits = NSCalendarUnitDay | NSCalendarUnitHour | 
                             NSCalendarUnitMinute | NSCalendarUnitSecond;
    formatter.maximumUnitCount = 2;
    formatter.collapsesLargestUnit = NO;
    
    return [formatter stringFromTimeInterval:seconds];
}
```

pmarkowsky (commented):
Curious if we can just use the standard lib.

Source/santactl/Commands/SNTCommandStatus.mm
mlw: Neat! didn't know about this API.

PR #643, opened 2025-11-17
Move pinning code to common where it belongs

mlw:
This will keep our code boundaries cleaner so non-common code isn't pulled into common.


pmarkowsky (approved):
LGTM.

PR #649, opened 2025-11-19
Add sync intervals to santactl status with a human-readable output

pmarkowsky:
This adds the following fields to `santactl status`

Screenshot 2025-11-19 at 10 22 08

When using JSON the number of seconds is used instead 

```json
"sync" : {
    "full_sync_interval_seconds" : 600,
    "bundle_scanning" : true,
    "execution_rules_hash" : "558fe44758a7a2d3140aa895a52113db",
    "server" : "https://mtls.xxx.workshop.cloud/santa/",
    "file_access_rules_hash" : "bfbfa645bf2235f419954b51635ec32d",
    "events_pending_upload" : 14,
    "last_successful_rule" : "2025/11/18 23:58:03 -0500",
    "push_notifications" : "Connected",
    "push_notifications_full_sync_interval_seconds" : 14400,
    "enabled" : true,
    "clean_required" : false,
    "last_successful_full" : "2025/11/18 23:58:03 -0500"
  },
```

Fixes SNT-178

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.mm
+++ b/Source/santactl/Commands/SNTCommandStatus.mm
@@ -416,12 +416,18 @@ REGISTER_COMMAND_NAME(@"status")
       printf("  %-25s | %s\n", "Bundle Scanning", (enableBundles ? "Yes" : "No"));
       printf("  %-25s | %lld\n", "Events Pending Upload", eventCount);
       printf("  %-25s | %s\n", "Execution Rules Hash", [executionRulesHash UTF8String]);
-      printf("  %-25s | %s\n", "Full Sync Interval", [FormatInterval(fullSyncInterval) UTF8String]);
+
+      // If push notifications are enabled, show the push notifications full
+      // sync interval since it's the active configuration.
+      NSString *fullSyncIntervalStr = FormatInterval(fullSyncInterval);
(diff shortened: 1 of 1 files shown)


Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -239,6 +251,8 @@ - (void)runWithArguments:(NSArray *)arguments {
           ?: @"Never";
 
   NSString *syncURLStr = configurator.syncBaseURL.absoluteString;
+  NSUInteger fullSyncInterval = configurator.fullSyncInterval;
+  NSUInteger pushNotificationsFullSyncInterval = configurator.pushNotificationsFullSyncInterval;
```
mlw, line 255: I think it might be helpful to only show the active interval

pmarkowsky: I showed both for debugging purposes. E.g. if you're having issues with with syncs and push notifications seemed easier to debug if you could see both values side by side..  

Happy to not show the second interval if push notifications are not enabled.

pmarkowsky: I've updated this to only print out if you have push notifications enabled.

PR #652, opened 2025-11-20
Fix santactl push notification status with NPS Push Service

pmarkowsky:
This fixes the santactl status output when connected to the NPS Push Notification service it also adds a field for the server you're connected to as passed by Workshop.

e.g. 

Screenshot 2025-11-19 at 21 47 10

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.mm
+++ b/Source/santactl/Commands/SNTCommandStatus.mm
@@ -178,7 +178,7 @@ REGISTER_COMMAND_NAME(@"status")
 
   __block NSString *pushNotifications = @"Unknown";
   __block NSString *pushServerAddress = nil;
-  if ([[SNTConfigurator configurator] syncBaseURL]) {
+  if ([configurator syncBaseURL]) {
     // The request to santad to discover whether push notifications are enabled
     // makes a call to santasyncservice. If it's unavailable the call can hang
     // so we run the request asynchronously with a semaphore timer; if we have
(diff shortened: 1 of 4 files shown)


Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -189,6 +190,30 @@ - (void)runWithArguments:(NSArray *)arguments {
...
+                if ([cleanedAddress hasPrefix:@"tls://"]) {
+                  cleanedAddress = [cleanedAddress substringFromIndex:6];
+                }
+                // Remove port number (e.g., :443)
+                NSRange portRange = [cleanedAddress rangeOfString:@":"];
+                if (portRange.location != NSNotFound) {
+                  cleanedAddress = [cleanedAddress substringToIndex:portRange.location];
+                }
+                pushServerAddress = cleanedAddress;
+              }
```
mlw, line 210: If the address looks like `scheme://my.server.com:port`, you can extract the `my.server.com` part via:
```suggestion
              if (serverAddress) {
                pushServerAddress = [[NSURL URLWithString:serverAddress] host];
              }
```

Source/santasyncservice/SNTPushClientNATS.h
```diff
@@ -16,4 +16,5 @@
 
 @interface SNTPushClientNATS : NSObject <SNTPushNotificationsClientDelegate>
 - (instancetype)initWithSyncDelegate:(id<SNTPushNotificationsSyncDelegate>)syncDelegate;
+- (NSString *)pushServerAddress;
```
mlw: We should just expose the `pushServer` property as `readonly`, then we could ditch the `pushServerAddress` impl in the `.mm` file.

PR #653, opened 2025-11-20
Add CEL rules to prevent enabling SSH and Remote Apple Events

pmarkowsky:
This PR adds example CEL rules for https://www.loobins.io/binaries/systemsetup/

### DIFF
--- a/docs/docs/configuration/faa.md
+++ b/docs/docs/configuration/faa.md
@@ -4,4 +4,222 @@ sidebar_position: 2
 
 # File-Access Authorization
 
-Coming soon!
+File Access Authorization (FAA) policies are defined using a plist configuration
+file. The policy can be specified either in a
+[separate file](/configuration/keys#FileAccessPolicyPlist) or
+[in-line](/configuration/keys#FileAccessPolicy) with the rest of the Santa
+configuration.
+
+If the policy is specified in a separate file, Santa will periodically re-read
+this file. By default this will occur every 10 minutes but the interval can be
+[overridden](/configuration/keys#FileAccessPolicyUpdateIntervalSec).
+
+## Policy Structure
+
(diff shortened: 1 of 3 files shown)


docs/docs/cookbook/cel.md
```diff
@@ -75,3 +78,27 @@ Cloud SDK installer and AI tools like claude code use osascript.
...
+## Prevent users from enabling SSH (Santa 2025.8+)
+
+As called out in [loobins](https://www.loobins.io/binaries/systemsetup/) the
+systemsetup command can be used to enable SSH.
+
+To block this create a signing ID rule for `platform:com.apple.systemsetup` and
+attach the following CEL program:
+
+```clike
+args.join(" ").contains("-setremotelogin on")
```
russellhancox, line 91: This would _only allow_ executions of systemsetup that are enabling SSH - a bool return of true is treated as ALLOWLIST. Better to be explicit like the other rules we have

```suggestion
args.join(" ").contains("-setremotelogin on") ? BLOCKLIST : ALLOWLIST
```

docs/docs/cookbook/cel.md
```diff
@@ -75,3 +78,27 @@ Cloud SDK installer and AI tools like claude code use osascript.
...
+```
+
+## Prevent users from enabling Remote Apple Event (Santa 2025.8+)
+
+As called out in [loobins](https://www.loobins.io/binaries/systemsetup/) the
+systemsetup command can be used to enable Remote Apple Events.
+
+To block this create a signing ID rule for `platform:com.apple.systemsetup` and
+attach the following CEL program:
```
russellhancox, line 100: I think this might be the first case of multiple cookbook rules targeting the same binary. It might be nice to show how to combine these 2 new rules together.

PR #660, opened 2025-11-23
Refactor temporary monitor mode logic

mlw:
This consolidates all the temporary monitor mode (TMM) logic into a single class. This pulls code out of the daemon controller and configurator. The configurator is still responsible for maintaining the client mode state when in TMM and saving/maintaining the state file when entering/leaving TMM - but it is not responsible for the structure/contents of the TMM state. The daemon controller methods are still needed as the entry point into TMM, but are essentially pass-thrus. This consolidation will make the audit events much simpler.

### DIFF
--- a/Source/santad/Info.plist
+++ b/Source/santad/Info.plist
@@ -1,5 +1,6 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
+	"http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
 	<key>CFBundleIdentifier</key>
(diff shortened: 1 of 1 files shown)


Source/common/PassKey.h
```diff
@@ -35,3 +38,5 @@ class PassKey {
  protected:
   static PassKey<T> MakeKey() { return PassKey<T>{}; }
```
pmarkowsky, line 44: nit: We might want to choose another name so as not to confuse would be contributors that this is somehow  WebAuthn related.

Source/santad/TemporaryMonitorMode.h
```diff
@@ -0,0 +1,93 @@
...
+#define SANTA__SANTAD__TEMPORARYMONITORMODE_H
+
+namespace santa {
+
+class TemporaryMonitorMode : public Timer<TemporaryMonitorMode>,
+                             public PassKey<TemporaryMonitorMode> {
+ public:
+  // Factory
+  static std::shared_ptr<TemporaryMonitorMode> Create(SNTConfigurator *configurator,
+                                                      SNTNotificationQueue *not_queue);
```
pmarkowsky: nit: the names a little funny e.g. the `not_queue` which is in fact a queue.

Source/santad/TemporaryMonitorMode.h
```diff
@@ -0,0 +1,93 @@
...
+  // Cancel an existing temporary Monitor Mode session.
+  // Return true if a session was active, otherwise false.
+  bool Cancel();
+
+  // Cancel an existing temporary Monitor Mode session and revoke any
+  // stored mode transition sync setting to prevent future sessions.
+  // Return true if a session was active, otherwise false.
+  bool Revoke();
+
+  bool OnTimer();
```
pmarkowsky, line 62: nit: Missing docstring.  Something revocation on timer.

pmarkowsky (approved):
Looks good, minor nits. Mostly wondering if we should change the name of the PassKey class as it could be confusing in the future if we do anything with WebAuthn.

Source/common/PassKey.h
pmarkowsky: We should add a link in the comments that this is the PassKey pattern https://playfulprogramming.com/posts/a-forgotten-idiom-revisited-pass-key

Source/santad/TemporaryMonitorMode.h
mlw: haha fair point, will change

pmarkowsky (approved):
LGTM.

PR #665, opened 2025-11-24
docs: Delete unused package-lock.json

russellhancox:
This was accidentally checked in when originally migrating to docusaurus and while it hasn't ever been used, Dependabot is still reading it and creating advisories that don't apply


pmarkowsky (approved):
LGTM.

PR #668, opened 2025-11-24
Add support for firing a webhook when an FAA rule triggers.

pmarkowsky:
This PR adds an option to FAA that will cause a web request to be made any time an FAA rule match occurs. 

Note this is a non-interactive web request. This supports similar templating as the event details URL. It can be used to allow FAA to watch files and trigger an action in other systems when they're  matched.


Source/common/faa/WatchItems.mm
```diff
@@ -539,6 +552,9 @@ bool ParseConfigSingleWatchItem(NSString *name, std::string_view fallback_policy
     return true;
   }
 
+  NSString *webhook_url = options[kWatchItemConfigKeyOptionsWebhookURL];
+  NSDictionary *webhook_headers = options[kWatchItemConfigKeyOptionsWebhookHeaders];
```
mlw: nit: delete. access directly in the constructors below

PR #672, opened 2025-11-27
Emit audit events from TemporaryMonitorMode

mlw:
This handles emitting audit events for the various enter/leave operations with appropriate reasons. The next part will take care of hooking up the newly emitted stored events to the EventUploadRequests.


pmarkowsky (approved):
LGTM.

PR #675, opened 2025-11-28
Add support for time duration strings.

pmarkowsky:
This PR extends the `--duration` flag in `santactl monitormode` to support for parsing simple duration strings 

Specifically it adds support for  the following durations:

- `m` - minute e.g. `10m` this is basically ignored since the default is minutes
- `h` -  hours e.g. `2h` this gives you a short hand rather than typing 120
- `d`   - day - e.g. `1d` this is short hand for multiplying by 1440.

Screenshot 2025-11-27 at 18 52 15

### DIFF
--- a/Source/santactl/Commands/SNTCommandMonitorMode.mm
+++ b/Source/santactl/Commands/SNTCommandMonitorMode.mm
@@ -61,36 +61,18 @@ REGISTER_COMMAND_NAME(@"monitormode")
   NSString *unit = nil;
 
   NSInteger intValue = 0;
-  NSUInteger locationBeforeScan = scanner.scanLocation;
   if ([scanner scanInteger:&intValue]) {
-    NSUInteger locationAfterInteger = scanner.scanLocation;
-
-    // Verify that we actually scanned something and didn't skip whitespace
-    if (locationAfterInteger == locationBeforeScan) {
-      return 0;  // Invalid: no integer found
-    }
-
     // Check if we're at the end (no unit specified)
     if ([scanner isAtEnd]) {
       return intValue;
     }
 
-    // Verify the next character is immediately adjacent (no whitespace)
-    unichar nextChar = [duration characterAtIndex:locationAfterInteger];
-    if ([[NSCharacterSet whitespaceAndNewlineCharacterSet] characterIsMember:nextChar]) {
-      return 0;  // Invalid: whitespace between integer and unit
-    }
-
     // Scan exactly one character from the unit set
     NSString *scannedUnit = nil;
-    NSUInteger locationBeforeUnit = scanner.scanLocation;
     if ([scanner scanCharactersFromSet:[NSCharacterSet characterSetWithCharactersInString:@"smhd"]
                             intoString:&scannedUnit]) {
-      NSUInteger locationAfterUnit = scanner.scanLocation;
-
       // Ensure unit is exactly one character and we're at the end
-      if (scannedUnit.length == 1 && locationAfterUnit == locationBeforeUnit + 1 &&
-          [scanner isAtEnd]) {
+      if (scannedUnit.length == 1 && [scanner isAtEnd]) {
         unit = scannedUnit;
       } else {
         return 0;  // Invalid: unit is not exactly one char or there's more content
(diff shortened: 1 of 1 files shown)


Source/santactl/Commands/SNTCommandMonitorMode.mm
```diff
@@ -53,8 +53,36 @@ + (NSString *)longHelpText {
...
+  if ([scanner scanInteger:&intValue]) {
+    [scanner scanCharactersFromSet:[NSCharacterSet characterSetWithCharactersInString:@"smhd"]
+                        intoString:&unit];
+    if (unit == nil) {
+      return intValue;
+    }
+
+    if ([unit isEqualToString:@"s"]) {
+      // Round up and properly handle truncation: e.g. 61s -> 2 minutes
+      return (intValue + 59) / 60;
```
mlw: I don't think we should support seconds. I think it'd be confusing when you request some duration and a longer duration is returned. If we really want to support seconds, we should change all the internal interfaces to support it - from santactl, to the daemon control controller, to the TMM class.

Source/santactl/Commands/SNTCommandMonitorMode.mm
```diff
@@ -53,8 +53,36 @@ + (NSString *)longHelpText {
...
+  NSScanner *scanner = [NSScanner scannerWithString:duration];
+  NSString *unit = nil;
+
+  NSInteger intValue = 0;
+  if ([scanner scanInteger:&intValue]) {
+    [scanner scanCharactersFromSet:[NSCharacterSet characterSetWithCharactersInString:@"smhd"]
+                        intoString:&unit];
+    if (unit == nil) {
+      return intValue;
+    }
```
mlw, line 68: Error handling should be expanded here. Did some testing on the side since I've never really played with NSScanner. But it is pretty easy to pass a value to this method that returns an unexpected result. E.g. variants around these inputs:
```
100.m --> parses as: 100s
100dxm --> parses as 100d
```
I think we could maybe use the `atEnd` property to confirm we've consumed input, and error out if not.

```suggestion
  if ([scanner scanInteger:&intValue]) {
    [scanner scanCharactersFromSet:[NSCharacterSet characterSetWithCharactersInString:@"smhd"]
                        intoString:&unit];
    if (!scanner.atEnd) {
      return 0;
    }
    if (unit == nil) {
      return intValue;
    }
```

Source/santactl/Commands/SNTCommandMonitorMode.mm
```diff
@@ -53,8 +53,36 @@ + (NSString *)longHelpText {
   return [NSSet setWithArray:@[ @"mm" ]];
 }
 
+// Parse a time interval string into a number of minutes.
+// e.g. "10m" -> 10, "2h" -> 120, "3d" -> 3600
+- (NSTimeInterval)parseTimeInterval:(NSString *)duration {
```
mlw, line 58: style nit: `ParseTimeInterval`

russellhancox: No, objc methods start lowercase

Source/santactl/Commands/SNTCommandMonitorMode.mm
pmarkowsky: I honestly threw it in there to see what people thought. Will remove.

Source/santactl/Commands/SNTCommandMonitorMode.mm
mlw: ope my bad

Source/santactl/Commands/SNTCommandMonitorMode.mm
```diff
@@ -53,8 +53,67 @@ + (NSString *)longHelpText {
...
+  NSScanner *scanner = [NSScanner scannerWithString:duration];
+  scanner.charactersToBeSkipped = nil;
+  NSString *unit = nil;
+
+  NSInteger intValue = 0;
+  NSUInteger locationBeforeScan = scanner.scanLocation;
+  if ([scanner scanInteger:&intValue]) {
+    NSUInteger locationAfterInteger = scanner.scanLocation;
```
mlw: This is overkill... I don't think it's possible for the scan location to not progress if success is returned, rendering much of the new code here inaccessible. Think the previous commit is fine: https://github.com/northpolesec/santa/pull/675/commits/7ea8cfe2da186293f298c8adcd3b375f68353fd2

PR #683, opened 2025-12-04
Fix some memory leaks in the sync service

mlw:
This addresses a retain cycle that was causing a leak. Additionally, some configuration settings could trigger an additional small leak.


pmarkowsky (approved):
LGTM.

russellhancox (approved):

PR #690, opened 2025-12-09
Add support for proc suspend/resume events

mlw:
For now this is gated by a config option (`EnableAntiTamperProcessSuspendResume`) and is off by default. In the near future, this will be on by default and eventually the key will be removed entirely.


pmarkowsky (approved):
LGTM.

PR #693, opened 2025-12-10
Helper code sign identifier utilities

mlw:
This code exists in some form throughout the code base already. Pulling into a single location to prevent further duplication.

Part of WS-487

### DIFF
--- a/Source/common/CodeSigningIdentifierUtils.mm
+++ b/Source/common/CodeSigningIdentifierUtils.mm
@@ -31,7 +31,7 @@ bool IsValidTeamID(NSString *tid) {
 }
 
 bool IsValidSigningID(NSString *sid) {
-  if ([sid hasPrefix:kPlatformTeamIDPrefix] && sid.length > kPlatformTeamIDPrefix.length) {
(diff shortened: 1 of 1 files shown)


Source/common/CodeSigningIdentifierUtils.mm
```diff
@@ -0,0 +1,62 @@
...
+bool IsValidTeamID(NSString *tid) {
+  static NSCharacterSet *nonAlnum = [[NSCharacterSet
+      characterSetWithCharactersInString:
+          @"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"] invertedSet];
+  return tid.length == kTeamIDLength &&
+         [tid rangeOfCharacterFromSet:nonAlnum].location == NSNotFound;
+}
+
+bool IsValidSigningID(NSString *sid) {
+  if ([sid hasPrefix:kPlatformTeamIDPrefix] && sid.length > kPlatformTeamIDPrefix.length) {
```
russellhancox: Checking the length should be much quicker than checking for the prefix so let's check that first.

```suggestion
  if (sid.length > kPlatformTeamIDPrefix.length && [sid hasPrefix:kPlatformTeamIDPrefix]) {
```

PR #694, opened 2025-12-11
Add classes to support kill command

mlw:
Classes the will be used in support of kill command.

Part of WS-487


pmarkowsky (approved):
LGTM.

PR #695, opened 2025-12-11
Fix rule download issue where success/failure was improperly determined

mlw:
Success/failure is defined by the result of the bool method, not the presence of NSError, which is also used for warnings. This change also allows bubbling up multiple errors that can be reported.

### DIFF
--- a/Source/santad/SNTExecutionController.mm
+++ b/Source/santad/SNTExecutionController.mm
@@ -633,7 +633,7 @@ static NSString *const kPrinterProxyPostMonterey =
                                        ruleCleanup:SNTRuleCleanupNone
                                             errors:&errors];
   if (errors.count > 0 || !success) {
-    LOGW(@"%@ enocuntered while adding a rule in standalone mode for: %@:",
+    LOGW(@"%@ encountered while adding a rule in standalone mode for: %@:",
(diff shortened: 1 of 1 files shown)


Source/santad/SNTExecutionController.mm
```diff
@@ -628,10 +628,16 @@ - (void)createRuleForStandaloneModeEvent:(SNTStoredExecutionEvent *)se {
...
-  [self.ruleTable addExecutionRules:@[ newRule ] ruleCleanup:SNTRuleCleanupNone error:&err];
-  if (err) {
-    LOGE(@"Failed to add rule in standalone mode for %@: %@", se.filePath,
-         err.localizedDescription);
+  NSArray<NSError *> *errors;
+  BOOL success = [self.ruleTable addExecutionRules:@[ newRule ]
+                                       ruleCleanup:SNTRuleCleanupNone
+                                            errors:&errors];
+  if (errors.count > 0 || !success) {
+    LOGW(@"%@ enocuntered while adding a rule in standalone mode for: %@:",
```
russellhancox:
```suggestion
    LOGW(@"%@ encountered while adding a rule in standalone mode for: %@:",
```

russellhancox (approved):

PR #702, opened 2025-12-17
Kill command impl

mlw:
This implements the functionality to match and kill processes based on a kill command request.

Part of WS-487

### DIFF
--- a/Source/common/SNTKillCommandTest.mm
+++ b/Source/common/SNTKillCommandTest.mm
@@ -48,7 +48,9 @@
 
 - (void)testSNTKillRequestRunningProcessValidInit {
   NSString *uuid = [[NSUUID UUID] UUIDString];
-  NSString *bootUUID = [[NSUUID UUID] UUIDString];
+  NSString *bootUUID = @"2470862D-9913-4B95-A2BB-556EDC163069";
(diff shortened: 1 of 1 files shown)


Source/santad/KillingMachine.mm
```diff
@@ -0,0 +1,353 @@
...
+  pid_t targetPidversion = audit_token_to_pidversion(*token);
+
+  if (targetPid == myPid || targetPid == 1) {
+    LOGW(@"Rejecting request to kill disallowed process");
+    return [[SNTKilledProcess alloc] initWithPid:targetPid
+                                      pidversion:targetPidversion
+                                           error:SNTKilledProcessErrorInvalidTarget];
+  }
+
+  int error = proc_signal_with_audittoken(token, SIGKILL);
```
pmarkowsky, line 199: Do we also need to check for things in the default mute set?

mlw: I went back and forth on whether or not we should prevent killing certain processes. I went the route of doing nothing here except for not allowing the killing of Santa itself (though even that is arguably something we should allow...). There is a check for initproc too, but that is a bit superfluous since the kernel will prevent killing that anyways... Open to discussion though if there are things that we should absolutely not allow!

PR #704, opened 2025-12-19
Block network mounts pt1

mlw:
This is part 1. Can block network mounts when configured and supports an allow list of hosts.

Still todo is UI and stored events.

Part of SNT-50

### DIFF
--- a/Source/common/SNTConfigurator.mm
+++ b/Source/common/SNTConfigurator.mm
@@ -1489,7 +1489,7 @@ static SNTConfigurator *sharedConfigurator = nil;
 
   for (NSString *host in hosts) {
     if (![host isKindOfClass:[NSString class]]) {
-      LOGE(@"Unexpected type in AllowedNetworkMountHosts: %@", [host class]);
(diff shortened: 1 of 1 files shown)


Source/common/SNTConfigurator.mm
```diff
@@ -1443,6 +1461,42 @@ - (BOOL)blockUSBMount {
...
+
+- (NSArray<NSString *> *)allowedNetworkMountHosts {
+  NSArray<NSString *> *hosts = self.syncState[kAllowedNetworkMountHosts];
+  if (!hosts) {
+    return nil;
+  }
+
+  for (NSString *host in hosts) {
+    if (![host isKindOfClass:[NSString class]]) {
+      LOGE(@"Unexpected type in AllowedNetworkMountHosts: %@", [host class]);
```
russellhancox:
```suggestion
      LOGE(@"Unexpected type in %@: %@", kAllowedNetworkMountHosts, [host class]);
```

russellhancox (approved):

PR #707, opened 2025-12-23
santad: Add REQUIRE_TOUCHID option to CELv2

russellhancox:


Source/santad/SNTExecutionController.mm
```diff
@@ -339,8 +339,7 @@ - (void)validateExecEvent:(const Message &)esMsg postAction:(bool (^)(SNTAction)
   bool holdAndAsk = false;
```
mlw: nit: Now that this is stored in the SNTCachedDecision, let's ditch this local variable that is now duplicative.

Source/santad/EventProviders/AuthResultCache.mm
```diff
@@ -128,12 +128,10 @@ static inline uint64_t TimestampFromCachedValue(uint64_t cachedValue) {
...
     // SNTActionHoldAllowed and SNTActionHoldDenied are used for transitions, however the
     // cached action is translated to SNTActionRespondAllow or SNTActionRespondDeny respectively.
+    // We do not want to cache this result and later execs need to go through this path again.
     case SNTActionHoldAllowed: OS_FALLTHROUGH;
-    case SNTActionHoldDenied:
-      return cache->set(vnode_id,
-                        CacheableAction(decision == SNTActionHoldAllowed ? SNTActionRespondAllow
-                                                                         : SNTActionRespondDeny),
-                        CacheableAction(SNTActionRespondHold, 0));
+    case SNTActionHoldDenied: cache->remove(vnode_id); return YES;
```
mlw, line 133: To confirm my understanding - previously, when this result was only used during Standalone Mode, we would be cache responses. But now since we want to ensure CEL TouchID expressions will always get authorized, we simply remove the placeholder from the cache. For users in standalone mode without a CEL TouchID rule, this will result in Santa performing one extra rule lookup on the traditional path before getting into the cache.

Source/common/cel/Evaluator.mm
```diff
@@ -150,27 +154,36 @@
   // Everything else is an error.
   if (bool value; result->GetValue(&value)) {
     if (value) {
-      return {{pbv1::ReturnValue::ALLOWLIST, activation.IsResultCacheable()}};
+      return {{Traits::ALLOWLIST, activation.IsResultCacheable()}};
+    }
+    return {{Traits::BLOCKLIST, activation.IsResultCacheable()}};
+  } else if (int64_t value; result->GetValue(&value)) {
```
mlw, line 160: The logic here has changed slightly - want to make sure it is intentional. Previously, the `ReturnValue_IsValid` check would happen within the `if` conditional check, if that failed, we would fall thru to the next `else if` case.

Now though, the "is valid" check happens within the if-statement, such that if it isn't valid, the next `else if` case isn't evaluated and we'll end up falling through to the error condition.

Is this desired? Should we instead make another function wrapper in CELProtoTraits for `ReturnValue_IsValid` similar to `ReturnValue_descriptor` or `ExecutionContext_descriptor`?

Source/santad/EventProviders/AuthResultCache.mm
russellhancox: Correct

Source/santad/SNTExecutionController.mm
russellhancox: Done

Source/common/cel/Evaluator.mm
russellhancox: The logic hasn't _really_ changed because it isn't possible for the return value to be both an int64 and a CelError, so the `else if` condition would never have evaluated before this change anyway.

mlw (approved):

PR #709, opened 2025-12-29
Remove stored events from the backoff cache when event upload fails

mlw:
This change will remove items just added to the backoff cache if the immediate upload failed for some reason. If upload was successful, the event will remain in the backoff cache for the normal amount of time.

### DIFF
--- a/Source/santad/SNTSyncdQueueTest.mm
+++ b/Source/santad/SNTSyncdQueueTest.mm
@@ -139,10 +139,25 @@
                 });
   OCMVerify(times(1), [mockProxy postEventsToSyncServer:[OCMArg any] reply:[OCMArg any]]);
 
-  // Now simulate the first upload failing, which should remove the backoff
+  // Simulate the first upload failing, which should remove the backoff
   replyBlock(NO);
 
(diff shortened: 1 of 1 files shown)


Source/santad/SNTSyncdQueueTest.mm
```diff
@@ -104,4 +109,44 @@ - (void)testAddEvents {
...
+  OCMVerify(times(1), [mockProxy postEventsToSyncServer:[OCMArg any] reply:[OCMArg any]]);
+
+  // Now simulate the first upload failing, which should remove the backoff
+  replyBlock(NO);
+
+  // Third attempt: Since backoff was removed, event should be dispatched again
+  [sut addStoredEvent:se];
+  dispatch_sync(sut.syncdQueue, ^{
+                });
+  OCMVerify(times(2), [mockProxy postEventsToSyncServer:[OCMArg any] reply:[OCMArg any]]);
```
pmarkowsky, line 149: We should probably add a fourth step that ensures that the events are cached upon a successful upload.

pmarkowsky (approved):
LGTM.

PR #712, opened 2025-12-30
Backfill decision cache on startup

mlw:
On startup, Santa will now asynchronously iterate active PIDs and pre-populate the SNTDecisionCache. Previously, there would be missing information about the instigating process if it started prior to Santa. This mainly affected processes that violated FAA rules.

### DIFF
--- a/Source/santad/EntitlementsFilter.h
+++ b/Source/santad/EntitlementsFilter.h
@@ -50,11 +50,11 @@ class EntitlementsFilter {
   void UpdatePrefixFilter(NSArray<NSString *> *filter);
 
  private:
-  void UpdateTeamIDFilterLocked(NSArray<NSString *> *filter);
-  void UpdatePrefixFilterLocked(NSArray<NSString *> *filter);
+  void UpdateTeamIDFilterLocked(NSArray<NSString *> *filter) ABSL_EXCLUSIVE_LOCKS_REQUIRED(lock_);
+  void UpdatePrefixFilterLocked(NSArray<NSString *> *filter) ABSL_EXCLUSIVE_LOCKS_REQUIRED(lock_);
 
-  std::set<std::string> teamid_filter_;
-  santa::PrefixTree<santa::Unit> prefix_filter_;
+  std::set<std::string> teamid_filter_ ABSL_GUARDED_BY(lock_);
+  santa::PrefixTree<santa::Unit> prefix_filter_ ABSL_GUARDED_BY(lock_);
   absl::Mutex lock_;
 };
 


Source/santad/Santad.mm
```diff
@@ -460,9 +461,8 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
                 LOGI(@"EntitlementsTeamIDFilter changed: '%@' -> '%@'. Flushing caches.", oldValue,
                      newValue);
 
-                // Get the value from the configurator since that method ensures proper structure
-                [exec_controller
-                    updateEntitlementsTeamIDFilter:[configurator entitlementsTeamIDFilter]];
+                // Get the value from the configurator since it will ensures proper types
```
russellhancox:
```suggestion
                // Get the value from the configurator since it ensures proper types
```

Source/santad/Santad.mm
```diff
@@ -483,9 +483,8 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
                 LOGI(@"EntitlementsPrefixFilter changed: '%@' -> '%@'. Flushing caches.", oldValue,
                      newValue);
 
-                // Get the value from the configurator since that method ensures proper structure
-                [exec_controller
-                    updateEntitlementsPrefixFilter:[configurator entitlementsPrefixFilter]];
+                // Get the value from the configurator since it will ensures proper types
```
russellhancox:
```suggestion
                // Get the value from the configurator since it ensures proper types
```

Source/santad/EntitlementsFilter.h
```diff
@@ -0,0 +1,63 @@
...
+  void UpdateTeamIDFilter(NSArray<NSString *> *filter);
+  void UpdatePrefixFilter(NSArray<NSString *> *filter);
+
+ private:
+  void UpdateTeamIDFilterLocked(NSArray<NSString *> *filter);
+  void UpdatePrefixFilterLocked(NSArray<NSString *> *filter);
+
+  std::set<std::string> teamid_filter_;
+  santa::PrefixTree<santa::Unit> prefix_filter_;
+  absl::Mutex lock_;
```
russellhancox, line 58: Consider adding the absl lock annotations (`EXCLUSIVE_LOCKS_REQUIRED`, `LOCKS_EXCLUDED` and `GUARDED_BY`). It's a simple class and easy enough right now to grok all the lock usage but it could grow and I think it's generally a good practice

PR #713, opened 2025-12-31
Upload stored network mount events

mlw:


Source/common/SNTStoredNetworkMountEvent.mm
```diff
@@ -64,4 +67,43 @@ - (NSString *)description {
...
 
+- (NSString *)sanitizedMountFromRemovingCredentials {
+  if (!self.mountFromName) {
+    return self.mountFromName;
+  }
+
+  NSURL *url = [NSURL URLWithString:self.mountFromName];
+  if (url && (url.user || url.password)) {
+    NSURLComponents *components = [NSURLComponents componentsWithURL:url
+                                             resolvingAgainstBaseURL:NO];
```
russellhancox: NSURLComponents can init directly from the string, why make an NSURL first?

PR #715, opened 2026-01-01
Santa command HMAC verification

mlw:
Adds HMAC verification, time window verification, and nonce cache to Santa commands to prevent replay attacks.

Part of WS-487


pmarkowsky (approved):
LGTM.

PR #722, opened 2026-01-07
gui: Add menu item

russellhancox:
The item shows the version of Santa and users can trigger a sync from it. When a sync triggered from the menu item succeeds or fails, the icon changes color for 5s to indicate success/failure (as well as a notification appearing).

For Workshop customers that use Temporary Monitor Mode, the menu allows users to enter, refresh and leave temporary monitor mode and the item will show the time remaining.

The menu item can be disabled with the EnableMenuItem config key, it is enabled by default.

The localized strings were generated with AI, improvements are welcome.

### Light mode:
Normal: 
image
Successful sync:
image
Failed sync:
image
TMM Time Remaining:
image
Menu:
image

### Dark mode:
Normal:
image
Successful sync:
image
Failed sync:
image
TMM Time Remaining:
image
Menu:
image

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
@@ -250,6 +250,11 @@ NS_ASSUME_NONNULL_BEGIN
 - (BOOL)shouldOperateOnDisk:(DADiskRef)disk {
   NSDictionary *diskInfo = CFBridgingRelease(DADiskCopyDescription(disk));
 
+  // Handle cases like time machine mounts where a disk info is not present.
+  if (!diskInfo) {
+    return false;
+  }
+
   BOOL isInternal = [diskInfo[(__bridge NSString *)kDADiskDescriptionDeviceInternalKey] boolValue];
   BOOL isRemovable = [diskInfo[(__bridge NSString *)kDADiskDescriptionMediaRemovableKey] boolValue];
   BOOL isEjectable = [diskInfo[(__bridge NSString *)kDADiskDescriptionMediaEjectableKey] boolValue];
@@ -273,8 +278,15 @@ NS_ASSUME_NONNULL_BEGIN
     return false;
   }
 
-  // We are okay with operations for devices that are non-removable as long as
-  // they are NOT a USB device, or an SD Card.
+  // If the device is external (not internal and not virtual), operate on it.
+  // This catches USB4/Thunderbolt/PCI-Express external drives that may not
+  // be flagged as removable or ejectable.
+  if (!isInternal && !isVirtual) {
+    return true;
+  }
+
+  // For remaining devices (which would be internal/virtual SD cards that
+  // passed the first check), only skip if they're also non-removable.
   if (!isRemovable && !isEjectable && !isUSB && !isSecureDigital) {
     return false;
   }
--- a/Source/santad/EventProviders/SNTEndpointSecurityDeviceManagerTest.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityDeviceManagerTest.mm
@@ -341,6 +341,53 @@ class MockAuthResultCache : public AuthResultCache {
   XCTAssertFalse([self.mockDA.insertedDevices allValues][0].wasMounted);
 }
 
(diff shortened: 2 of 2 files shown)


Source/santad/TemporaryMonitorMode.mm
```diff
@@ -225,6 +238,8 @@
 }
 
 bool TemporaryMonitorMode::BeginLocked(uint32_t seconds, bool gen_uuid_on_start) {
+  NSDate *expiry = [NSDate dateWithTimeIntervalSinceNow:seconds];
```
mlw: nit: Could we remove this temporary variable?

docs/src/lib/santaconfig.ts
```diff
@@ -149,6 +149,13 @@ export const SantaConfigKeyGroups: SantaConfigGroups = {
       type: "bool",
       defaultValue: false,
     },
+    {
+      key: "EnableMenuItem",
+      description: `If true, Santa will show an item in the macOS menu bar to allow syncing and management of
+        temporary monitor mode`,
+      type: "bool",
+      defaultVlaue: true,
```
mlw: nit: Add `versionAdded` key

Source/santad/TemporaryMonitorMode.mm
```diff
@@ -152,13 +155,19 @@
   } else {
     [configurator_ setSyncServerModeTransition:mode_transition];
   }
+
+  // Notify the GUI about policy availability
+  BOOL policyAvailable = (mode_transition.type == SNTModeTransitionTypeOnDemand);
+  [[notification_queue_.notifierConnection remoteObjectProxy]
+      temporaryMonitorModePolicyAvailable:policyAvailable];
```
mlw: nit: Could we remove the temporary?

Source/gui/SNTStatusItemManager.mm
```diff
@@ -0,0 +1,408 @@
...
+                            // Tint the icon green to indicate success
+                            [self
+                                setMenuItemImageWithTintColor:[NSColor
+                                                                  colorNamed:@"SyncSuccessColor"]];
+                            [NSTimer
+                                scheduledTimerWithTimeInterval:2.0
+                                                       repeats:NO
+                                                         block:^(NSTimer *_Nonnull timer) {
+                                                           [self setMenuItemImageWithTintColor:nil];
+                                                         }];
```
mlw: nit: This scheduled timer seems to be the same in both the success and failure case. Recommend moving to down below the `if` statement

Source/santad/SNTDaemonControlController.mm
```diff
@@ -657,4 +657,9 @@ - (void)cancelTemporaryMonitorMode:(void (^)(NSError *))reply {
   reply(err);
 }
 
+- (void)checkTemporaryMonitorModePolicyAvailable:(void (^)(BOOL))reply {
+  SNTModeTransition *modeTransition = [[SNTConfigurator configurator] modeTransition];
```
mlw: Within the TemporaryMonitorMode class's `RequestMinutes` method, we have a few checks that determine whether or not TMM should be available:

* Policy existing
* In an expected client mode (lockdown or already TMM)
* SyncBaseURL Domain is pinned

To better reflect what's available, maybe we abstract these checks within the TMM class and use that here instead of only doing a policy check?

russellhancox: I moved the checks from `RequestMinutes` into an `Available` method which is now called from `RequestMinutes` as well as here.

Source/santad/SNTDaemonControlController.mm
```diff
@@ -658,8 +658,7 @@ - (void)cancelTemporaryMonitorMode:(void (^)(NSError *))reply {
 }
 
 - (void)checkTemporaryMonitorModePolicyAvailable:(void (^)(BOOL))reply {
-  SNTModeTransition *modeTransition = [[SNTConfigurator configurator] modeTransition];
-  reply(modeTransition.type == SNTModeTransitionTypeOnDemand);
+  reply(_temporaryMonitorMode->Available(nullptr));
```
mlw: Super nit: `s/nullptr/nil/`

PR #727, opened 2026-01-09
gui: Allow users to show/hide menu item

russellhancox:
Adds a checkbox to the bottom of the About window that lets users decide whether to enable/disable the menu item. 

image

This turns the `EnableMenuItem` config key into a _default_ value, which takes effect when users haven't made a choice.


pmarkowsky (approved):
LGTM.

PR #730, opened 2026-01-13
santad: Add telemetry for TouchID/hold-and-ask execution events

russellhancox:
When holdAndAsk is set, do not emit telemetry during the normal NOTIFY flow. Once the authorization block completes, change the original decision and trigger logging the enriched event.

### DIFF
--- a/Source/common/SNTXPCControlInterface.h
+++ b/Source/common/SNTXPCControlInterface.h
@@ -74,6 +74,11 @@ typedef NS_ENUM(NSInteger, SNTRuleAddSource) {
 ///
 - (void)installSantaApp:(NSString *)appPath reply:(void (^)(BOOL))reply;
 
+///
+/// Network Extension Ops
+///
+- (void)installNetworkExtension:(void (^)(BOOL))reply;
+
 @end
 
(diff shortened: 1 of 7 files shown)


Source/santad/SantadDeps.mm
```diff
@@ -156,6 +145,46 @@
...
+  if (!tree_status.ok()) {
+    LOGE(@"Failed to create process tree: %@", @(tree_status.status().ToString().c_str()));
+    exit(EXIT_FAILURE);
+  }
+  process_tree = *tree_status;
+
+  std::shared_ptr<::Enricher> enricher = std::make_shared<::Enricher>(process_tree);
+
+  LogExecutionBlock logBlock = ^(santa::Message esMsg) {
+    std::unique_ptr<santa::EnrichedMessage> enrichedMsg = enricher->Enrich(santa::Message(esMsg));
```
mlw: This should be a move now
```suggestion
    std::unique_ptr<santa::EnrichedMessage> enrichedMsg = enricher->Enrich(std::move(esMsg));
```

mlw (changes requested):

PR #735, opened 2026-01-15
Cobranding support

### LINKED ISSUE #232: Feature request: custom branding
It would be lovely to have the ability for additional custom branding in Santa - it's a common request at enterprise businesses and otherwise, often from leadership. Branding can also help with end-user trust that notifications are from trusted software deployed by their organization. 

Something like logo support and ability to add an organization name to a line that appears in the notifications (not the notification center ones, the blocking ones) would be sufficient to me personally. Something like 'Managed by <Organization>' - similar to what Tailscale does. I don't mind if the existing / santa-specific branding remains, just adding our own elsewhere would be sufficient. I think some folks might like that but I don't think it's necessary.

On a lesser note, if custom logo support was added, I think it would be nice to be able to point to a static URL or something on the local file system. But as long as there's one option, works for me.

mlw:
This enables admins to configure cobranding to display on Santa UIs. This initial work (placement, size, even availability) is subject to change. This PR also enhances the SNTTestGUI driver app, abstracting some settings to a common layer that can apply to any of the supported views.

#### Company Logo

image

#### Company Logo (Dark Mode)

image

#### Company Name

image

#### Configuration page updates:

image

---

Fixes #232 
Fixes SNT-11

### DIFF
--- a/Source/common/SNTConfigurator.h
+++ b/Source/common/SNTConfigurator.h
@@ -537,7 +537,7 @@
 
 ///
 ///  If set, the logo that will be displayed on Santa UI. The image will be
-///  scaled down appropriately to fit within image bounds.
+///  scaled down appropriately to fit within image bounds (currently 84x28 pixels).
 ///
 ///  Supported URL schemes:
 ///    - file:// for local file paths
@@ -552,7 +552,7 @@
 ///
 ///  If set, the logo that will be displayed on Santa UI when the UI will be
 ///  shown in dark mode. The image will be scaled down appropriately to fit
-///  within image bounds.
+///  within image bounds (currently 84x28 pixels).
 ///
 ///  Supported URL schemes:
 ///    - file:// for local file paths


docs/src/lib/santaconfig.ts
```diff
@@ -256,6 +256,45 @@ sequences will be replaced in the final URL:
...
+    {
+      key: "BrandingCompanyName",
+      description: `The company name to display on Santa UIs. This setting is ignored if \`BrandingCompanyLogo\` is set.`,
+      type: "string",
+      versionAdded: "2026.1",
+    },
+    {
+      key: "BrandingCompanyLogo",
+      description: `A URL referencing a logo image to display on Santa UIs. The image will be scaled down
+        appropriately to fit within image bounds. If set, this overrides \`BrandingCompanyName\`.
```
russellhancox: > within image bounds

It would be helpful to document what the bounds are (e.g. `(w: 78px, h: 26px)`). Same for BrandingCompanyLogoDark and same in SNTConfigurator.h

russellhancox (dismissed):

russellhancox (approved):

PR #738, opened 2026-01-16
santad: Add REQUIRE_TOUCHID_SILENT to CELv2

russellhancox:
Also unify all TouchID/LocalAuthorization checks in ObjC


pmarkowsky (approved):
LGTM.

PR #741, opened 2026-01-22
Content filter and XPC channel with network extension setup/configuration

mlw:
This PR sets up the XPC channel with the network extension - handling registration, protocol version matching, and settings - as well as connecting back to the extension.

This also implements the stubs to allow non-maintainers to continue building Santa without access to the extension repo.

Part of SNT-261

### DIFF
--- a/Source/gui/main.mm
+++ b/Source/gui/main.mm
@@ -38,6 +38,7 @@ int main(int argc, const char *argv[]) {
       delegate = [SNTSystemExtensionDelegate delegateForSantanetdActivation];
     } else if ([args containsObject:@"--unload-network-extension"]) {
       LOGI(@"Requesting Santa Network Extension (Content Filter) deactivation");
+      LOGW(@"WARNING: All network connections will reset when filter activates");
       delegate = [SNTSystemExtensionDelegate delegateForSantanetdDeactivation];
     }
 #endif


Source/gui/main.mm
```diff
@@ -13,89 +13,50 @@
...
+#ifdef DEBUG
+    else if ([args containsObject:@"--load-network-extension"]) {
       LOGI(@"Requesting Santa Network Extension (Content Filter) activation");
       LOGW(@"WARNING: All network connections will reset when filter activates");
-      req = [OSSystemExtensionRequest
-          activationRequestForExtension:[SNTXPCControlInterface santanetdExtensionBundleID]
-                                  queue:q];
+      delegate = [SNTSystemExtensionDelegate delegateForSantanetdActivation];
     } else if ([args containsObject:@"--unload-network-extension"]) {
       LOGI(@"Requesting Santa Network Extension (Content Filter) deactivation");
```
russellhancox, line 40: Won't this also reset all network connections?

Source/santad/SNTNetworkExtensionQueue.h
```diff
@@ -0,0 +1,26 @@
...
+
+#import <Foundation/Foundation.h>
+
+extern NSString *const kSantaNetworkExtensionProtocolVersion;
+
+@interface SNTNetworkExtensionQueue : NSObject
+
+@property(readonly) NSString *connectedProtocolVersion;
+
+- (NSDictionary *)handleRegistrationWithProtocolVersion:(NSString *)protocolVersion
```
russellhancox, line 23: Consider making protocol version an object type instead of a string. It'll be much easier to later check version compatibility if we have a type with major/minor fields than parsing a string for this.

mlw: I'm OK with this - but would like to punt until a future PR when "settings" is more strongly typed as well.

PR #744, opened 2026-01-23
Add more NATS error logging

pmarkowsky:
This PR adds more details error logging for NATS. This is NPS Push service specific.

### DIFF
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -465,7 +465,7 @@ objc_library(
         ":SNTRule",
         ":SNTStrengthify",
         ":SNTSystemInfo",
-        "//Source/common/ne:SNTNetworkExtensionSettings",
+        "//Source/common/ne:SNTSyncNetworkExtensionSettings",
     ],
 )
 
@@ -479,7 +479,7 @@ objc_library(
         ":SNTCommonEnums",
         ":SNTExportConfiguration",
         ":SNTModeTransition",
-        "//Source/common/ne:SNTNetworkExtensionSettings",
+        "//Source/common/ne:SNTSyncNetworkExtensionSettings",
     ],
 )
 
@@ -491,7 +491,7 @@ santa_unit_test(
         ":SNTConfigBundle",
         ":SNTExportConfiguration",
         ":SNTModeTransition",
-        "//Source/common/ne:SNTNetworkExtensionSettings",
+        "//Source/common/ne:SNTSyncNetworkExtensionSettings",
     ],
 )
 
--- a/Source/common/SNTCachedDecision.h
+++ b/Source/common/SNTCachedDecision.h
@@ -57,5 +57,6 @@
 @property BOOL cacheable;
 @property BOOL holdAndAsk;
 @property BOOL silentTouchID;
+@property NSNumber *touchIDCooldownMinutes;  // nil = no caching (prompt every time)
 
 @end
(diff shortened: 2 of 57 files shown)


Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -665,17 +690,33 @@ static void errorHandler(natsConnection *nc, natsSubscription *sub, natsStatus e
...
+
+  LOGE(@"NATS Error on %@: %@ (status: %d) on subscription: %s", self.pushServer ?: @"server",
+       errorDetail, err, subSubject);
 
   // Check for specific error types
   // NATS doesn't expose specific permission violation status, but we can check the error text
-  if (err == NATS_ERR || (lastError && strstr(lastError, "violation")) ||
-      (lastError && strstr(lastError, "Permitted"))) {
-    LOGE(@"NATS: Permission/Subscription violation on subject: %s", subSubject);
+  if (err == NATS_ERR || (statusText && strstr(statusText, "violation")) ||
```
mlw, line 736: Why do we only attempt reconnecting if the connection was closed for these specific errors? This should prolly be in the comment here as well.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -665,17 +690,33 @@ static void errorHandler(natsConnection *nc, natsSubscription *sub, natsStatus e
...
+  const char *statusText = natsStatus_GetText(err);
   const char *subSubject = sub ? natsSubscription_GetSubject(sub) : "unknown";
 
-  LOGE(@"NATS Error: %s (status: %d) on subscription: %s", lastError ? lastError : "unknown error",
-       err, subSubject);
+  // Also get the detailed last error from the connection if available
+  const char *connLastError = NULL;
+  if (nc) {
+    natsConnection_GetLastError(nc, &connLastError);
+  }
```
mlw: It might be useful to abstract this to a function that returns an NSString. This pattern shows up a few times in the PR, and then is followed by several NULL checks before calling string functions. Having an NSString will simplify a lot of the if conditions that include several different checks.

Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -741,15 +798,28 @@ static void reconnectedCallback(natsConnection *nc, void *closure) {
...
+    natsConnection_GetLastError(nc, &lastError);
+  }
+  NSString *errorInfo =
+      (lastError && strlen(lastError) > 0) ? [NSString stringWithFormat:@" - %s", lastError] : @"";
+  LOGI(@"NATS: Connection to %@ closed%@", self.pushServer ?: @"server", errorInfo);
+
   dispatch_async(self.connectionQueue, ^{
+    if (lastError && strlen(lastError) > 0) {
+      self.lastConnectionError = @(lastError);
+    }
```
mlw: This is an unsafe - the buffer returned by `natsConnection_GetLastError` is an internal pointer, but this code here is in an async block. We should either implement the previous suggestion of abstracting the function to return an NSString so that the buffer is copied, or move to using `natsConnection_ReadLastError`

Source/santasyncservice/SNTPushClientNATS.mm
pmarkowsky: The NATS libraries will handle any disconnected states. Closed in the NATS libraries means that the connection is shutdown and their client libraries will not attempt to reconnect. I'll update the comment.

PR #751, opened 2026-01-29
Add check to protect NPS processes from kill commands.

pmarkowsky:
Adds a check to protect NPS processes.


Source/santad/KillingMachine.mm
```diff
@@ -196,6 +196,18 @@ SNTKilledProcessError LibprocSignalErrorToKilledProcessError(int error) {
                                            error:SNTKilledProcessErrorInvalidTarget];
   }
 
+#ifndef DEBUG
+  // Prevent killing NPS processes in non-debug builds
+  static NSString *kNPSTeamID = @"ZMCG7MLDV9";
+  auto teamIDMatcher = BufferMatcher::TeamID(kNPSTeamID);
+  if (teamIDMatcher->Matches(targetPid)) {
```
mlw, line 203: I don't think we should do this. The daemon itself is protected just above, line 192. But the ancillary processes should be bounceable.

mlw (changes requested):

PR #756, opened 2026-02-05
santad: Replace syncservice telemetry export with Sleigh

russellhancox:
Moves from telemetry being exported directly by syncservice to utilizing a new binary built during release builds and included in the main app bundle. A stub binary is left here non-prod builds to use.

Other related changes:
1. The `santactl telemetry` export is no longer debug only but is hidden. Users should not need to use this command.
2. The sync export config is no longer parsed from V1 requests. This was not supported in the first place.

Part of WS-859

### DIFF
--- a/Source/santad/Logs/EndpointSecurity/LoggerTest.mm
+++ b/Source/santad/Logs/EndpointSecurity/LoggerTest.mm
@@ -540,6 +540,7 @@ class MockSleighLauncher : public santa::SleighLauncher {
   MockSleighLauncher *mockSleighPtr = mockSleigh.get();
 
   // No Launch call expected since there are no files
+  EXPECT_CALL(*mockSleighPtr, Launch(testing::_, testing::_)).Times(0);
 
   LoggerPeer l(std::move(mockSleigh), self.exportConfigBlock, TelemetryEvent::kEverything, 5, 1, 3,
                nullptr, mockWriter);


Source/santad/Logs/EndpointSecurity/Logger.h
```diff
@@ -48,22 +49,22 @@ class Logger : public Timer<Logger> {
...
     kUnknown = 0,
-    kUncompressedStream,
-    kGzipStream,
     kZstdStream,
   };
 
   static std::unique_ptr<Logger> Create(
-      std::shared_ptr<santa::EndpointSecurityAPI> esapi, SNTSyncdQueue *syncd_queue,
+      std::shared_ptr<santa::EndpointSecurityAPI> esapi,
+      std::shared_ptr<santa::SleighLauncher> sleigh_launcher,
```
tburgin: It doesn't look like the passed in SleighLauncher is used anywhere else but here. `std::unique_ptr` or just an object member variable seems more appropriate.

Source/santad/Logs/EndpointSecurity/Logger.mm
```diff
@@ -256,39 +227,35 @@
...
-      NSFileHandle *handle = [NSFileHandle fileHandleForReadingAtPath:path];
-      if (!handle) {
-        LOGW(@"Failed to get a file handle for telemetry file to export: %@", path);
-        tracker_.AckCompleted(*file_to_export);
-        continue;
-      }
-
       struct stat sb;
-      if (fstat(handle.fileDescriptor, &sb) != 0) {
+      if (stat(path.UTF8String, &sb) != 0) {
```
tburgin:
```suggestion
      if (stat(path.fileSystemRepresentation, &sb) != 0) {
```

Source/santad/Logs/EndpointSecurity/Logger.mm
```diff
@@ -300,90 +267,40 @@
         continue;
       }
 
-      ExportLogType log_type = GetLogType(handle, path);
-      if (log_type == ExportLogType::kUnknown) {
```
tburgin, line 304: If the log type is `ExportLogType::kUnknown` shouldn't we still skip?

Source/santad/SleighLauncher.h
```diff
@@ -0,0 +1,64 @@
...
+  static std::unique_ptr<SleighLauncher> Create(NSString *sleigh_path, uint32_t timeout_seconds);
+
+  SleighLauncher(NSString *sleigh_path, uint32_t timeout_seconds);
+
+  virtual ~SleighLauncher() = default;
+
+  SleighLauncher(SleighLauncher &&) = default;
+  SleighLauncher &operator=(SleighLauncher &&rhs) = default;
+  SleighLauncher(SleighLauncher &) = delete;
+  SleighLauncher &operator=(SleighLauncher &rhs) = delete;
```
tburgin: If you make SleighLauncher a simple member object on its caller you can delete the move constructors. The helper Create could also be deleted.

When deleting the copy operations, the move operations are implicitly disabled.

```suggestion
  SleighLauncher(SleighLauncher &) = delete;
  SleighLauncher &operator=(SleighLauncher &rhs) = delete;
```

Source/santad/SleighLauncher.h
```diff
@@ -0,0 +1,64 @@
...
+
+  virtual SleighResult Launch(const std::vector<std::string> &input_files);
+
+  void SetTimeoutSeconds(uint32_t timeout_seconds);
+
+ private:
+  NSString *sleigh_path_;
+  std::unique_ptr<std::atomic_uint32_t> timeout_seconds_;
```
tburgin: Could this just be an arg to Launch()?

Source/santad/Logs/EndpointSecurity/Logger.mm
russellhancox: We don't know the log type anymore, we leave that up to Sleigh to figure out

Source/santad/SleighLauncher.h
russellhancox: Good idea, that's much better

Source/santad/SleighLauncher.mm
```diff
@@ -0,0 +1,149 @@
...
+
+  // Check Sleigh binary exists and is correctly signed.
+  MOLCodesignChecker *csc = [[MOLCodesignChecker alloc] initWithBinaryPath:sleigh_path_];
+  if (!csc || ![csc.teamID isEqualToString:@"ZMCG7MLDV9"] ||
+      csc.signatureFlags & kSecCodeSignatureAdhoc) {
+    LOGD(@"SleighLauncher::Launch(): Sleigh code signature is invalid");
+#ifndef DEBUG
+    return absl::FailedPreconditionError("Sleigh code signature is invalid");
+#endif
+  }
```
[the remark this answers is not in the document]
russellhancox: No.

PR #760, opened 2026-02-06
Fix issues with about dialog re-opening

mlw:
Addresses a timing race in the UI code that could trigger the About dialog to reopen when switching virtual desktops.

Also addresses an issue that made the About dialog appear to "flicker" when it opened.

### DIFF
--- a/Source/gui/SNTAppDelegate.mm
+++ b/Source/gui/SNTAppDelegate.mm
@@ -81,8 +81,9 @@
   __block BOOL hasVisibleWindows = NO;
   [NSApp enumerateWindowsWithOptions:0
                           usingBlock:^(NSWindow *_Nonnull window, BOOL *_Nonnull stop) {
-                            if ([window isKindOfClass:NSClassFromString(@"NSStatusBarWindow")])
+                            if ([window isKindOfClass:NSClassFromString(@"NSStatusBarWindow")]) {
                               return;
+                            }
                             *stop = hasVisibleWindows = window.visible;
                           }];
   if (hasVisibleWindows) {


Source/gui/SNTAppDelegate.mm
```diff
@@ -75,7 +75,19 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
...
 - (void)applicationDidBecomeActive:(NSNotification *)notification {
-  NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
+  // Only show the dock icon if the app has visible windows (excluding status bar windows).
+  // Without this guard, events such as virtual desktop switches can reactivate the app and
+  // cause the dock icon to reappear when there are no windows to display.
+  __block BOOL hasVisibleWindows = NO;
+  [NSApp enumerateWindowsWithOptions:0
+                          usingBlock:^(NSWindow *_Nonnull window, BOOL *_Nonnull stop) {
+                            if ([window isKindOfClass:NSClassFromString(@"NSStatusBarWindow")])
+                              return;
```
tburgin:
```suggestion
                            if ([window isKindOfClass:NSClassFromString(@"NSStatusBarWindow")]) {
                              return;
                            }
```

tburgin:
```suggestion
                            if ([window isKindOfClass:NSClassFromString(@"NSStatusBarWindow")]) {
                              return;
                            }
```

Source/gui/SNTAppDelegate.mm
```diff
@@ -75,7 +75,20 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
...
+  [NSApp enumerateWindowsWithOptions:0
+                          usingBlock:^(NSWindow *_Nonnull window, BOOL *_Nonnull stop) {
+                            if ([window isKindOfClass:NSClassFromString(@"NSStatusBarWindow")]) {
+                              return;
+                            }
+                            *stop = hasVisibleWindows = window.visible;
+                          }];
+  if (hasVisibleWindows) {
+    NSApp.activationPolicy = NSApplicationActivationPolicyRegular;
+  }
```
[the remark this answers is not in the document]
mlw: This is not valid

PR #762, opened 2026-02-09
Support CLONE events in the compiler controller for transitive rules

mlw:
This adds tracking for CLONE events for "compiler" binaries that will generate transitive rules. This will help improve coverage for additional toolchains, such as rust/cargo.

### DIFF
--- a/Source/santad/SNTCompilerController.mm
+++ b/Source/santad/SNTCompilerController.mm
@@ -188,6 +188,8 @@ static constexpr std::string_view kIgnoredCompilerProcessPathPrefix = "/dev/";
         [ruleTable executionRuleForIdentifiers:(struct RuleIdentifiers){
                                                    .binarySHA256 = targetFile.SHA256,
                                                }];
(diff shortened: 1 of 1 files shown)


Source/santad/SNTCompilerController.mm
```diff
@@ -128,6 +128,30 @@ - (BOOL)handleEvent:(const Message &)esMsg withLogger:(std::shared_ptr<Logger>)l
...
       }
 
+      break;
+    case ES_EVENT_TYPE_NOTIFY_CLONE:
+      if (![self isCompiler:esMsg->process->audit_token]) {
+        return NO;
+      }
+
+      // Note: For CLONE events, we first attempt to process the `source`. But if that doesn't exist
+      // exist (it normally will since clones are copies), we fall back to the `destination`.
```
russellhancox:
```suggestion
      // Note: For CLONE events, we first attempt to process the `source`. But if that doesn't
      // exist (it normally will since clones are copies), we fall back to the `destination`.
```

russellhancox (dismissed):

PR #763, opened 2026-02-09
santad: Pass files to Sleigh as open FDs

russellhancox:


Source/santad/SleighLauncher.mm
```diff
@@ -34,90 +41,152 @@
...
+  // Write serialized config to stdin pipe
+  const char *data = serialized->data();
+  size_t remaining = serialized->size();
+  bool write_failed = false;
+  while (remaining > 0) {
+    ssize_t written = write(stdin_pipe[1], data, remaining);
+    if (written < 0) {
+      write_failed = true;
+      break;
+    }
```
mlw, line 141: Would be nice to handle `EINTR` and retry
```suggestion
    if (written < 0) {
      if (errno == EINTR) {
        // Retry
        continue;
      }
      write_failed = true;
      break;
    }
```

Source/santad/SleighLauncher.mm
```diff
@@ -34,90 +41,152 @@
...
+    data += written;
+    remaining -= written;
+  }
+  close(stdin_pipe[1]);
+
+  if (write_failed) {
+    kill(pid, SIGTERM);
```
mlw: I'm a bit curious about the use of `SIGTERM` vs `SIGKILL` - is Sleigh handling `SIGTERM`? (couldn't find it with a cursory search...). If there are no plans to handle `SIGTERM`, should maybe just use `SIGKILL`. If we do plan to handle `SIGTERM`, then we should do the dance where we forcibly kill via `SIGKILL` after some time if the process still hasn't exited.

Source/santad/SleighLauncher.mm
```diff
@@ -34,90 +41,152 @@
...
+  // Parent process
+  close(stdin_pipe[0]);           // Close read end
+  std::move(close_fds).Invoke();  // Close input FDs (child inherited copies)
+
+  // Write serialized config to stdin pipe
+  const char *data = serialized->data();
+  size_t remaining = serialized->size();
+  bool write_failed = false;
+  while (remaining > 0) {
+    ssize_t written = write(stdin_pipe[1], data, remaining);
```
mlw, line 133: There is a problem here if the child process exits/crashes - writing to the pipe will trigger a SIGPIPE, which will cause santad to terminate. We should probably set `F_SETNOSIGPIPE`. This issue previously existed, but we overlooked it.

Source/santad/SleighLauncher.mm
```diff
@@ -34,90 +41,152 @@
...
+    waitpid(pid, &child_status, 0);
+    dispatch_semaphore_signal(sema);
+  });
+
   if (dispatch_semaphore_wait(sema,
                               dispatch_time(DISPATCH_TIME_NOW, timeout_secs * NSEC_PER_SEC))) {
-    // Timeout - kill the process
-    [task terminate];
+    // Timeout - kill the process and wait for async waitpid to complete
+    kill(pid, SIGTERM);
```
mlw: Same comment as above re: SIGTERM/SIGKILL

Source/santad/SleighLauncher.mm
russellhancox: Done

Source/santad/SleighLauncher.mm
russellhancox: Done

Source/santad/SleighLauncher.mm
russellhancox: Done

Source/santad/SleighLauncher.mm
russellhancox: Done

mlw (approved):

PR #765, opened 2026-02-09
Add NetworkActivity to protobuf schema

mlw:
Add messages to the santa.proto schema for tracking network flow activity. Telemetry is grouped per-process, per-window in order to help deduplicate process data when several flows are generated.

### DIFF
--- a/Source/common/santa.proto
+++ b/Source/common/santa.proto
@@ -1173,7 +1173,7 @@ message NetworkActivity {
   // A single network socket flow observed during the monitoring window
   message Flow {
     // Unique identifier for this flow
-    optional bytes id = 1;
+    optional string id = 1;
 
     // Remote endpoint address (IPv4 or IPv6 string representation)
     optional string remote_address = 2;


Source/common/santa.proto
```diff
@@ -1168,6 +1168,61 @@ message XProtect {
   }
 }
 
+// Per-process network activity collected during a single monitoring window
+message NetworkActivity {
+  // A single network socket flow observed during the monitoring window
+  message Flow {
+    // Remote endpoint address (IPv4 or IPv6 string representation)
+    optional string remote_address = 1;
```
russellhancox: Add a flow UUID and make that field 1, this way we can track flows and have a way to tie flows together, particular start & end times

Source/common/santa.proto
```diff
@@ -1168,6 +1168,61 @@ message XProtect {
...
+      DIRECTION_OUTBOUND = 2;
+    }
+    optional Direction direction = 7;
+
+    // Cumulative byte counts for this flow
+    optional uint64 bytes_inbound = 8;
+    optional uint64 bytes_outbound = 9;
+
+    // When the flow was first observed
+    // Not present for flows that were re-tracked after idle eviction
```
russellhancox: Does that imply that it's only on the first event for a flow?

mlw: I originally intended to set this on all flow updates. But thinking more, I'm not sure that's very useful. Updated comment.

Source/common/santa.proto
mlw: As a heads up - we talked about various id encoding schemes. I ended up keeping this as a UUID for the main telemetry. We can use the rewriter to transform to whatever makes sense. If we decide down the line we'd rather have this in Santa proper we can change it without much hassle in the near term

Source/common/santa.proto
```diff
@@ -1172,42 +1172,47 @@ message XProtect {
 message NetworkActivity {
   // A single network socket flow observed during the monitoring window
   message Flow {
+    // Unique identifier for this flow
+    optional bytes id = 1;
```
russellhancox: Other UUIDs in the proto are strings. We can handle this either way and bytes _might_ be a little bit smaller in the produced file, just wanted to mention it for consistency.

mlw: Moved to string

PR #766, opened 2026-02-10
Restore jitter for push notifications sent to tags

pmarkowsky:
This PR adds up to 3 minutes of jitter for a push notification sent to a tag.

Fix for SNT-273.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -18,7 +18,7 @@ bazel_dep(name = "boringssl", version = "0.20251110.0")
 bazel_dep(name = "protos", version = "1.0.1", repo_name = "northpolesec_protos")
 git_override(
     module_name = "protos",
-    commit = "d7a5b41f4477c6c1ff366d825cb2b62421b64fd7",
+    commit = "f095446ff47302a2aa48e0efa19859517281a361",
     remote = "https://github.com/northpolesec/protos",
 )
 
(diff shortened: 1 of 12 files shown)


Source/santasyncservice/SNTPushClientNATS.mm
```diff
@@ -594,6 +594,36 @@ - (void)subscribe {
...
+        });
+      } else {
+        LOGI(@"NATS: Triggering immediate sync due to message on %@", subject);
+
+        dispatch_async(dispatch_get_main_queue(), ^{
+          if (!self.isShuttingDown) {
+            [self.syncDelegate sync];
+          }
+        });
+      }
```
mlw, line 611: nit - I think it would be easier to read if this was simplified a bit (will require fixing the strict checks in the test):
```suggestion
      uint32_t jitterSeconds = 0;
      if ([subject hasPrefix:@"santa.tag."]) {
        jitterSeconds = arc4random_uniform(181);
        LOGI(@"NATS: Scheduling sync in %u seconds (jitter) due to tag message on %@",
             jitterSeconds, subject);
      } else {
        LOGI(@"NATS: Triggering immediate sync due to message on %@", subject);
      }

      dispatch_async(dispatch_get_main_queue(), ^{
        if (!self.isShuttingDown) {
          [self.syncDelegate syncSecondsFromNow:jitterSeconds];
        }
      });
```

mlw: Something happened to the commit history and the old code is back?

PR #771, opened 2026-02-11
Support File Access variants for EventDetailURL and EventDetailText

mlw:
Addresses issue where File Access URL used in block messages was inappropriately using the `EventDetailURL`/`EventDetailText` configuration keys. However, FAA blocks and execution blocks support different replacement keys.

This PR adds support for two new config keys that will act as a fallback when no custom url is provided in the policy dictionary (either at the global level or per-rule).
* `FileAccessEventDetailURL`
* `FileAccessEventDetailText`

Updated docs screenshots:
FAA Config:
image

Config Keys:
image

### DIFF
--- a/Source/common/SNTBlockMessage.mm
+++ b/Source/common/SNTBlockMessage.mm
@@ -283,9 +283,9 @@ static id ValueOrNull(id value) {
 
 + (NSURL *)eventDetailURLForFileAccessEvent:(SNTStoredFileAccessEvent *)event
                                   customURL:(NSString *)url {
-  return [self
-      eventDetailURLForEvent:event
-                   customURL:(url ?: [[SNTConfigurator configurator] fileAccessEventDetailURL])
+  return
+      [self eventDetailURLForEvent:event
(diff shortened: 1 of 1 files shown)


Source/common/SNTBlockMessage.mm
```diff
@@ -284,8 +285,8 @@ + (NSURL *)eventDetailURLForFileAccessEvent:(SNTStoredFileAccessEvent *)event
                                   customURL:(NSString *)url {
   return [self
       eventDetailURLForEvent:event
-                   customURL:(url ?: [[SNTConfigurator configurator] eventDetailURL])templateMapping
-                            :[self fileAccessEventDetailTemplateMappingForEvent:event]];
+                   customURL:(url ?: [[SNTConfigurator configurator] fileAccessEventDetailURL])
+                   templateMapping:[self fileAccessEventDetailTemplateMappingForEvent:event]];
```
russellhancox, line 289: Why aren't the colons aligned?

mlw: ugh, because clang format doesn't always understand parentheses. Updated.

PR #772, opened 2026-02-11
Add EventUpload santa command handler

pmarkowsky:
This implements a NATS santa command handler for EventUploadRequest that performs the equivalent of `santactl eventupload <path>` — creating file/bundle events for a given path (with glob support) and uploading them to the sync server.

This is done by doing the following:

- Add generateEventsFromPath:enableBundles:reply: XPC method to the bundle service, which expands globs, creates SNTFileInfo/event objects, and optionally performs full bundle hashing
- Add EventUpload handler to NATS command dispatch with 300s timeout. (We should probably change to at least 600s)
- Add eventUploadForPath:reply:  method to SNTSyncManager that checks if enableBundles is set and calls the bundle service, and uploads resulting events via postEventsToSyncServer:reply:
- Update protos dependency for EventUploadRequest/EventUploadResponse
- Adds tests

### DIFF
--- a/.github/workflows/check-markdown.yml
+++ b/.github/workflows/check-markdown.yml
@@ -13,7 +13,7 @@ jobs:
       - name: "Checkout Santa"
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
       - name: "Check for deadlinks"
-        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # ratchet:lycheeverse/lychee-action@v2
+        uses: lycheeverse/lychee-action@2b973e86fc7b1f6b36a93795fe2c9c6ae1118621 # ratchet:lycheeverse/lychee-action@v1
         with:
           fail: true
       - name: "Check for trailing whitespace and newlines"
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,7 +17,7 @@ bazel_dep(name = "boringssl", version = "0.20260211.0")
 bazel_dep(name = "protos", version = "1.0.1", repo_name = "northpolesec_protos")
 git_override(
     module_name = "protos",
-    commit = "53b9f4e4db95e927f095754c24c12028bae03abc",
+    commit = "fa0ad3f19cd1e32ec4946331e94fb1e13e9df4ea",
     remote = "https://github.com/northpolesec/protos",
 )
 
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -26,7 +26,6 @@ cc_proto_library(
     deps = [":santa_proto"],
 )
 
-
 objc_library(
     name = "PowerMonitor",
     srcs = ["PowerMonitor.mm"],
--- a/Source/common/cel/Activation.h
+++ b/Source/common/cel/Activation.h
@@ -49,12 +49,14 @@ class Activation : public ::google::api::expr::runtime::BaseActivation {
 
   Activation(std::unique_ptr<ExecutableFileT> file, std::vector<std::string> (^args)(),
              std::map<std::string, std::string> (^envs)(), uid_t (^euid)(), std::string (^cwd)(),
-             std::vector<AncestorT> (^ancestors)(), std::vector<FileDescriptorT> (^fds)())
+             std::string (^path)(), std::vector<AncestorT> (^ancestors)(),
+             std::vector<FileDescriptorT> (^fds)())
       : file_(std::move(file)),
         args_(args),
         envs_(envs),
         euid_(euid),
         cwd_(cwd),
+        path_(path),
         ancestors_(ancestors),
         fds_(fds) {};
   ~Activation() = default;
@@ -80,6 +82,7 @@ class Activation : public ::google::api::expr::runtime::BaseActivation {
   Memoizer<std::map<std::string, std::string>> envs_;
   Memoizer<uid_t> euid_;
   Memoizer<std::string> cwd_;
+  Memoizer<std::string> path_;
   Memoizer<std::vector<AncestorT>> ancestors_;
   Memoizer<std::vector<FileDescriptorT>> fds_;
 
--- a/Source/common/cel/Activation.mm
+++ b/Source/common/cel/Activation.mm
@@ -158,6 +158,8 @@ std::optional<cel_runtime::CelValue> Activation<IsV2>::FindValue(
     return CELValue(euid_(), arena);
   } else if (name == "cwd") {
     return CELValue(cwd_(), arena);
+  } else if (name == "path") {
+    return CELValue(path_(), arena);
   }
 
   // Handle the V2 specific fields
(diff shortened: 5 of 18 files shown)


Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -37,6 +37,9 @@
 // Semi-arbitrary number of seconds to wait for santad to finish killing processes
 static constexpr int64_t kKillResponseTimeoutSeconds = 90;
 
+// Bundle hashing + event upload can take a while for large bundles
+static constexpr int64_t kEventUploadResponseTimeoutSeconds = 300;
```
pmarkowsky: Should we make this 600s?

Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -291,6 +298,46 @@ - (BOOL)checkAndRecordNonce:(NSString *)uuid {
...
+    LOGE(@"NATS: EventUploadRequest timed out for path: %@", path);
+    return nil;
+  }
+
+  if (uploadError) {
+    LOGE(@"NATS: EventUploadRequest failed for path %@: %@", path, uploadError);
+    return nil;
+  }
+
+  return google::protobuf::Arena::Create<::pbv1::EventUploadResponse>(arena);
```
mlw: The `EventUploadResponse` proto should be updated with an enum that has values for the various ways this could fail.

Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -343,6 +390,19 @@ - (BOOL)checkAndRecordNonce:(NSString *)uuid {
...
 
+    case ::pbv1::SantaCommandRequest::kEventUpload: {
+      LOGI(@"NATS: Dispatching EventUploadRequest command");
+      auto *eventUploadResponse = [self handleEventUploadRequest:command.event_upload()
+                                                 withCommandUUID:uuid
+                                                         onArena:arena];
+      if (eventUploadResponse) {
+        response->unsafe_arena_set_allocated_event_upload(eventUploadResponse);
+      } else {
+        response->set_error(::pbv1::SantaCommandResponse::ERROR_UNSPECIFIED);
```
mlw: The top level error is meant to refer something about the mechanics of the command (e.g. bad UUID, unknown request type, etc.). Commands are meant to have command-specific errors in order to accurately describe what went wrong.

mlw (changes requested):

pmarkowsky:
@coderabbit full review

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -361,6 +363,72 @@ - (MOLXPCConnection *)daemonConnection {
...
+    __block NSArray<SNTStoredExecutionEvent *> *resultEvents;
+    [[bs remoteObjectProxy] generateEventsFromPath:path
+                                     enableBundles:enableBundles
+                                             reply:^(NSArray<SNTStoredExecutionEvent *> *events) {
+                                               resultEvents = events;
+                                               dispatch_semaphore_signal(eventSema);
+                                             }];
+
+    dispatch_semaphore_wait(eventSema, DISPATCH_TIME_FOREVER);
+
```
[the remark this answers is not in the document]
pmarkowsky: @coderabbitai MOLXPCConnection's remoteObjectProxy already uses remoteObjectProxyWithErrorHandler: internally, but its error handler only invalidates the connection — it  doesn't signal our semaphore. And if the connection is down, the proxy is nil and the reply block never fires.

The correct action is to check for a nil proxy before sending the message, and handle the failure path.

pmarkowsky:
@mlw PTAL when you have a chance.

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -361,6 +363,88 @@ - (MOLXPCConnection *)daemonConnection {
...
+    __block NSArray<SNTStoredExecutionEvent *> *resultEvents;
+    [[bs remoteObjectProxy] generateEventsFromPath:path
+                                     enableBundles:enableBundles
+                                             reply:^(NSArray<SNTStoredExecutionEvent *> *events) {
+                                               resultEvents = events;
+                                               dispatch_semaphore_signal(eventSema);
+                                             }];
+
+    dispatch_semaphore_wait(eventSema, DISPATCH_TIME_FOREVER);
```
mlw: We should probably bound this to some timeout - running on the `syncQueue` will block all syncs while this runs. If something bad happens, we don't want to block forever.

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -361,6 +363,88 @@ - (MOLXPCConnection *)daemonConnection {
...
+    return;
+  }
+
+  dispatch_async(self.syncQueue, ^{
+    __block BOOL replied = NO;
+    void (^guardedReply)(NSError *) = ^(NSError *e) {
+      if (replied) return;
+      replied = YES;
+      if (reply) reply(e);
+    };
```
mlw: Not very likely, but the code is racy. Maybe something like:
```
    auto replied = std::make_shared<std::atomic<bool>>(false);
    void (^guardedReply)(NSError *) = ^(NSError *e) {
      if (!replied->exchange(true)) {
        if (reply) reply(e);
      }
    };
```
Will need to add these above:
```
#include <atomic>
#include <memory>
```

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -361,6 +366,114 @@ - (MOLXPCConnection *)daemonConnection {
...
+  }
+
+  if (dispatch_semaphore_wait(self.syncLimiter, DISPATCH_TIME_NOW)) {
+    if (reply) {
+      reply([NSError errorWithDomain:@"com.northpolesec.santa.syncservice"
+                                code:5
+                            userInfo:@{NSLocalizedDescriptionKey : @"Too many syncs in progress"}]);
+    }
+    return;
+  }
```
mlw: Not necessarily a blocker for now, and I see this was flagged by CR - but not sure if this is appropriate here. If the command triggers successfully, it will be difficult for an operator to then go back and track down why results didn't appear.

If we want rate limiting for commands, we should add that at the NATS command handler layer, and then we could return an appropriate error back in the response so that the sender could be correctly notified.

My suggestion would be to remove this code, then go back in a followup PR to support command rate limiting. What do you think?

Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -291,6 +295,42 @@ - (BOOL)checkAndRecordNonce:(NSString *)uuid {
...
+    return pbResponse;
+  }
+
+  // Fire off the event upload asynchronously - don't wait for completion
+  [strongSyncDelegate
+      eventUploadForPath:path
+                   reply:^(NSError *error) {
+                     if (error) {
+                       LOGE(@"NATS: EventUploadRequest failed for path %@: %@", path, error);
+                       [strongSyncDelegate sync];
```
mlw: What's the goal of the sync here? Bundle hash path takes a while for me to trace, but it doesn't seem like any of the events here are getting put into the event table(?). Maybe related issue, it doesn't look like these events go through `LogBundleHashingEvents` either (e.g., `syncBundleEvent:relatedEvents:` in the SNTDaemonControlController will put things in the event table and call into the logger).

PR #774, opened 2026-02-12
docs: Add shareable link support to CEL playground

russellhancox:
This PR adds a "Copy Link" button next to the "Evaluate" button, which copies a URL to the clipboard that can be shared to return to the page with the expression and context fields filled in.

image


pmarkowsky (approved):
LGTM.

PR #777, opened 2026-02-13
events: add USB Mount decision matrix and remount args to the event

sharvilshah:
SNT-282

### DIFF
--- a/Source/common/SNTStoredUSBMountEvent.h
+++ b/Source/common/SNTStoredUSBMountEvent.h
@@ -49,10 +49,8 @@ typedef NS_ENUM(NSInteger, SNTStoredUSBMountEventDecision) {
 - (instancetype)initWithDeviceModel:(NSString *)deviceModel
                        deviceVendor:(NSString *)deviceVendor
                         mountOnName:(NSString *)mountOnName
-                           protocol:(NSString *)protocol;
-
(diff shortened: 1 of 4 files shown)


Source/common/SNTStoredUSBMountEvent.h
```diff
@@ -31,8 +37,22 @@
...
+
 - (instancetype)initWithDeviceModel:(NSString *)deviceModel
                        deviceVendor:(NSString *)deviceVendor
-                        mountOnName:(NSString *)mountOnName;
+                        mountOnName:(NSString *)mountOnName
+                           protocol:(NSString *)protocol;
+
+- (void)setRemountArgs:(NSArray<NSString *> *)remountArgs;
+
+- (void)setDecision:(SNTStoredUSBMountEventDecision)decision;
```
mlw: The usage looks like we could add these to the initializer (or even make two initializers for one with/without the remount args). We wouldn't need the setters then at all. If I missed something and the setters are needed, we can probably jsut make the properties readwrite to get the compiler-generated setters instead of writing it ourselves.

mlw (dismissed):

Source/common/SNTStoredUSBMountEvent.h
sharvilshah: That makes sense, and is simpler. Updating to have two initializers.

mlw (approved):

PR #780, opened 2026-02-13
santad: Add Ancestors field to CELv2 rules

russellhancox:
Adds an `ancestors` field to CELv2 activations which follows parents of a process starting with the parent and going back to launchd. Passes the process tree to the execution controller and creates it regardless of annotation configurations. When backfilling the process tree, the signing info for processes is collected too. This makes the backfilling a little slower but still sub-10s on a well loaded machine.

Video Below 🎥 

[Process Tree Aware CELv2 Rules](https://www.youtube.com/watch?v=JppG95sINNQ)

### DIFF
--- a/Source/common/String.h
+++ b/Source/common/String.h
@@ -60,9 +60,6 @@ static inline std::string_view StringTokenToStringView(es_string_token_t es_str)
 }
 
 static inline std::string StringTokenToString(es_string_token_t es_str) {
-  if (es_str.data == nullptr || es_str.length == 0) {
-    return {};
-  }
   return std::string(es_str.data, es_str.length);
 }
 


MODULE.bazel
```diff
@@ -18,7 +18,7 @@ bazel_dep(name = "boringssl", version = "0.20251110.0")
 bazel_dep(name = "protos", version = "1.0.1", repo_name = "northpolesec_protos")
 git_override(
     module_name = "protos",
-    commit = "f095446ff47302a2aa48e0efa19859517281a361",
+    commit = "a7b43043101c154639cfda9c1d681b378ef3ca0f",
     remote = "https://github.com/northpolesec/protos",
```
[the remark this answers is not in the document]
russellhancox: No, we don't create new version numbers for the protos repo, updating the commit directly is fine

Source/santad/ProcessTree/process_tree_macos.mm
```diff
@@ -14,24 +14,81 @@
...
 
 namespace {
+
+// csops operations (from XNU bsd/sys/codesign.h)
+static constexpr unsigned int kCsopStatus = 0;
+static constexpr unsigned int kCsopCDHash = 5;
+static constexpr unsigned int kCsopIdentity = 11;
+static constexpr unsigned int kCsopTeamID = 14;
+
+static constexpr size_t kCDHashSize = 20;
```
mlw: nit: Use `CS_CDHASH_LEN` from `#include <Kernel/kern/cs_blobs.h>`

Source/santad/ProcessTree/process_tree_macos.mm
```diff
@@ -14,24 +14,81 @@
...
+  uint32_t len;
+  char data[];
+};
+
+std::optional<CodeSigningInfo> LoadCodeSigningInfoForPID(pid_t pid) {
+  uint32_t flags = 0;
+  if (csops(pid, kCsopStatus, &flags, sizeof(flags)) != 0) {
+    return std::nullopt;
+  }
+  if (!(flags & CS_VALID)) {
```
mlw: We should probably check both `CS_VALID` and `CS_SIGNED` are set - It's not wholly clear to me if even possible for this to not be true, but it isn't extra effort to handle.

Source/santad/ProcessTree/process_tree_macos.mm
```diff
@@ -14,24 +14,81 @@
...
+static constexpr unsigned int kCsopIdentity = 11;
+static constexpr unsigned int kCsopTeamID = 14;
+
+static constexpr size_t kCDHashSize = 20;
+
+struct csops_blob {
+  uint32_t type;
+  uint32_t len;
+  char data[];
+};
```
mlw: We should abstract the extern function decl, the csops operation values, and the csops blob struct to a new file (e.g. CSOpsHelpers maybe fits in with our convention). We can then use this from here and in `KillingMachine.mm`.

Source/santad/ProcessTree/SNTEndpointSecurityAdapter.mm
```diff
@@ -48,14 +49,25 @@ void InformFromESEvent(ProcessTree &tree, const Message &msg) {
...
       }
 
-      es_string_token_t executable = msg->event.exec.target->executable->path;
+      const es_process_t *target = msg->event.exec.target;
+      es_string_token_t executable = target->executable->path;
+
+      // Extract code signing info from the target process
+      CodeSigningInfo cs_info{
+          .signing_id = std::string(santa::StringTokenToStringView(target->signing_id)),
+          .team_id = std::string(santa::StringTokenToStringView(target->team_id)),
```
mlw: Would be nice to add a new function to `String.h` like `StringTokenToString`.

Source/santad/ProcessTree/process_tree_macos.mm
```diff
@@ -14,24 +14,81 @@
...
+    return std::nullopt;
+  }
+  if (!(flags & CS_VALID)) {
+    return std::nullopt;
+  }
+
+  CodeSigningInfo info;
+  info.is_platform_binary = (flags & CS_PLATFORM_BINARY) != 0;
+
+  // Get CDHash (raw 20-byte buffer, no blob wrapper)
```
mlw: We should confirm the lengths of everything returned contains expected byte counts. Easy enough for CDHash/TeamID, for SigningID we can just check some min value (1?). Note: The length is network byte order. See `KillingMachine.mm`.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -67,6 +66,7 @@ @interface SNTPolicyProcessor () {
   std::unique_ptr<santa::cel::Evaluator<false>> celEvaluatorV1_;
   std::unique_ptr<santa::cel::Evaluator<true>> celEvaluatorV2_;
   std::shared_ptr<santa::EntitlementsFilter> entitlementsFilter_;
+  std::shared_ptr<santa::santad::process_tree::ProcessTree> processTree_;
```
mlw: Is this still needed? I'm not seeing how this is getting used.

Source/santad/ProcessTree/process_tree_macos.mm
```diff
@@ -14,24 +14,81 @@
...
+  uint8_t cdhash[kCDHashSize] = {};
+  if (csops(pid, kCsopCDHash, cdhash, sizeof(cdhash)) == 0) {
+    info.cdhash = santa::BufToHexString(cdhash, sizeof(cdhash));
+  }
+
+  // Get signing identity (blob-wrapped, null-terminated string)
```
mlw: For ad hoc signed binaries, we probably shouldn't grab signing ID info.

Source/santad/SNTExecutionController.mm
```diff
@@ -697,4 +709,56 @@ - (void)flushTouchIDApprovalCache {
...
+  for (const auto &p : processTree->RootSlice(*proc)) {
+    if (!p->program_) {
+      continue;
+    }
+
+    AncestorT ancestor;
+    ancestor.set_path(p->program_->executable);
+
+    if (p->program_->code_signing) {
+      const auto &cs = *p->program_->code_signing;
```
mlw, line 743: Duplicating previous comment here - we should not be using SIDs for ad hoc signed code.

Source/santad/SNTPolicyProcessor.mm
russellhancox: Nope, leftover from an earlier implementation that I missed cleaning back up

Source/santad/SNTExecutionController.mm
russellhancox: With the previous comment resolved and the same check in InformFromESEvent, the process tree will never have TeamID or SigningID populated for adhoc binaries.

pmarkowsky: This would match our execution rule behavior where we ignore them unless validly signed. 

It doesn't seem unreasonable since adhoc signed stuff can put what ever they want in those fields.

Alternatively we'd need to expose the signing flags, so you could add the check if you cared or not. Which I think we could do in a future PR.

Source/common/String.h
```diff
@@ -59,6 +59,10 @@ static inline std::string_view StringTokenToStringView(es_string_token_t es_str)
   return std::string_view(es_str.data, es_str.length);
 }
 
+static inline std::string StringTokenToString(es_string_token_t es_str) {
+  return std::string(es_str.data, es_str.length);
+}
```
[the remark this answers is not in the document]
mlw: This is not true in C++20 https://cplusplus.github.io/LWG/issue3111

Source/santad/SNTExecutionController.mm
```diff
@@ -698,4 +710,56 @@ - (void)flushTouchIDApprovalCache {
...
+  if (!processTree) return {};
+
+  using Traits = santa::cel::CELProtoTraits<true>;
+  using AncestorT = typename Traits::AncestorT;
+
+  auto pid = santa::santad::process_tree::PidFromAuditToken(esMsg->process->parent_audit_token);
+  auto proc = processTree->Get(pid);
+  if (!proc) {
+    return {};
+  }
```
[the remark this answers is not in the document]
russellhancox: No, you're wrong. The parent audit token is the same in both process and exec, this is not skipping a parent.

PR #784, opened 2026-02-16
Don't trigger FAA violations from the Santa bundle service

mlw:
Adds a probe callback to the Authorizer client looking for executions of the bundle service. When it fires, that process is muted from the Data FAA client so that it won't trigger FAA violations as part of its scanning.

Currently, the allowlisting only applies to the bundle service. Other components, namely `santactl` will still trigger a violation.

### DIFF
--- a/Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
+++ b/Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
@@ -36,6 +36,7 @@ using santa::Message;
 @property SNTCompilerController *compilerController;
 @property SNTExecutionController *execController;
 @property NSMutableArray<id<SNTEndpointSecurityProbe>> *probes;
+@property BOOL enabled;
 @end
 
 @implementation SNTEndpointSecurityAuthorizer {
@@ -57,6 +58,8 @@ using santa::Message;
     _compilerController = compilerController;
     _authResultCache = authResultCache;
     _ttyWriter = std::move(ttyWriter);
+
+    _probes = [NSMutableArray array];
 
     [self establishClientOrDie];
   }
@@ -239,6 +242,7 @@ using santa::Message;
 }
 
 - (void)enable {
+  self.enabled = YES;
   [super subscribeAndClearCache:{
                                     ES_EVENT_TYPE_AUTH_EXEC,
                                     ES_EVENT_TYPE_AUTH_PROC_SUSPEND_RESUME,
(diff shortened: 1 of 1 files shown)


Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
```diff
@@ -245,8 +245,11 @@ - (void)enable {
   }];
 }
 
-- (void)registerAuthExecProbe:(id<SNTEndpointSecurityProbe>)watcher {
-  self.procWatcherProbe = watcher;
+- (void)registerAuthExecProbe:(id<SNTEndpointSecurityProbe>)probe {
+  if (!self.probes) {
+    self.probes = [NSMutableArray array];
+  }
+  [self.probes addObject:probe];
```
russellhancox: Can this method be called on multiple threads?

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
```diff
@@ -245,8 +245,11 @@ - (void)enable {
   }];
 }
 
-- (void)registerAuthExecProbe:(id<SNTEndpointSecurityProbe>)watcher {
-  self.procWatcherProbe = watcher;
+- (void)registerAuthExecProbe:(id<SNTEndpointSecurityProbe>)probe {
+  if (!self.probes) {
+    self.probes = [NSMutableArray array];
+  }
```
russellhancox: Why not create this in init?

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
mlw: It isn't today, but suppose nothing prevents it...

Added a `@synchronized` guard. Additionally, turned it into a programming error to try and register additional probes after the client is enabled.

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.mm
```diff
@@ -239,14 +242,23 @@ - (bool)postAction:(SNTAction)action forMessage:(const Message &)esMsg {
...
+  if (self.enabled) {
+    // This is a programming error. Bail.
+    [NSException raise:@"Probe registration error"
+                format:@"Cannot register probes after authorizer is enabled"];
+  }
+
+  @synchronized(self) {
+    [self.probes addObject:probe];
+  }
 }
```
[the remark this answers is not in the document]
mlw: This is acceptable.

PR #790, opened 2026-02-18
Normalize language around removable media

sharvilshah:
SNT-263


Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -297,8 +297,8 @@ - (void)runWithArguments:(NSArray *)arguments {
         @"watchdog_ram_events" : @(ramEvents),
         @"watchdog_cpu_peak" : @(cpuPeak),
         @"watchdog_ram_peak" : @(ramPeak),
-        @"block_usb" : @(blockUSBMount),
-        @"remount_usb_mode" : (blockUSBMount && remountUSBMode.count ? remountUSBMode : @""),
+        @"block_mountable_removable_media" : @(blockUSBMount),
+        @"remount_removable_media_mode" : (blockUSBMount && remountUSBMode.count ? remountUSBMode : @""),
```
mlw: Let's keep the old keys for now, and add new ones that are just duplicate data. We try to keep the JSON more stable compared to the table output since we know some folks have tooling that parses this.

We can call it out in the release notes that it will be removed by release X.

mlw (dismissed):

russellhancox (approved):

PR #795, opened 2026-02-19
Add ThinLTO to Builds

pmarkowsky:
Part of SNT-308

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -18,9 +18,9 @@ build --per_file_copt=.*\.mm\$@-std=c++20
 build --cxxopt=-std=c++20
 build --host_cxxopt=-std=c++20
 
-# Use ThinLTO for faster link times with LTO benefits
-build --copt=-flto=thin
-build --linkopt=-flto=thin
+# Use ThinLTO for faster link times than full LTO with 90%+ of the LTO benefits
+build:opt --copt=-flto=thin
+build:opt --linkopt=-flto=thin
 
 # Enable C++ interoperability in Swift
 build --@rules_swift//swift:copt=-cxx-interoperability-mode=default


.bazelrc
```diff
@@ -18,6 +18,10 @@ build --per_file_copt=.*\.mm\$@-std=c++20
 build --cxxopt=-std=c++20
 build --host_cxxopt=-std=c++20
 
+# Use ThinLTO for faster link times with LTO benefits
+build --copt=-flto=thin
+build --linkopt=-flto=thin
```
russellhancox:
```suggestion
build --features=thin_lto
```

russellhancox: I haven't tested this, but it _should_ work

pmarkowsky: I wasn't sure if apple_rules supported the `--feature` based on https://github.com/bazelbuild/bazel/issues/16269

pmarkowsky: Doesn't look like it works, sadly.

russellhancox: Damn, 3+ years and still not resolved 😢

.bazelrc
```diff
@@ -18,6 +18,10 @@ build --per_file_copt=.*\.mm\$@-std=c++20
 build --cxxopt=-std=c++20
 build --host_cxxopt=-std=c++20
 
+# Use ThinLTO for faster link times with LTO benefits
```
russellhancox: I feel like this comment could be misinterpreted - it's faster than Full Fat LTO but still slower than no-LTO at all.

.bazelrc
russellhancox: This change LGTM but we may want to limit these flags to `opt` builds

PR #796, opened 2026-02-19
deps: Update NATS dep to use local_defines

russellhancox:
Without this, any target that has a dependency on NATS will have these same defines. They should be attached to the nats lib only


pmarkowsky (approved):
LGTM.

PR #798, opened 2026-02-19
docs: force-update ajv dependency

mlw:


russellhancox:
I tried this [yesterday](https://github.com/northpolesec/santa/pull/794), Docusaurus isn't compatible with the new version

PR #804, opened 2026-02-23
[WIP] telemetry: record oflags for `FileAccess` events

sharvilshah:
Still testing and updating


Source/common/santa.proto
```diff
@@ -558,6 +558,15 @@ message FileAccess {
   // This can happen, for example, when a single operation violates both Data
   // and Process File Access Authorization rules.
   optional string operation_id = 7;
+
+  // The access mode of the file in question
+  enum AccessMode {
```
grimley517, line 563: have you considered using a flagged enum for this?

eg
[flags]
enum
{
CanRead= 0b0001
CanWrite= 0b0010
}

This gives all 4 access modes in a similar manner to unix.  and prevents people having to check for read and write separately.

grimley517 (commented):
access mode - suggest using flags to identify capabilities rarther than the logical modes. the enum is MECE, but this makes what you do with the mode more complicated.

pmarkowsky:
Lets close this for now and pick it up after the next release.

PR #810, opened 2026-02-24
Install/upgrade network extension on system wake

mlw:
This adds support for automatic install/upgrade of the network extension when the system wakes up.

Part of SNT-261

### DIFF
--- a/Source/santad/SNTDaemonControlController.h
+++ b/Source/santad/SNTDaemonControlController.h
@@ -42,7 +42,7 @@
                           checkCacheBlock:(SNTAction (^)(SantaVnode))checkCacheBlock;
 
 /// Install the network extension, optionally checking whether an upgrade is needed first.
-/// When force is YES, delegates directly to installNetworkExtension:.
+/// When force is YES, delegates to installNetworkExtension: as long as installation is authorized.
 /// When force is NO, skips install if the loaded version already matches the on-disk version.
 - (void)installNetworkExtensionForce:(BOOL)force reply:(void (^)(BOOL))reply;
 


Source/santad/SNTNetworkExtensionQueue.h
```diff
@@ -53,4 +53,10 @@ extern NSString *const kSantaNetworkExtensionProtocolVersion;
 /// Returns YES if the network extension is currently connected.
 - (BOOL)isLoaded;
 
+/// Determines whether the network extension needs to be upgraded.
+/// Compares the loaded extension's CFBundleVersion against the on-disk version.
+/// Returns YES if the extension is not connected, versions differ, or loaded version is unknown.
+/// Returns NO if versions match, or if the on-disk bundle is unreadable.
+- (BOOL)networkExtensionNeedsUpgrade;
```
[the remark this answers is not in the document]
mlw: Covered by "or loaded version is unknown".

Source/santad/Santad.mm
```diff
@@ -803,17 +804,30 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
...
   }
 
-  // Trigger network extension install/upgrade if this is the first launch after boot
+  void (^installNetworkExtension)(NSString *reason) = ^(NSString *reason) {
+    LOGD(@"%@, triggering network extension install/upgrade if authorized", reason);
+    [dc installNetworkExtensionForce:NO
+                               reply:^(BOOL success) {
+                                 if (!success) {
+                                   LOGD(@"Network extension install skipped "
+                                        @"(not enabled, not authorized, or version matches)");
```
russellhancox, line 813: Not necessary today but it would be nice in the future if we could distinguish these 3 failure reasons

mlw: Agreed, right now you should be able to see the matching log message. But would be a bit easier if we only printed from one location, returned an NSError or something.

PR #824, opened 2026-02-27
device: enforce drive encryption

sharvilshah:
SNT-216 and SNT-344

### DIFF
--- a/Source/santad/BUILD
+++ b/Source/santad/BUILD
@@ -642,6 +642,9 @@ objc_library(
     name = "SNTEndpointSecurityDeviceManager",
     srcs = ["EventProviders/SNTEndpointSecurityDeviceManager.mm"],
     hdrs = ["EventProviders/SNTEndpointSecurityDeviceManager.h"],
+    sdk_frameworks = [
+        "DiskArbitration",
+    ],
     deps = [
         ":AuthResultCache",
         ":EndpointSecurityAPI",
@@ -785,6 +788,9 @@ objc_library(
     name = "EndpointSecuritySerializerBasicString",
     srcs = ["Logs/EndpointSecurity/Serializers/BasicString.mm"],
     hdrs = ["Logs/EndpointSecurity/Serializers/BasicString.h"],
+    sdk_frameworks = [
+        "DiskArbitration",
+    ],
     deps = [
         ":EndpointSecurityAPI",
         ":EndpointSecuritySanitizableString",
@@ -805,6 +811,9 @@ objc_library(
     name = "EndpointSecuritySerializerProtobuf",
     srcs = ["Logs/EndpointSecurity/Serializers/Protobuf.mm"],
     hdrs = ["Logs/EndpointSecurity/Serializers/Protobuf.h"],
+    sdk_frameworks = [
+        "DiskArbitration",
+    ],
     deps = [
         ":EndpointSecurityAPI",
         ":EndpointSecuritySerializer",
@@ -1347,9 +1356,6 @@ santa_unit_test(
         "bsm",
         "EndpointSecurity",
     ],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     structured_resources = [
         "//Source/santad/testdata:binaryrules_testdata",
     ],
@@ -1423,9 +1429,6 @@ santa_unit_test(
         "bsm",
         "EndpointSecurity",
     ],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     deps = [
         ":EndpointSecurityEnrichedTypes",
         ":EndpointSecurityEnricher",
@@ -1448,9 +1451,6 @@ santa_unit_test(
 santa_unit_test(
     name = "EndpointSecuritySerializerProtobufTest",
     srcs = ["Logs/EndpointSecurity/Serializers/ProtobufTest.mm"],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     structured_resources = [
         "//Source/santad/testdata:protobuf_json_testdata",
     ],
@@ -1546,9 +1546,6 @@ santa_unit_test(
     sdk_dylibs = [
         "bsm",
         "EndpointSecurity",
-    ],
-    sdk_frameworks = [
-        "DiskArbitration",
     ],
     deps = [
         ":EndpointSecurityLogger",
@@ -1707,9 +1704,6 @@ santa_unit_test(
     sdk_dylibs = [
         "EndpointSecurity",
     ],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     deps = [
         ":AuthResultCache",
         ":EndpointSecurityClient",
@@ -1733,9 +1727,6 @@ santa_unit_test(
     srcs = ["EventProviders/SNTEndpointSecurityDataFileAccessAuthorizerTest.mm"],
     sdk_dylibs = [
         "EndpointSecurity",
-    ],
-    sdk_frameworks = [
-        "DiskArbitration",
     ],
     deps = [
         ":MockEndpointSecurityAPI",
@@ -1767,9 +1758,6 @@ santa_unit_test(
 santa_unit_test(
     name = "FAAPolicyProcessorTest",
     srcs = ["EventProviders/FAAPolicyProcessorTest.mm"],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     deps = [
         ":EndpointSecurityMessage",
         ":FAAPolicyProcessor",
@@ -1792,9 +1780,6 @@ santa_unit_test(
     sdk_dylibs = [
         "EndpointSecurity",
     ],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     deps = [
         ":EndpointSecurityClient",
         ":EndpointSecurityMessage",
@@ -1815,9 +1800,6 @@ santa_unit_test(
     srcs = ["EventProviders/SNTEndpointSecurityRecorderTest.mm"],
     sdk_dylibs = [
         "EndpointSecurity",
-    ],
-    sdk_frameworks = [
-        "DiskArbitration",
     ],
     deps = [
         ":AuthResultCache",
@@ -1850,9 +1832,6 @@ santa_unit_test(
         "bsm",
         "EndpointSecurity",
     ],
-    sdk_frameworks = [
-        "DiskArbitration",
-    ],
     deps = [
         ":AuthResultCache",
         ":DiskArbitrationTestLib",
@@ -1878,9 +1857,6 @@ santa_unit_test(
     srcs = ["SNTCompilerControllerTest.mm"],
     sdk_dylibs = [
         "EndpointSecurity",
-    ],
-    sdk_frameworks = [
-        "DiskArbitration",
     ],
     deps = [
         ":EndpointSecurityLogger",


Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -409,73 +418,75 @@ - (void)runWithArguments:(NSArray *)arguments {
...
-    printf("  %-35s | %lld  (Peak: %.2fMB)\n", "Watchdog RAM Events", ramEvents, ramPeak);
+    printf("  %-40s | %lld\n", "Static Rules", staticRuleCount);
+    printf("  %-40s | %lld  (Peak: %.2f%%)\n", "Watchdog CPU Events", cpuEvents, cpuPeak);
+    printf("  %-40s | %lld  (Peak: %.2fMB)\n", "Watchdog RAM Events", ramEvents, ramPeak);
 
     printf(">>> Cache Info\n");
-    printf("  %-35s | %lld\n", "Root cache count", rootCacheCount);
-    printf("  %-35s | %lld\n", "Non-root cache count", nonRootCacheCount);
+    printf("  %-40s | %lld\n", "Root cache count", rootCacheCount);
+    printf("  %-40s | %lld\n", "Non-root cache count", nonRootCacheCount);
```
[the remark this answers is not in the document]
sharvilshah: This is very minor and a non-issue

Source/common/santa.proto
```diff
@@ -485,6 +485,10 @@ message Disk {
 
   // Path mounted from
   optional string mount_from = 11;
+
+  // Whether the disk media is encrypted (from DiskArbitration kDADiskDescriptionMediaEncryptedKey).
+  // Absent if the encryption status was not yet determined by the system.
+  optional bool encrypted = 12;
```
russellhancox, line 491: Don't forget to copy this to the telemetry proto in the protos repo

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
```diff
@@ -525,9 +546,31 @@ - (es_auth_result_t)handleAuthDeviceMount:(const Message &)m
     return ES_AUTH_RESULT_ALLOW;
   }
 
+  // Read the encryption status from DiskArbitration. This key is populated
+  // asynchronously by diskarbitrationd via _FSGetMediaEncryptionStatusAtPath().
+  // A nil value means the key has not yet been populated — we treat this as
+  // "not encrypted" (deny-by-default) since this is a security-critical decision.
+  NSNumber *encryptedValue = diskInfo[(__bridge NSString *)kDADiskDescriptionMediaEncryptedKey];
+  BOOL isEncrypted = (encryptedValue != nil && [encryptedValue boolValue]);
```
russellhancox, line 559: You don't technically need the nil check here, if encryptedValue is nil the result of calling boolValue on it will be false anyway. It's fine to leave it if you want to be explicit.

Source/common/santa.proto
sharvilshah: Yep, thanks, good catch! Updated: https://github.com/northpolesec/protos/pull/100

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
sharvilshah: Ah, good to know, thanks

Source/santad/BUILD
```diff
@@ -1422,6 +1422,9 @@ santa_unit_test(
         "bsm",
         "EndpointSecurity",
     ],
+    sdk_frameworks = [
+        "DiskArbitration",
+    ],
```
mlw: Nit: Are these new `sdk_frameworks` args really needed? I'd expect them only on the targets that are directly using the DA framework.

Source/santad/Santad.mm
```diff
@@ -119,6 +119,10 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
                                                remountUSBMode:[configurator remountUSBMode]
                                            startupPreferences:[configurator onStartUSBOptions]];
 
+  // maybe have a new device event callback?
+  device_client.blockUnencryptedRemovableMediaMount =
+      [configurator blockUnencryptedRemovableMediaMount];
```
mlw: nit: we should move this into the initializer (e.g. similar to other configurator settings). This will also be important when addressing the "perform startup tasks" portion which is called from the initializer and will need the proper setting in place.

Source/santasyncservice/SNTSyncConfigBundle.mm
```diff
@@ -27,6 +27,7 @@ @interface SNTConfigBundle (ConfigBundleCreator)
 @property NSString *allowlistRegex;
 @property NSString *blocklistRegex;
 @property NSNumber *blockUSBMount;
+@property NSNumber *blockUnencryptedRemovableMediaMount;
```
mlw, line 30: nit: Also add this to `Source/santasyncservice/SNTSyncConfigBundleTest.mm`

Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -230,6 +230,8 @@ BOOL Preflight(SNTSyncPreflight *self, google::protobuf::Arena *arena,
     }
   }
 
+  // TODO(sbs): Umm...handle block_unencrypted_removable_media_mount, need to update protos
```
mlw, line 233: I assume this is for a followup PR after the sync protos are updated?

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
mlw, line 1: Fine to do in a followup PR - but we will need to handle `performStartupTasks:`, as well as think about any additional startup prefs that might need to be defined.

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
```diff
@@ -592,6 +637,102 @@ - (void)remount:(DADiskRef)disk
...
+  [self.remountingDisks addObject:bsdName];
+
+  NSArray<NSString *> *args = self.remountArgs;
+  CFStringRef *argv = (CFStringRef *)calloc(args.count + 1, sizeof(CFStringRef));
+  CFArrayGetValues((__bridge CFArrayRef)args, CFRangeMake(0, (CFIndex)args.count),
+                   (const void **)argv);
+
+  DADiskMountWithArguments(disk, NULL, kDADiskMountOptionDefault, DiskRemountCompletionCallback,
+                           (__bridge void *)self, argv);
+  free(argv);
```
mlw: This code is very similar to `remount:mountMode:semaphore:` - could we reuse that method?

mlw (changes requested):

Source/santad/BUILD
sharvilshah: Yep, need these for tests, otherwise they fail

Source/santasyncservice/SNTSyncPreflight.mm
sharvilshah: Yep, it's WIP, doing it in tandem with the WS1092

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
sharvilshah: Gotcha, thanks. I want to dive a bit deeper to grok `performStartupTasks` will ping you

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
sharvilshah: Yeah, those three lines are very similar to our existing `remount` method, I had already given that a quick shot, but found that having separate paths for "remount" encrypted disks was a bit easier to reason about and test. I can try and give it another shot

Source/santasyncservice/SNTSyncConfigBundle.mm
sharvilshah: Good catch, thanks! Updated

Source/santad/Santad.mm
sharvilshah: Thanks, updated

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
sharvilshah: Updated to generalize and reuse `remount`

Source/santad/EventProviders/SNTEndpointSecurityDeviceManagerTest.mm
```diff
@@ -791,4 +813,313 @@ - (void)testNetworkMountFailOpenWithInvalidURL {
...
+- (void)testMountApproval_EncryptedDevice_ReturnsDissenter {
+  SNTEndpointSecurityDeviceManager *dm =
+      [self createDeviceManagerForApprovalTestsWithBlockUnencryptedRemovableMediaMount:YES];
+  dm.remountArgs = @[ @"rdonly", @"noexec" ];
+
+  MockDADisk *mockDisk = [self createMockDiskWithEncrypted:YES];
+
+  DADissenterRef result = [dm handleMountApproval:(__bridge DADiskRef)mockDisk];
+  XCTAssertTrue(result != NULL, @"Should return a dissenter for encrypted device with remountArgs");
+
```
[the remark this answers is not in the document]
sharvilshah: This is just the test, so it's okay

Source/santad/EventProviders/SNTEndpointSecurityDeviceManager.mm
```diff
@@ -580,18 +630,111 @@ - (es_auth_result_t)handleAuthDeviceMount:(const Message &)m
...
+  DADiskMountWithArguments(disk, NULL, kDADiskMountOptionDefault, callback, context,
+                           (CFStringRef *)argv);
 
   free(argv);
 }
 
+- (DADissenterRef __nullable)handleMountApproval:(DADiskRef)disk {
+  const char *bsdNameStr = DADiskGetBSDName(disk);
+  NSString *bsdName = bsdNameStr ? @(bsdNameStr) : @"";
+
```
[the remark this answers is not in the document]
sharvilshah: Returning NULL means we approve the mount, so in this case it's okay to use `""` as a sentinel key

Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -417,73 +426,75 @@ - (void)runWithArguments:(NSArray *)arguments {
...
   } else {
     printf(">>> Daemon Info\n");
-    printf("  %-35s | %s\n", "Mode", [clientMode UTF8String]);
-    printf("  %-35s | %s\n", "Log Type", [eventLogType UTF8String]);
-    printf("  %-35s | %s\n", "File Logging", (fileLogging ? "Yes" : "No"));
-    printf("  %-35s | %s\n", "Removable Media Blocking", (blockUSBMount ? "Yes" : "No"));
-    if (blockUSBMount && remountUSBMode.count > 0) {
-      printf("  %-35s | %s\n", "Removable Media Remounting Mode",
+    printf("  %-40s | %s\n", "Mode", [clientMode UTF8String]);
+    printf("  %-40s | %s\n", "Log Type", [eventLogType UTF8String]);
```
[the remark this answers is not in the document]
sharvilshah: This is not an issue, this has been in the codebase for a while

PR #831, opened 2026-03-04
Fix unbounded arena growth in CEL Evaluator

pmarkowsky:
The Evaluator holds a single `google::protobuf::Arena` that was shared across every `CompileAndEvaluate` call.  Due to  Arena being a monotonic bump allocator that never frees individual allocations, this caused us to leak compilation temporaries and materialized variable values (e.g. args, envs etc. ) on  each evaluation.

*  `CompileAndEvaluate()` now creates a stack-local Arena that is destroyed at end of scope, freeing all temporaries.
*   Updated helpers to take an arena as an argument and pass in the stack-local arena.

Added tests to ensure this actually fixes the problem and does not regress.

Before these changes my tests showed 600MB+ of leak after 5000 iterations after these changes it's constant.

### DIFF
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -1111,6 +1111,7 @@ test_suite(
         ":ScopedFileTest",
         ":ScopedIOObjectRefTest",
         ":TelemetryEventMapTest",
+        "//Source/common/cel:ArenaGrowthTest",
         "//Source/common/cel:CELTest",
         "//Source/common/faa:unit_tests",
         "//Source/common/ne:unit_tests",


Source/common/cel/BUILD
```diff
@@ -84,3 +84,13 @@ santa_unit_test(
         "@northpolesec_protos//celv2:v2_cc_proto",
     ],
 )
+
+santa_unit_test(
+    name = "ArenaGrowthTest",
```
russellhancox, line 89: These needs referencing in the test_suite in Source/common/BUILD so it gets run regularly

PR #838, opened 2026-03-05
Add fallback CEL expressions

russellhancox:
Fallback expressions can be synced, restricted to SyncV2. They are not accessible from MDM config (to avoid merging issues). Only 10 expressions are allowed.

### DIFF
--- a/Source/common/SNTConfigurator.mm
+++ b/Source/common/SNTConfigurator.mm
@@ -1886,33 +1886,11 @@ static SNTConfigurator *sharedConfigurator = nil;
 }
 
 - (NSArray<SNTCELFallbackRule *> *)celFallbackRules {
-  NSData *data = self.syncState[kCELFallbackRulesKey];
-  if (![data isKindOfClass:[NSData class]]) {
-    return nil;
-  }
-  NSError *error;
-  NSSet *classes = [NSSet setWithObjects:[NSArray class], [SNTCELFallbackRule class], nil];
-  NSArray *rules = [NSKeyedUnarchiver unarchivedObjectOfClasses:classes fromData:data error:&error];
-  if (error) {
-    LOGE(@"Failed to deserialize CEL fallback rules: %@", error);
-    return nil;
-  }
-  return rules;
+  return [SNTCELFallbackRule deserializeArray:self.syncState[kCELFallbackRulesKey]];
 }
 
 - (void)setSyncServerCELFallbackRules:(NSArray<SNTCELFallbackRule *> *)rules {
-  NSData *data = nil;
-  if (rules) {
-    NSError *error;
-    data = [NSKeyedArchiver archivedDataWithRootObject:rules
-                                 requiringSecureCoding:YES
-                                                 error:&error];
-    if (error) {
-      LOGE(@"Failed to serialize CEL fallback rules: %@", error);
-      return;
-    }
-  }
-  [self updateSyncStateForKey:kCELFallbackRulesKey value:data];
+  [self updateSyncStateForKey:kCELFallbackRulesKey value:[SNTCELFallbackRule serializeArray:rules]];
 }
 
 - (void)migrateDeprecatedStatsStatePath:(NSString *)oldPath {
--- a/Source/santad/SNTPolicyProcessor.mm
+++ b/Source/santad/SNTPolicyProcessor.mm
@@ -514,12 +514,13 @@ static void UpdateCachedDecisionSigningInfo(
   }
 
   NSDictionary *entitlements = csInfo.entitlements;
+  cd.rawEntitlements = [entitlements sntDeepCopy];
 
   if (entitlementsFilterCallback) {
     cd.entitlements = entitlementsFilterCallback(entitlements);
     cd.entitlementsFiltered = (cd.entitlements.count != entitlements.count);
   } else {
-    cd.entitlements = [entitlements sntDeepCopy];
+    cd.entitlements = cd.rawEntitlements;
     cd.entitlementsFiltered = NO;
   }
 
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -643,6 +643,7 @@ objc_library(
     ],
     deps = [
         ":CoderMacros",
+        ":SNTLogging",
     ],
 )
 
--- a/Source/common/SNTCELFallbackRule.h
+++ b/Source/common/SNTCELFallbackRule.h
@@ -33,4 +33,7 @@
                       customMsg:(NSString *)customMsg
                       customURL:(NSString *)customURL;
 
++ (NSData *)serializeArray:(NSArray<SNTCELFallbackRule *> *)rules;
++ (NSArray<SNTCELFallbackRule *> *)deserializeArray:(NSData *)data;
+
 @end
--- a/Source/common/SNTCELFallbackRule.mm
+++ b/Source/common/SNTCELFallbackRule.mm
@@ -15,6 +15,7 @@
 #import "Source/common/SNTCELFallbackRule.h"
 
 #import "Source/common/CoderMacros.h"
+#import "Source/common/SNTLogging.h"
 
 @interface SNTCELFallbackRule ()
 @property(readwrite, copy) NSString *celExpr;
@@ -54,6 +55,35 @@
     DECODE(decoder, customURL, NSString);
   }
   return self;
+}
+
++ (NSData *)serializeArray:(NSArray<SNTCELFallbackRule *> *)rules {
+  if (!rules) {
+    return nil;
+  }
+  NSError *error;
+  NSData *data = [NSKeyedArchiver archivedDataWithRootObject:rules
+                                       requiringSecureCoding:YES
+                                                       error:&error];
+  if (error) {
+    LOGE(@"CEL fallback rules serialization failed: %@", error.localizedDescription);
+    return nil;
+  }
+  return data;
+}
+
++ (NSArray<SNTCELFallbackRule *> *)deserializeArray:(NSData *)data {
+  if (!data) {
+    return nil;
+  }
+  NSError *error;
+  NSSet *classes = [NSSet setWithObjects:[NSArray class], [SNTCELFallbackRule class], nil];
+  NSArray *rules = [NSKeyedUnarchiver unarchivedObjectOfClasses:classes fromData:data error:&error];
+  if (error) {
+    LOGE(@"CEL fallback rules deserialization failed: %@", error.localizedDescription);
+    return nil;
+  }
+  return rules;
 }
 
 @end
--- a/Source/common/SNTCachedDecision.h
+++ b/Source/common/SNTCachedDecision.h
@@ -50,6 +50,7 @@
 @property NSString *rawSigningID;
 @property NSString *cdhash;
 @property NSDictionary *entitlements;
+@property NSDictionary *rawEntitlements;
 @property BOOL entitlementsFiltered;
 @property BOOL platformBinary;
 @property uint32_t codesigningFlags;
--- a/Source/common/SNTCachedDecision.mm
+++ b/Source/common/SNTCachedDecision.mm
@@ -46,6 +46,7 @@
     _certCommonName = previous.certCommonName;
     _certChain = previous.certChain;
     _entitlements = previous.entitlements;
+    _rawEntitlements = previous.rawEntitlements;
     _entitlementsFiltered = previous.entitlementsFiltered;
     _secureSigningTime = previous.secureSigningTime;
     _signingTime = previous.signingTime;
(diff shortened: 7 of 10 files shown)


Source/common/SNTConfigurator.mm
```diff
@@ -1878,6 +1884,14 @@ - (void)setSyncServerTelemetryFilterExpressions:(NSArray<NSString *> *)expressio
                         value:EnsureArrayOfStrings(expressions)];
 }
 
+- (NSArray *)celFallbackExpressions {
+  return EnsureArrayOfStrings(self.syncState[kCELFallbackExpressionsKey]);
+}
+
+- (void)setSyncServerCELFallbackExpressions:(NSArray<NSString *> *)expressions {
+  [self updateSyncStateForKey:kCELFallbackExpressionsKey value:EnsureArrayOfStrings(expressions)];
+}
```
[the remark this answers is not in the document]
russellhancox: No, enforcing this as the expressions come in is the simpler option and repeating the restriction in 2 places means we have to update 2 places later if we adjust the limit. This is good as-is.

Source/santad/CELActivation.mm
```diff
@@ -114,6 +114,31 @@ ActivationCallbackBlock CreateCELActivationBlock(
...
+                @try {
+                  jsonData = [NSJSONSerialization dataWithJSONObject:value
+                                                             options:NSJSONWritingFragmentsAllowed
+                                                               error:&err];
+                } @catch (NSException *) {
+                }
+                if (!jsonData) {
+                  // Skip entitlements that can't be serialized to JSON.
+                  return;
+                }
```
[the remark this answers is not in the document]
russellhancox: Objective-C objects are nil initialized by the compiler.

Source/santad/Logs/EndpointSecurity/Serializers/Protobuf.mm
```diff
@@ -345,13 +345,15 @@ static inline void EncodeCertificateInfo(::pbv1::CertificateInfo *pb_cert_info,
     case SNTEventStateAllowCompilerSigningID: return ::pbv1::Execution::REASON_SIGNING_ID;
     case SNTEventStateAllowCDHash: return ::pbv1::Execution::REASON_CDHASH;
     case SNTEventStateAllowCompilerCDHash: return ::pbv1::Execution::REASON_CDHASH;
+    case SNTEventStateAllowCELFallback: return ::pbv1::Execution::REASON_UNKNOWN;
```
tburgin: We should have a new reason for these, same on L356.

russellhancox: Good call, done

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -66,6 +68,13 @@ @interface SNTPolicyProcessor () {
   std::unique_ptr<santa::cel::Evaluator<false>> celEvaluatorV1_;
   std::unique_ptr<santa::cel::Evaluator<true>> celEvaluatorV2_;
   std::shared_ptr<santa::EntitlementsFilter> entitlementsFilter_;
+  // Arena used for constant folding during compilation of fallback expressions.
+  // Declared before celFallbackExpressions_ so that C++ reverse destruction
+  // order destroys expressions first, then the arena they reference.
+  std::shared_ptr<google::protobuf::Arena> celFallbackArena_;
+  std::vector<std::shared_ptr<::google::api::expr::runtime::CelExpression>> celFallbackExpressions_;
+  dispatch_queue_t celFallbackQueue_;
+  SNTKVOManager *celFallbackExpressionsObserver_;
```
mlw: nit: move these two to be regular properties.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -107,28 +132,111 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable
...
+- (void)compileFallbackExpressions:(NSArray<NSString *> *)expressions {
+  if (!celEvaluatorV2_) {
+    return;
+  }
+
+  // Create a fresh arena for this batch of compiled expressions.
+  // The arena must outlive the compiled plans since constant folding stores
+  // data on it.
+  __block auto arena = std::make_shared<google::protobuf::Arena>();
+  __block std::vector<std::shared_ptr<::google::api::expr::runtime::CelExpression>> compiled;
```
mlw: Reserve size on `compiled` to mitigate internal vector reallocations / moves.

Source/santasyncservice/SNTSyncConfigBundle.mm
```diff
@@ -46,6 +46,7 @@ @interface SNTConfigBundle (ConfigBundleCreator)
 @property SNTSyncNetworkExtensionSettings *networkExtensionSettings;
 @property NSArray<NSString *> *pushTokenChain;
 @property NSArray<NSString *> *telemetryFilterExpressions;
+@property NSArray<NSString *> *celFallbackExpressions;
```
mlw: nit: add to the test too

Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -438,6 +438,21 @@ void HandleV2Responses(const ::pbv2::PreflightResponse &resp, SNTSyncState *sync
     syncState.telemetryFilterExpressions = [expressions copy];
   }
 
+  // Always set CEL fallback expressions (even if empty) to allow the server to clear them.
+  // Limit to 10 expressions.
```
mlw: outside scope here - but just calling out we should tackle `SNT-183` soon

Source/santad/SNTPolicyProcessor.mm
mlw, line 1: As a general comment - the policy processor is getting pretty massive with this change. A lot of the new code seems like it could stand on its own - not needing things from the SNTPolicyProcessor. Next time we're messing with CEL features we  should consider abstracting this into its own class.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -107,28 +132,111 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable
   return self;
 }
 
-- (CELEvaluationResult)evaluateCELExpressionForRule:(SNTRule *)rule
-                                     cachedDecision:(SNTCachedDecision *)cd
-                                 activationCallback:(ActivationCallbackBlock)activationCallback {
-  bool useV2 = (rule.state == SNTRuleStateCELv2);
-  auto activation = activationCallback(useV2);
+- (void)compileFallbackExpressions:(NSArray<NSString *> *)expressions {
+  if (!celEvaluatorV2_) {
```
mlw, line 142: nit: There is a check on the sync service side, but might be also nice here to ensure that the number of expressions doesn't exceed the max allowed.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -167,6 +274,9 @@ - (CELEvaluationResult)evaluateCELExpressionForRule:(SNTRule *)rule
   if (useV2) {
     using ReturnValue = santa::cel::CELProtoTraits<true>::ReturnValue;
     switch (static_cast<ReturnValue>(returnValue)) {
+      case ReturnValue::UNSPECIFIED:
+        LOGD(@"CEL expression returned UNSPECIFIED, skipping");
+        return {.succeeded = false, .decisionMade = false, .resultState = {}};
       case ReturnValue::ALLOWLIST: resultState = SNTRuleStateAllow; break;
       case ReturnValue::ALLOWLIST_COMPILER: resultState = SNTRuleStateAllowTransitive; break;
```
mlw: Existing issue - but is this correct? This seems like we're mapping the wrong value, yeah? (same in v1 path)

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -107,28 +132,111 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable
...
+      if (celResult.decisionMade) {
+        return YES;
+      }
+      continue;
+    }
+
+    cd.decision = (celResult.resultState == SNTRuleStateAllow ||
+                   celResult.resultState == SNTRuleStateAllowTransitive)
+                      ? SNTEventStateAllowCELFallback
+                      : SNTEventStateBlockCELFallback;
```
mlw, line 228: I'm a bit lost on how `ALLOWLIST_COMPILER` results get handled

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -107,28 +132,111 @@ - (instancetype)initWithRuleTable:(SNTRuleTable *)ruleTable
...
+
+  for (size_t i = 0; i < expressions.size(); ++i) {
+    LOGD(@"Evaluating CEL fallback expression %zu", i);
+
+    CELEvaluationResult celResult = [self evaluateCompiledCELExpression:expressions[i].get()
+                                                                  useV2:true
+                                                         cachedDecision:cd
+                                                             activation:*activation
+                                                              evalArena:&evalArena];
+
```
mlw, line 217: I think we should chat about supporting some common "rule" fields for fallback expressions - i.e. custom URL / Text / Message options. (This would be a followup PR.)

Source/santasyncservice/SNTSyncEventUpload.mm
```diff
@@ -227,6 +227,8 @@ BOOL EventUpload(SNTSyncEventUpload *self, NSArray<SNTStoredEvent *> *events) {
     case SNTEventStateBlockTeamID: e->set_decision(Traits::BLOCK_TEAMID); break;
     case SNTEventStateBlockSigningID: e->set_decision(Traits::BLOCK_SIGNINGID); break;
     case SNTEventStateBlockCDHash: e->set_decision(Traits::BLOCK_CDHASH); break;
+    case SNTEventStateBlockCELFallback: e->set_decision(Traits::BLOCK_UNKNOWN); break;
+    case SNTEventStateAllowCELFallback: e->set_decision(Traits::ALLOW_UNKNOWN); break;
```
mlw: We lose fidelity here, it might be nice to communicate in the events if this was allowed/blocked due to a fallback cel expression.

Source/santad/SNTPolicyProcessor.mm
russellhancox: No this was wrong. Good catch

Source/santad/SNTPolicyProcessor.mm
russellhancox: This should have been AllowCompiler, not AllowTransitive.

Source/common/SNTConfigBundle.h
```diff
@@ -49,5 +50,6 @@
 - (void)networkExtensionSettings:(void (^)(SNTSyncNetworkExtensionSettings *))block;
 - (void)pushTokenChain:(void (^)(NSArray<NSString *> *))block;
 - (void)telemetryFilterExpressions:(void (^)(NSArray<NSString *> *))block;
+- (void)celFallbackRules:(void (^)(NSArray<SNTCELFallbackRule *> *))block;
```
mlw, line 53: nit - add to test

Source/common/SNTConfigurator.mm
```diff
@@ -1878,6 +1885,36 @@ - (void)setSyncServerTelemetryFilterExpressions:(NSArray<NSString *> *)expressio
...
+    return nil;
+  }
+  return rules;
+}
+
+- (void)setSyncServerCELFallbackRules:(NSArray<SNTCELFallbackRule *> *)rules {
+  NSData *data = nil;
+  if (rules) {
+    NSError *error;
+    data = [NSKeyedArchiver archivedDataWithRootObject:rules
```
mlw: We should move this to `serialize`/`deserialize` methods on `SNTCELFallbackRule` (e.g. similar to `SNTExportConfiguration` or `SNTSyncNetworkExtensionSettings`)

Source/santasyncservice/ProtoTraits.h
```diff
@@ -73,6 +73,9 @@ struct ProtoTraits<false> {
   static constexpr Decision BLOCK_SIGNINGID = ::santa::sync::v1::BLOCK_SIGNINGID;
   static constexpr Decision BLOCK_CDHASH = ::santa::sync::v1::BLOCK_CDHASH;
   static constexpr Decision BUNDLE_BINARY = ::santa::sync::v1::BUNDLE_BINARY;
+  // v1 doesn't have CEL fallback decisions; fall back to UNKNOWN.
+  static constexpr Decision ALLOW_CEL_FALLBACK = ::santa::sync::v1::ALLOW_UNKNOWN;
+  static constexpr Decision BLOCK_CEL_FALLBACK = ::santa::sync::v1::BLOCK_UNKNOWN;
```
mlw, line 78: Are these needed to compile? Ideally they shouldn't need to be defined.

russellhancox: They are but only because the value is referenced in the decision regardless of SyncV1/V2 in MessageForExecutionEvent. We could make that conditional and not define it here

PR #841, opened 2026-03-09
pkg: Add lite version

russellhancox:


Source/common/SNTLiteDetector.mm
```diff
@@ -0,0 +1,31 @@
...
+///    distributed under the License is distributed on an "AS IS" BASIS,
+///    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+///    See the License for the specific language governing permissions and
+///    limitations under the License.
+
+#import "Source/common/SNTLiteDetector.h"
+#import "Source/common/SNTCommonEnums.h"
+
+BOOL SNTIsLiteAppBundle(NSString *appPath) {
+  NSString *plistPath = [appPath stringByAppendingPathComponent:@"Contents/Info.plist"];
```
pmarkowsky, line 21: nit: might be worth checking our signature in case someone has tampered with it.

Source/common/SNTLiteDetector.h
```diff
@@ -0,0 +1,21 @@
...
+///    Unless required by applicable law or agreed to in writing, software
+///    distributed under the License is distributed on an "AS IS" BASIS,
+///    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+///    See the License for the specific language governing permissions and
+///    limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+/// Returns YES if this is a Lite installation (sleigh or santanetd is missing).
+BOOL SNTIsLiteInstall(void);
```
mlw, line 20: nit: We should wrap these in the `santa` namespace. Update - maybe there are issues with this because of Swift interop? Not sure offhand...

Source/santad/SleighLauncher.mm
```diff
@@ -45,6 +45,10 @@
 
 absl::Status SleighLauncher::Launch(const std::vector<std::string> &input_files,
                                     uint32_t timeout_secs) {
+  if (access(sleigh_path_.c_str(), X_OK) != 0) {
+    return absl::NotFoundError("Sleigh binary not found: " + sleigh_path_);
```
pmarkowsky: The error message is a little misleading as the binary may be present. Shouldn't we also check `errno` for `ENOENT`?

pmarkowsky (commented):
Overall looks good. Some minor nits.

Source/common/SNTLiteDetector.h
```diff
@@ -0,0 +1,25 @@
...
+
+#import <Foundation/Foundation.h>
+
+namespace santa {
+
+/// Returns YES if this is a Lite installation (sleigh or santanetd is missing).
+BOOL SNTIsLiteInstall(void);
+
+/// Returns YES if the app bundle at the given path would be a Lite installation.
+BOOL SNTIsLiteAppBundle(NSString *appPath);
```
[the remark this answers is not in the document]
russellhancox: The plist is the correct form but the result is that santanetd and sleigh are missing

Source/common/SNTLiteDetector.h
russellhancox: Done

Source/santad/SleighLauncher.mm
russellhancox: Updated to say that it's not executable. I don't really want to print the exact error as we're only _really_ interested in whether it exists and can be executed.

Source/common/SNTLiteDetector.mm
russellhancox: As discussed, we do validation elsewhere and a faliure to validate here can't be sensibly handled.

PR #843, opened 2026-03-13
santactl/fileinfo: Add Expected Decision field

russellhancox:
Given that many entries cannot be determined statically there are many cases where this will print "Unknown" but it's best-effort and useful in many cases.


pmarkowsky (approved):
LGTM.

PR #845, opened 2026-03-16
config: Persist sync intervals, display in status

russellhancox:


Source/common/SNTConfigBundle.h
```diff
@@ -51,5 +51,7 @@
 - (void)pushTokenChain:(void (^)(NSArray<NSString *> *))block;
 - (void)telemetryFilterExpressions:(void (^)(NSArray<NSString *> *))block;
 - (void)celFallbackRules:(void (^)(NSArray<SNTCELFallbackRule *> *))block;
+- (void)fullSyncInterval:(void (^)(NSUInteger))block;
+- (void)pushNotificationsFullSyncInterval:(void (^)(NSUInteger))block;
```
mlw, line 55: nit - add to tests

Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -189,8 +189,9 @@ BOOL Preflight(SNTSyncPreflight *self, google::protobuf::Arena *arena,
   // Don't let these go too low
   uint64_t value = resp.push_notification_full_sync_interval_seconds()
                        ?: resp.deprecated_fcm_full_sync_interval_seconds();
-  self.syncState.pushNotificationsFullSyncInterval =
-      (value < kMinimumFullSyncInterval) ? kMinimumFullSyncInterval : value;
+  if (value > 0) {
+    self.syncState.pushNotificationsFullSyncInterval = @(MAX(value, kMinimumFullSyncInterval));
+  }
```
mlw: There might be an issue here - I think we may want to always set the interval (same with the change below in this file).

The scenario that seems off:

1. `santasyncservice` starts up, happily syncs, and we set some sync interval to 60s
2. This interval is sent to `santad`, which gets persisted in `sync-state.plist`
3. `santasyncservice` restarts (or the system restarts or Santa re-installs)
4. On startup, `santasyncservice` will sync in 15s
5. If no interval settings are received from the sync server, the default values will be used (e.g. in Source/santasyncservice/SNTSyncManager.mm).
6. `santactl status` will continue reporting 60s received from the prior sync, despite now operating at the default interval

If we always set the interval, falling back to the default like used to be done, then we should always be sending the proper value over to santad to start in the sync state. I think that'll ensure we never get out of date values.

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -409,11 +409,13 @@ - (SNTSyncStatusType)preflightWithSyncState:(SNTSyncState *)syncState {
       [self rescheduleTimerQueue:self.fullSyncTimer
                   secondsFromNow:self.pushNotifications.fullSyncInterval];
     } else {
-      LOGD(@"Push notifications are not enabled. Sync every %lu min.",
-           syncState.fullSyncInterval / 60);
+      NSUInteger interval = syncState.fullSyncInterval
+                                ? syncState.fullSyncInterval.unsignedIntegerValue
+                                : kDefaultFullSyncInterval;
```
mlw, line 414: This is maybe not needed anymore now that we always set in preflight?

mlw (approved):

Source/santasyncservice/SNTSyncManager.mm
russellhancox: It might not be but it can't hurt to have it as a fallback

PR #849, opened 2026-03-17
config: added a configuration key to allowlist Santa commands

sharvilshah:
Fixes SNT-324

Adds a new MDM only key `AllowedSantaCommands` which is an array of allowed commands. 
If the key is unset (nil), all commands are allowed. If the key is an empty array, none of the commands are allowed. 

#### When `kill` is allowed

Screenshot 2026-03-19 at 12 09 52 AM

#### When `kill` is NOT allowed

Screenshot 2026-03-19 at 12 10 06 AM

#### Test Plan

Unit tests and manual testing

### DIFF
--- a/Source/santactl/Commands/SNTCommandStatus.mm
+++ b/Source/santactl/Commands/SNTCommandStatus.mm
@@ -362,12 +362,13 @@ REGISTER_COMMAND_NAME(@"status")
         @"enabled" : @(networkExtensionEnabled),
         @"loaded" : @(networkExtensionLoaded),
       };
-    }
 
-    if (allowedCommands) {
-      NSMutableDictionary *daemon = [stats[@"daemon"] mutableCopy];
-      daemon[@"allowed_commands"] = [allowedCommands sortedArrayUsingSelector:@selector(compare:)];
-      stats[@"daemon"] = daemon;
+      if (allowedCommands) {
+        NSMutableDictionary *daemon = [stats[@"daemon"] mutableCopy];
+        daemon[@"allowed_commands"] =
+            [allowedCommands sortedArrayUsingSelector:@selector(compare:)];
+        stats[@"daemon"] = daemon;
+      }
     }
 
     if (syncURLStr.length) {
@@ -459,9 +460,9 @@ REGISTER_COMMAND_NAME(@"status")
       } else {
         printf("\n");
       }
+      printf("  %-40s | %s\n", "Allowed Commands", [allowedStr UTF8String]);
     }
     printf("  %-40s | %lld\n", "Static Rules", staticRuleCount);
-    printf("  %-40s | %s\n", "Allowed Commands", [allowedStr UTF8String]);
     printf("  %-40s | %lld  (Peak: %.2f%%)\n", "Watchdog CPU Events", cpuEvents, cpuPeak);
     printf("  %-40s | %lld  (Peak: %.2fMB)\n", "Watchdog RAM Events", ramEvents, ramPeak);
 


Source/common/SNTConfigurator.mm
```diff
@@ -1770,6 +1776,10 @@ - (NSDictionary *)extraMetricLabels {
   return annotations;
 }
 
+- (NSArray<NSString *> *)allowedSantaCommands {
+  return EnsureArrayOfStrings(self.configState[kAllowedSantaCommandsKey]);
+}
```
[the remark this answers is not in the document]
sharvilshah: This has been fixed now. Check again

Source/santasyncservice/SNTPushClientNATSCommandTest.mm
```diff
@@ -788,6 +788,80 @@ - (void)testNonceCacheThrottling {
...
+  [self signCommandRequest:&command];
+
+  // When: Dispatching a ping command
+  ::pbv1::SantaCommandResponse *response = [self.client dispatchSantaCommandToHandler:command
+                                                                              onArena:self.arena];
+
+  // Then: Should succeed
+  XCTAssertFalse(response->has_error(), @"Ping command should be allowed when in allowlist");
+  XCTAssertEqual(response->result_case(), ::pbv1::SantaCommandResponse::kPing);
+}
```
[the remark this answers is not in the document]
sharvilshah: Fixed

Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -448,6 +456,14 @@ - (void)runWithArguments:(NSArray *)arguments {
       }
     }
     printf("  %-40s | %lld\n", "Static Rules", staticRuleCount);
+    NSArray<NSString *> *allowedCommands = configurator.allowedSantaCommands;
```
russellhancox: Can we make a variable for this once and re-use it, like we do with almost all the other variables here?

Source/santactl/Commands/SNTCommandStatus.mm
```diff
@@ -448,6 +456,14 @@ - (void)runWithArguments:(NSArray *)arguments {
...
+    if (allowedCommands) {
+      NSString *allowedStr = allowedCommands.count > 0
+                                 ? [[allowedCommands sortedArrayUsingSelector:@selector(compare:)]
+                                       componentsJoinedByString:@", "]
+                                 : @"None (all blocked)";
```
russellhancox: I think it might be worth still printing this if the value is "All" just to be super clear

Source/santad/SNTDaemonControlController.mm
```diff
@@ -609,6 +609,13 @@ - (void)saveStatsSubmissionAttemptTime:(NSDate *)timestamp version:(NSString *)v
 #pragma mark Command Ops
 
 - (void)killProcesses:(SNTKillRequest *)killRequest reply:(void (^)(SNTKillResponse *))reply {
+  NSArray<NSString *> *allowed = [[SNTConfigurator configurator] allowedSantaCommands];
+  if (allowed && ![allowed containsObject:@"kill"]) {
+    LOGW(@"Kill command rejected - not in AllowedSantaCommands");
+    reply([[SNTKillResponse alloc] initWithError:SNTKillResponseErrorInvalidRequest]);
+    return;
+  }
```
russellhancox: This seems a bit overkill - it's already checked in SNTPushClientNATS+Commands, which should be the only caller of this.

Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -324,6 +325,24 @@ - (BOOL)checkAndRecordNonce:(NSString *)uuid {
...
 
   ::pbv1::SantaCommandRequest::CommandCase commandCase = command.command_case();
+
+  // Check if command type is allowed by client configuration
+  NSString *commandName = nil;
+  switch (commandCase) {
+    case ::pbv1::SantaCommandRequest::kPing: commandName = @"ping"; break;
+    case ::pbv1::SantaCommandRequest::kKill: commandName = @"kill"; break;
+    default: break;  // Unknown commands are handled by the existing default case
+  }
```
russellhancox: I'm pretty certain the generated API has a function for getting a string for an enum, something like

```
std::string commandName = ::pbv1::SantaCommandRequest_Name(commandCase)
```

Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -324,6 +325,24 @@ - (BOOL)checkAndRecordNonce:(NSString *)uuid {
   }
 
   ::pbv1::SantaCommandRequest::CommandCase commandCase = command.command_case();
+
+  // Check if command type is allowed by client configuration
+  NSString *commandName = nil;
+  switch (commandCase) {
+    case ::pbv1::SantaCommandRequest::kPing: commandName = @"ping"; break;
+    case ::pbv1::SantaCommandRequest::kKill: commandName = @"kill"; break;
+    default: break;  // Unknown commands are handled by the existing default case
```
russellhancox: We should have no default here, or an assertion (and a test?) so that we can tell if there's an unhandled case.

Source/santactl/Commands/SNTCommandStatus.mm
sharvilshah: Sorry, I am a little confused, currently `Allowed Commands` is not printed in the `santactl status` output when the `allowedSantaCommands` key is unset (which means all commands are allowed). Are you suggesting that we print All when this is the case? 

Or support a value of `all` in the `AllowedSantaCommands` array?

russellhancox: I'm saying that `santactl status` should always print the `Allowed Commands` key and if the key is unset we print the string `All`

Source/santad/SNTDaemonControlController.mm
sharvilshah: I thought the `santactl command kill ...` uses this..?

russellhancox: That subcommand only exists in debug builds

sharvilshah: Gotcha, thanks!

Source/santactl/Commands/SNTCommandStatus.mm
mlw: To add to this - we should only print the "Allowed Commands" line if sync v2 is enabled.

Source/santasyncservice/SNTPushClientNATS+Commands.mm
```diff
@@ -324,6 +325,26 @@ - (BOOL)checkAndRecordNonce:(NSString *)uuid {
...
+    case ::pbv1::SantaCommandRequest::kKill: commandName = @"kill"; break;
+    case ::pbv1::SantaCommandRequest::kEventUpload: commandName = @"event_upload"; break;
+    case ::pbv1::SantaCommandRequest::COMMAND_NOT_SET: break;
+  }
+
+  if (commandName) {
+    NSArray<NSString *> *allowed = [[SNTConfigurator configurator] allowedSantaCommands];
+    if (allowed && ![allowed containsObject:commandName]) {
+      LOGW(@"NATS: Command '%@' rejected - not in AllowedSantaCommands", commandName);
+      response->set_error(::pbv1::SantaCommandResponse::ERROR_UNKNOWN_REQUEST_TYPE);
```
mlw: I think the error code `ERROR_COMMAND_DISABLED` should be used.

@pmarkowsky - wanted to confirm - that error code was meant for this purpose, yeah?

Source/common/SNTConfigurator.mm
```diff
@@ -1794,6 +1800,30 @@ - (NSDictionary *)extraMetricLabels {
...
+- (NSArray<NSString *> *)allowedSantaCommands {
+  NSArray *commands = self.configState[kAllowedSantaCommandsKey];
+  if (!commands) {
+    return nil;
+  }
+
+  if (![commands isKindOfClass:[NSArray class]]) {
+    LOGW(@"AllowedSantaCommands is not an array, blocking all commands");
+    return @[];
+  }
```
mlw: Checking the type of the container here is redundant, this is enforced when you specify the type `array` in the list of `_forcedConfigKeyTypes` above. (Note: keep the check below where you verify each item in the container is a NSString since that is not enforced above.)

sharvilshah: Ah, good to know, thanks! I was being very defensive, and thinking, what can folks do to bypass this

Source/santasyncservice/SNTPushClientNATS+Commands.mm
sharvilshah: Yeah, I was thinking in terms of request/response error codes, but this makes more sense, updated!

Source/santactl/Commands/SNTCommandStatus.mm
sharvilshah: @mlw for printing only when syncv2 enabled, even though this is a config only setting?

mlw: My concern is that for non-syncv2 users, showing "Allowed Commands: All" would be confusing. There are no commands as it isn't a feature for non-syncv2 deployments. This would be similar to how other things aren't included in the output when syncv2 isn't being used - such as network extension status.

sharvilshah: Makes sense, updated

PR #869, opened 2026-03-25
metrics: Add sync-based export

russellhancox:


Source/santasyncservice/SNTSyncPublishMetrics.mm
```diff
@@ -0,0 +1,136 @@
...
+          SetMetricValue(value, entry[@"data"], type);
+        }
+      }
+    }
+  }
+
+  for (NSString *key in metrics[@"root_labels"]) {
+    NSString *value = metrics[@"root_labels"][key];
+    (*request->mutable_root_labels())[[key UTF8String]] = value ? [value UTF8String] : "";
+  }
```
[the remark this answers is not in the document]
russellhancox: False positive - the 2 metric labels that contain a date (`created` and `last_updated`) are not populated into the proto.

Source/common/SNTConfigurator.mm
```diff
@@ -1757,21 +1764,28 @@ - (SNTMetricFormatType)metricFormat {
     return SNTMetricFormatTypeRawJSON;
   } else if ([normalized isEqualToString:@"monarchjson"]) {
     return SNTMetricFormatTypeMonarchJSON;
+  } else if ([normalized isEqualToString:@"proto"]) {
+    return SNTMetricFormatTypeProto;
+  } else if (normalized.length == 0 && santa::IsDomainPinned([self syncBaseURL])) {
```
mlw, line 1770: Wanted to confirm explicit usage of `IsDomainPinned` instead of `isSyncV2Enabled`

Source/santad/Santad.mm
```diff
@@ -102,6 +102,15 @@ void SantadMain(std::shared_ptr<EndpointSecurityAPI> esapi, std::shared_ptr<Logg
             return auth_result_cache->CheckCache(vnode).action;
           }];
 
+  std::weak_ptr<Metrics> weak_metrics(metrics);
+  dc.metricsExportBlock = ^(void (^reply)(BOOL)) {
```
mlw: Can we move this into the initializer

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -216,6 +217,31 @@ - (void)pushNotificationReconnect {
   [self syncSecondsFromNow:2];
 }
 
+- (void)publishMetrics:(NSDictionary *)metrics reply:(void (^)(BOOL))reply {
```
mlw, line 225: It looks like we could have simultaneous uploads - via normal metrics interval as well `santactl`. While this seems like it would largely be safe (modulo the xsrf stuff), could there be a good reason to serialize this operation on some new queue? (I wouldn't want to block the normal sync queue)

Source/santasyncservice/SNTSyncPublishMetrics.mm
```diff
@@ -0,0 +1,136 @@
...
+          }
+          SetMetricValue(value, entry[@"data"], type);
+        } else {
+          Value *value = protoMetric->add_values();
+          SetMetricValue(value, entry[@"data"], type);
+        }
+      }
+    }
+  }
+
```
mlw, line 96: We're not setting the machine_id [field in the proto](https://github.com/northpolesec/protos/blob/96ef1e30fd33c455b90ca40815fcda789a2ee609/syncv2/v2.proto#L1316), maybe because it's part of the URL?

Source/common/SNTConfigurator.mm
russellhancox: Yeah - we can't use isSyncV2Enabled because that's only accessible from the daemon right now. This is only a nice-to-have so that existing users don't need to update their profiles to get metrics, but we'll include this key in the generated config anyway.

Source/santad/Santad.mm
russellhancox: Done

Source/santasyncservice/SNTSyncManager.mm
russellhancox: Done

Source/santasyncservice/SNTSyncPublishMetrics.mm
russellhancox: Done

PR #921, opened 2026-04-21
sync: enforce HTTP for loopback addresses only

sharvilshah:
Fixes SNT-386

### DIFF
--- a/Source/common/SNTConfigurator.mm
+++ b/Source/common/SNTConfigurator.mm
@@ -566,6 +566,10 @@ static SNTConfigurator* sharedConfigurator = nil;
   return [self configStateSet];
 }
 
++ (NSSet*)keyPathsForValuesAffectingSyncBaseURLConfigured {
+  return [self configStateSet];
+}
+
 + (NSSet*)keyPathsForValuesAffectingSyncEnableProtoTransfer {
   return [self configStateSet];
 }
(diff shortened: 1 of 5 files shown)


Source/common/SNTConfigurator.mm
```diff
@@ -1124,9 +1128,27 @@ - (NSURL*)syncBaseURL {
     urlString = [urlString stringByAppendingString:@"/"];
   }
   NSURL* url = [NSURL URLWithString:urlString];
+
+  // Only allow HTTP for loopback addresses. Everything else must use HTTPS.
+  if ([[url.scheme lowercaseString] isEqualToString:@"http"]) {
```
mlw, line 1133: Probably worth a couple simple test cases in `SNTConfiguratorTest.mm`.

mlw, line 1133: Also update docs/src/lib/santaconfig.ts with the new requirements.

Source/santactl/Commands/SNTCommandDoctor.mm
mlw: Preexisting issue - but shouldn't this be `return NO`?

mlw (dismissed):

Source/common/SNTConfigurator.mm
sharvilshah: Adding!

Source/santactl/Commands/SNTCommandDoctor.mm
sharvilshah: It is a bit weird, but `validateSync` returns `YES` when something is amiss, and returns `NO` when everything is a-ok

mlw (approved):

PR #927, opened 2026-04-26
Add identity verification step to exec policy evaluation

mlw:
Also includes a `MOLCodesignChecker` change: when a file descriptor is supplied to `initWithBinaryPath:fileDescriptor:`, route `SecStaticCodeCreateWithPath` through `/dev/fd/N` so the static code ref operates on the caller's vnode. This avoids a redundant path resolution and aligns disk-side identity with the bytes the caller already has open.

### DIFF
--- a/Source/common/MOLCodesignChecker.h
+++ b/Source/common/MOLCodesignChecker.h
@@ -155,11 +155,16 @@
 - (instancetype)initWithBinaryPath:(NSString*)binaryPath;
 
 /**
-  Wrapper around initWithBinaryPath:error: that takes a file descriptor for reading.
-  The file descriptor will be used to read binary header infomation. This provides a minor
-  performace increase if the caller already has the file open.
+  Initialize with a binary on disk via a caller-supplied file descriptor.
 
-  @note The file offset will be set to the amount of bytes read while parsing the header.
+  The descriptor must refer to a regular-file Mach-O. SecStaticCode operates
+  on the descriptor's vnode (via /dev/fd/N) rather than re-resolving
+  `binaryPath`, so a rename of `binaryPath` between the caller's open() and
+  this call does not affect the result. The caller-supplied path is retained
+  for display via the `binaryPath` accessor only.
+
+  @note The file offset will be set to the amount of bytes read while parsing
+  the header.
 */
 - (instancetype)initWithBinaryPath:(NSString*)binaryPath
                     fileDescriptor:(int)fileDescriptor
--- a/Source/common/MOLCodesignChecker.h
+++ b/Source/common/MOLCodesignChecker.h
@@ -16,6 +16,7 @@
 @class MOLCertificate;
 
 #import <Foundation/Foundation.h>
+#import <mach/machine.h>
 
 /**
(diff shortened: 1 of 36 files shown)


Source/santad/SNTPolicyProcessor.mm
sharvilshah: Should we not add one more check here: `&& !identityExtracted`? (Or call `UpdateCachedDecisionSigningInfo` like we do on line 695)

sharvilshah (dismissed):
Just a question, but lgtm

Source/common/MOLCodesignChecker.mm
```diff
@@ -30,6 +30,24 @@
 using ScopedCFString = santa::ScopedCFTypeRef<CFStringRef>;
 using ScopedSecStaticCode = santa::ScopedCFTypeRef<SecStaticCodeRef>;
 
+// Reads exactly `len` bytes from `fd` at `offset`, retrying on EINTR and
```
tburgin, line 33: nit: May be nice to have an HANDLE_EINTR macro. See:

https://source.chromium.org/chromium/chromium/src/+/main:base/posix/eintr_wrapper.h

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -519,18 +530,125 @@ static void UpdateCachedDecisionSigningInfo(
...
+        NSString* esTeam = @(targetProc->team_id.data);
+        if (![esTeam isEqualToString:csInfo.teamID]) {
+          LOGW(@"Identity verification: team ID mismatch path=%@ es_team=%@ disk_team=%@ sid=%@",
+               path, esTeam, csInfo.teamID, esSID);
+          return IdentityVerifyResult::kMismatch;
+        }
+      }
+      if (CdhashEqual(targetProc->cdhash, csInfo.cdhash)) {
+        return IdentityVerifyResult::kMatch;
+      }
```
tburgin, line 597: Should we test for CDHash equality first thing inside of the `if (esHasSID)` branch?

Source/common/MOLCodesignChecker.h
```diff
@@ -155,11 +156,16 @@
...
 
 /**
-  Wrapper around initWithBinaryPath:error: that takes a file descriptor for reading.
-  The file descriptor will be used to read binary header infomation. This provides a minor
-  performace increase if the caller already has the file open.
+  Initialize with a binary on disk via a caller-supplied file descriptor.
 
-  @note The file offset will be set to the amount of bytes read while parsing the header.
+  The descriptor must refer to a regular-file Mach-O. SecStaticCode operates
+  on the descriptor's vnode (via /dev/fd/N) rather than re-resolving
```
tburgin, line 162: Are we sure that having SecStaticCode looking `/dev/fd/N` paths is not masking signature failing errors behind non-failing errors like `errSecCSInfoPlistFailed`?

Source/common/MOLCodesignChecker.mm
mlw: Think this is a fair quality of life macro for the code base, but I think it would be better to split this out on its own PR. Filed SNT-408

mlw:
This is being superseded by #942

PR #934, opened 2026-05-05
sync: clear sync state on clean sync

sharvilshah:
Fixes SNT-357

### DIFF
--- a/Source/common/SNTConfigurator.mm
+++ b/Source/common/SNTConfigurator.mm
@@ -1959,11 +1959,17 @@ static SNTConfigurator* sharedConfigurator = nil;
 ///  immediately after the call completes.
 ///
 - (void)updateSyncStateForKey:(NSString*)key value:(id)value {
+  [self updateSyncStateForKey:key value:value writeToDisk:YES];
+}
+
+- (void)updateSyncStateForKey:(NSString*)key value:(id)value writeToDisk:(BOOL)writeToDisk {
   void (^block)(void) = ^{
     NSMutableDictionary* syncState = self.syncState.mutableCopy;
     syncState[key] = value;
     self.syncState = syncState;
-    [self saveSyncStateToDisk];
+    if (writeToDisk) {
+      [self saveSyncStateToDisk];
+    }
   };
   // Avoid deadlocks by directly calling the block when already running on the
   // main thread.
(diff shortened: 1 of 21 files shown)


Source/common/SNTConfigurator.mm
```diff
@@ -2061,9 +2067,133 @@ - (void)saveSyncStateToDisk {
...
+}
+
+- (void)replaceSyncStateWithBundle:(SNTConfigBundle*)bundle {
+  void (^block)(void) = ^{
+    NSMutableDictionary* newSyncState = [NSMutableDictionary dictionary];
+
+    NSArray* preservedChain = self.syncState[kPushTokenChainKey];
+    if ([preservedChain isKindOfClass:[NSArray class]] && preservedChain.count > 0) {
+      newSyncState[kPushTokenChainKey] = preservedChain;
+    }
```
tburgin: The push token chain comes down in every preflight, no need to have special handling of it.

Source/santactl/Commands/SNTCommandSync.mm
```diff
@@ -51,9 +51,14 @@ + (NSString*)longHelpText {
...
-          @"           requesting a clean sync from the server.\n"
+          @"           requesting a clean sync from the server. Also clears Workshop-managed\n"
+          @"           sync settings; pass --keep-old-settings to retain them.\n"
           @"  --clean-all: Perform a clean sync, erasing all existing rules and requesting a\n"
-          @"               clean sync from the server.\n"
+          @"               clean sync from the server. Also clears Workshop-managed sync\n"
+          @"               settings; pass --keep-old-settings to retain them.\n"
+          @"  --keep-old-settings: Only valid with --clean or --clean-all. Preserve existing\n"
+          @"                       Workshop-managed sync settings on disk during the clean\n"
+          @"                       sync, instead of clearing them.\n"
```
tburgin: What is the point of this? We are changing the default behavior to clear any lingering settings, which should have been the behavior all along. Adding an escape hatch here seems to just be adding complexity for no real gain, additionally it is not discoverable and I doubt anyone would ever call it. If we are worried that clearing lingering settings is not the right approach we should solve for that worry.

I vote to remove the flag and behavior.

Source/santad/SNTDaemonControlController.mm
```diff
@@ -609,6 +609,11 @@ - (void)updateSyncSettings:(SNTConfigBundle*)result reply:(void (^)(void))reply
   reply();
 }
 
+- (void)replaceSyncSettings:(SNTConfigBundle*)bundle reply:(void (^)(void))reply {
+  [[SNTConfigurator configurator] replaceSyncStateWithBundle:bundle];
+  reply();
+}
```
tburgin: Why do we need an new XPC method for this? Instead do this in updateSyncSettings:reply during a clean sync.

Source/common/SNTConfigurator.mm
sharvilshah: That's good to know, thanks

Source/santactl/Commands/SNTCommandSync.mm
sharvilshah: I too am of the opinion that we should remove this. We can add it later in case we need to. The point of concern was that this is technically a behavior change, in case someone relied on this.

Source/santad/SNTDaemonControlController.mm
sharvilshah: Yeah, makes sense, thanks! Updating

Source/santad/SNTDaemonControlController.mm
```diff
@@ -473,6 +473,19 @@ - (void)disableUnknownEventUpload:(void (^)(BOOL))reply {
...
+  // resetting it on disk). That exact transition is the signal that this
+  // bundle is a clean-sync postflight: wipe persisted sync state before
+  // applying so settings the server stops sending don't linger. The
+  // per-key setters that follow recreate `sync-state.plist` from the
+  // bundle's contents.
+  [result syncType:^(SNTSyncType val) {
+    if (val == SNTSyncTypeNormal) {
+      [configurator clearSyncState];
+    }
+  }];
```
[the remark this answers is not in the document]
sharvilshah: the wipe only fires on clean-sync postflight, ordinary syncs go through SNTSyncManager (preflight directly, no XPC bundle) and the postflight bundle sets syncType = nil, never @(Normal)

Source/santactl/Commands/SNTCommandSync.mm
mlw: Would like to discuss this a bit more offline.

Source/santasyncservice/SNTSyncConfigBundleTest.mm
```diff
@@ -258,4 +258,61 @@ - (void)testSyncTypeConfigBundle {
   XCTAssertNil(bundle.celFallbackRules);
 }
 
+// PostflightConfigBundle invariants: these tests pin behavior the SNT-357
+// atomic-replace path depends on. If a future refactor breaks either invariant,
+// `replaceSyncStateWithBundle:` would silently leave the on-disk plist in an
```
mlw: Comment is stale - this method was removed

Source/santad/SNTDaemonControlController.mm
```diff
@@ -473,6 +473,19 @@ - (void)disableUnknownEventUpload:(void (^)(BOOL))reply {
...
+  // PostflightConfigBundle sets the bundle's syncType to Normal only when
+  // the in-flight sync was Clean or CleanAll (the directive is consumed by
+  // resetting it on disk). That exact transition is the signal that this
+  // bundle is a clean-sync postflight: wipe persisted sync state before
+  // applying so settings the server stops sending don't linger. The
+  // per-key setters that follow recreate `sync-state.plist` from the
+  // bundle's contents.
+  [result syncType:^(SNTSyncType val) {
+    if (val == SNTSyncTypeNormal) {
+      [configurator clearSyncState];
```
mlw: Similar to the CR comment, this feels a bit brittle. Maybe we could add an explicit `clearSyncStateBeforeApply` field to `PostflightConfigBundle`. This would then be explicit from the caller and replace the in band signaling.

mlw: I'm concerned about the "clear then rebuild" strategy. We have several configurator keys that could report the wrong unintended values and KVO watchers that could potentially needlessly fire during the gap.

We should investigate building an in-memory dictionary, and doing an atomic swap.

mlw (dismissed):

Source/santad/SNTDaemonControlController.mm
tburgin: +1 - I was in the middle of writing this as a comment when I saw Matt's comment.

tburgin: +1

tburgin: I suggest adding a new method to SNTConfigurator that takes a SNTConfigBundle and does an atomic swap.

sharvilshah:
@coderabbitai full review

Source/common/SNTConfigurator.mm
```diff
@@ -2060,10 +2060,133 @@ - (void)saveSyncStateToDisk {
...
+    [[NSFileManager defaultManager] removeItemAtPath:self.syncStateFilePath error:NULL];
+  };
+  if ([NSThread isMainThread]) {
+    block();
+  } else {
+    dispatch_sync(dispatch_get_main_queue(), block);
+  }
+}
+
+- (void)atomicallyApplyBundle:(SNTConfigBundle*)bundle {
```
mlw, line 2074: We shouldn't duplicate all the logic from the individual `setSyncServer*` methods.

Source/common/SNTConfigurator.mm
```diff
@@ -2060,10 +2060,133 @@ - (void)saveSyncStateToDisk {
 }
 
 - (void)clearSyncState {
-  self.syncState = [NSMutableDictionary dictionary];
-  // TODO: Start a timer to flush the state to disk. On startup, Santa should
-  // check for the presence of the state file and, if no SyncBaseURL is
-  // configured, start the timer to clear sync state and flush to disk.
+  void (^block)(void) = ^{
```
mlw, line 2063: nit: we should add an "is authorized" check:
```
  if (!self.syncStateAccessAuthorizerBlock()) {
    return;
  }
```

Source/common/SNTConfigurator.mm
```diff
@@ -2060,10 +2060,133 @@ - (void)saveSyncStateToDisk {
...
+    if ([lower isEqualToString:@"auditonly"] || [lower isEqualToString:@"disable"] ||
+        [lower isEqualToString:@"none"] || [lower isEqualToString:@""]) {
+      newSyncState[kOverrideFileAccessActionKey] = v;
+    }
+  }];
+  [bundle exportConfiguration:^(SNTExportConfiguration* v) {
+    newSyncState[kExportConfigurationKey] = [v serialize];
+  }];
+  [bundle fullSyncLastSuccess:^(NSDate* v) {
+    newSyncState[kFullSyncLastSuccess] = v;
```
mlw, line 2145: Discrepancy - the rule sync last success is also set on the setter side, but not here.

Source/santad/SNTDaemonControlController.mm
```diff
@@ -473,6 +473,35 @@ - (void)disableUnknownEventUpload:(void (^)(BOOL))reply {
...
+    // Capture pre-swap CEL state so the post-swap comparison detects "rules
+    // disappeared from the server's config" as a change too — not just
+    // modifications.
+    NSData* oldCELData = [SNTCELFallbackRule serializeArray:[configurator celFallbackRules]];
+
+    [configurator atomicallyApplyBundle:result];
+
+    // Side-effects pass. Persistence is already done by atomicallyApplyBundle:
+    // for every slot except modeTransition.
+    [result modeTransition:^(SNTModeTransition* val) {
```
mlw, line 491: This will result in not being part of the atomic swap. I don't think there are any KVO watchers for it today, but if any were added in the future it would result in unexpected fires.

mlw (changes requested):

sharvilshah:
Closing this in favor of #944

PR #947, opened 2026-05-11
gui: Add ability to silence device notifications

russellhancox:
Add silences to device mount notifications

image

The silence key for a USB/SD mount is `usb:<model>|<uuid>` and network is `netmount:<host & path>`. I've also name-spaced the Binary and FAA silences with `binary:` and `faa:` to prevent the possibility of clashes. This does mean that any existing silences will not be respected after the new version starts for the first time.


Source/gui/SNTDeviceMessageWindowController.mm
```diff
@@ -37,15 +38,27 @@ - (void)showWindow:(id)sender {
...
 
 - (NSString*)messageHash {
-  return self.event.mntonname;
+  // Use model + media UUID — these are stable across re-mounts of the same
+  // physical device, unlike the user-controllable mount-on path or the
+  // per-mount BSD device path. If neither is available, return nil so no
+  // silence entry is recorded (better than silencing by an unstable key).
+  NSString* model = self.event.deviceModel;
+  NSString* uuid = self.event.mediaUUID;
+  if (!model.length && !uuid.length) return nil;
```
mlw, line 61: This will result in a bit of a weird state where the user could select "prevent notifications..." but nothing would be silenced and there'd be no feedback about it. Could we maybe remove the checkbox when there is nothing to silence?

mlw (approved):

PR #951, opened 2026-05-12
verifyinghasher: add per-call skip_page_hash option

mlw:
Adds Options.skip_page_hash to VerifyingHasherCore and a corresponding RunOptions.skip_page_hash on the public VerifyingHasher facade. When set, per-page CodeDirectory verification is bypassed while cdhash, full-file SHA-256, signing-id, and team-id extraction are preserved. Intended for callers with independent assurance of page-hash integrity.

### Public surface

| Component | Field | Purpose |
| --- | --- | --- |
| `VerifyingHasherCore::Options` | `bool skip_page_hash` | per-`Run()` knob (default false) | | `VerifyingHasher::RunOptions` | `bool skip_page_hash` | facade-level mirror | | `VerifyingHasherCore::Mismatches()` | `std::optional<uint32_t>` | `std::nullopt` under skip | | `VerifyingHasherCore::PageHashSkipped()` | `bool` | returns `Options.skip_page_hash` |

`Mismatches()` returning `std::nullopt` under skip makes a caller that ignored `PageHashSkipped()` fail loudly (`.value()` throws) instead of silently consuming a falsely-clean count.

Two `Run()` overloads on the facade (4-arg forwarder + 5-arg primary) work around clang's complete-class-context restriction on a defaulted `const RunOptions& opts = {}` parameter inside the enclosing class.

### Implementation

Dispatch in `RunStreamingPhases<>` substitutes `NoopHashTraits` from `HashTraits.h` when `skip_page_hash` is set. The if-constexpr branch in `PageVerifierT::Update` performs an O(1) bulk-advance of `cur_slot_` / `cur_page_bytes_` instead of running the per-page hashing loop. The iterative loop is the if-constexpr discarded substatement under `NoopHashTraits` and is not instantiated, which lets the trait keep `kDigestSize = 0`. The gap-detect prologue and end-state arithmetic apply on both code paths, so `StreamCorrupt()` and `Complete()` defense-in-depth remain active under skip.

Apple's CodeDirectory only ships pow2 page sizes (4096, 16384), so the constructor detects pow2 via `std::has_single_bit` and stores `page_size_shift_` / `page_size_mask_`. The hot path uses shift+mask instead of `UDIV`/`UREM` — the latter would serialize against the SHA-256 stream in cold-cache regimes where OoO slack is tighter. A UDIV fallback covers hypothetical non-pow2 page sizes.

### Tests

- `PageVerifierTest`: bulk-advance Complete()/StreamCorrupt() invariants with partial-last-page, page-aligned, gap-detect, short-stream, and non-pow2 fallback cases.
- `HashTraitsTest`: `NoopHashTraits` size invariants.
- `VerifyingHasherCoreTest`: skip preserves digest equivalence with non-skip, bypasses tamper, preserves I/O and phase-1 errors, `PageHashSkipped()` reflects intent even on early failures.
- `VerifyingHasherTest`: facade-level skip on clean and tampered.
- `Testing/OneOffs/verifying_hasher_smoke.sh` Case 5: `-s` produces byte-identical cdhash and full-file digest vs. the no-flag run.
- Exerciser gains a `-s` flag and prints `page mismatches: (skipped)`.

### DIFF
--- a/Source/common/verifyinghasher/HashTraits.h
+++ b/Source/common/verifyinghasher/HashTraits.h
@@ -120,6 +120,32 @@ struct Sha384Traits {
   static constexpr uint8_t kCsHashType = CS_HASHTYPE_SHA384;
 };
 
+// NoopHashTraits substitutes into PageVerifierT<> to make the per-page
+// hashing work disappear. Used when
+// VerifyingHasherCore::Options::skip_page_hash is set — e.g., when the caller
+// has independent assurance that the kernel will enforce page hashes (CS_HARD /
+// CS_KILL on AUTH EXEC), or via a future configuration escape hatch.
+// PageVerifierT<>::Update has an if-constexpr branch for this trait that
+// bulk-advances cur_slot_/cur_page_bytes_ in O(1) per chunk instead of running
+// the iterative per-page hashing loop; the iterative branch is the if-constexpr
+// discarded statement for this trait and is never instantiated, which is why
+// kDigestSize can be zero.
+//
+// Trait constants satisfy PageVerifierT's static_asserts trivially:
+//   kCompareSize (0) <= kSlotStride (0) <= kDigestSize (0).
+// Init/Update/Final are kept (rather than removed) only to model the trait
+// concept symmetrically with the real hash traits; the discarded-statement
+// rule means they're never called.
+struct NoopHashTraits {
+  struct Ctx {};
+  static constexpr size_t kDigestSize = 0;
+  static constexpr size_t kSlotStride = 0;
+  static constexpr size_t kCompareSize = 0;
+  static void Init(Ctx*) {}
+  static void Update(Ctx*, const void*, size_t) {}
+  static void Final(unsigned char*, Ctx*) {}
+};
+
 #undef VERIFYINGHASHER_CC_UPDATE_LOOP
 
 }  // namespace santa
--- a/Source/common/verifyinghasher/HashTraits.h
+++ b/Source/common/verifyinghasher/HashTraits.h
@@ -134,8 +134,11 @@ struct Sha384Traits {
 // Trait constants satisfy PageVerifierT's static_asserts trivially:
 //   kCompareSize (0) <= kSlotStride (0) <= kDigestSize (0).
 // Init/Update/Final are kept (rather than removed) only to model the trait
-// concept symmetrically with the real hash traits; the discarded-statement
-// rule means they're never called.
+// concept symmetrically with the real hash traits. Init is called once
+// unconditionally by PageVerifierT's constructor (no-op body). Update and
+// Final live in the iterative loop, which is the if-constexpr discarded
+// substatement under NoopHashTraits and is not instantiated, so those two
+// are never called.
 struct NoopHashTraits {
   struct Ctx {};
   static constexpr size_t kDigestSize = 0;
--- a/Source/common/verifyinghasher/PageVerifier.h
+++ b/Source/common/verifyinghasher/PageVerifier.h
@@ -18,11 +18,13 @@
 #include <os/overflow.h>
 
 #include <algorithm>
+#include <bit>
 #include <cassert>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
 #include <span>
+#include <type_traits>
 #include <vector>
 
 #include "Source/common/verifyinghasher/HashTraits.h"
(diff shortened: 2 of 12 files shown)


tburgin (dismissed):

Source/common/verifyinghasher/HashTraits.h
```diff
@@ -120,6 +120,32 @@ struct Sha384Traits {
...
+// CS_KILL on AUTH EXEC), or via a future configuration escape hatch.
+// PageVerifierT<>::Update has an if-constexpr branch for this trait that
+// bulk-advances cur_slot_/cur_page_bytes_ in O(1) per chunk instead of running
+// the iterative per-page hashing loop; the iterative branch is the if-constexpr
+// discarded statement for this trait and is never instantiated, which is why
+// kDigestSize can be zero.
+//
+// Trait constants satisfy PageVerifierT's static_asserts trivially:
+//   kCompareSize (0) <= kSlotStride (0) <= kDigestSize (0).
+// Init/Update/Final are kept (rather than removed) only to model the trait
```
russellhancox, line 136: Init _is_ called, isn't it? The constructor in PageVerifierT calls it unconditionally.

Source/common/verifyinghasher/PageVerifier.h
```diff
@@ -146,6 +203,13 @@ class PageVerifierT {
   uint32_t page_size_;
   std::span<const uint8_t> slot_hashes_;
 
+  // Pow2 page-size hints set in the constructor. shift_ == -1 means
+  // page_size_ is not a power of two; the bulk-advance falls back to UDIV.
+  // In production, Apple's CodeDirectory always uses pow2 page sizes
+  // (kSecCodeMagicEmbeddedSignature.pageSize == 4096 or 16384).
```
russellhancox: kSecCodeMagicEmbeddedSignature is a magic number. I'm assuming this is referring to the pageSize field on the CS_CodeDirectory struct?

Source/common/verifyinghasher/VerifyingHasher.h
```diff
@@ -62,8 +62,25 @@ class VerifyingHasher {
...
+    // Skip per-page CodeDirectory verification while still computing the
+    // full-file SHA-256, cdhash, and signing-id/team-id extraction.
+    // Threaded through to VerifyingHasherCore::Options::skip_page_hash;
+    // see Core's documentation for the full contract.
+    bool skip_page_hash = false;
+  };
+
+  static Result Run(int fd, cpu_type_t cputype, cpu_subtype_t cpusubtype,
+                    const Expected& expected, const RunOptions& opts);
+  // No-options overload. (We avoid `const RunOptions& opts = {}` directly
```
russellhancox, line 75: Does `const RunOptions& opts = {}` actually fail? Given that RunOptions is declared earlier within this class I'm struggling to understand why this would actually fail (and I didn't think the referenced error applied to default arguments)

Source/common/verifyinghasher/PageVerifierTest.mm
```diff
@@ -159,4 +160,80 @@ - (void)testStreamGapDetected {
   XCTAssertTrue(pv.StreamCorrupt());
 }
 
+// Under NoopHashTraits, PageVerifier replaces the per-page hashing loop
+// with an O(1) bulk advance of cur_slot_/cur_page_bytes_. Verify the end
+// state is observationally equivalent to the iterative path: Complete()
+// returns true once the full signed region has been streamed, even when
+// the last page is partial (code_len not page-aligned).
+- (void)testNoopBulkAdvanceCompletePartialLastPage {
+  // 123 KiB, page size 4096 -> 30 full pages + one 3072-byte partial page.
```
russellhancox: Isn't it 30 full pages + 2048? The code length is 123KiB (123*1024) - 1024 (because signed_lo is 1024) = 124928. 30 * 4096 = 122880, leaving 2048 bytes.

mlw: Fixed.

Source/common/verifyinghasher/PageVerifier.h
mlw: Fixed

Source/common/verifyinghasher/HashTraits.h
mlw: Clarified

Source/common/verifyinghasher/VerifyingHasher.h
mlw: This is apparently a very old compiler bug: https://github.com/llvm/llvm-project/issues/36032

We could get around it by declaring this class as `VerifyingHasherRunOptions` at the namespace scope, but then we'd lose encapsulation. Or skip making it an aggregate and having a constructor with a default `skip` param - but then we'd have to have positional arguments which will be more awkward when more options are added.

russellhancox (approved):

PR #956, opened 2026-05-15
celv2: expose AUDIT return value, trigger event upload with flag set

russellhancox:


Source/santad/SNTPolicyProcessor.mm
```diff
@@ -291,6 +291,13 @@ - (BOOL)evaluateCELFallbackExpressions:(SNTCachedDecision*)cd
       case ReturnValue::ALLOWLIST_COMPILER: resultState = SNTRuleStateAllowCompiler; break;
       case ReturnValue::BLOCKLIST: resultState = SNTRuleStateBlock; break;
       case ReturnValue::SILENT_BLOCKLIST: resultState = SNTRuleStateSilentBlock; break;
+      case ReturnValue::AUDIT:
+        // AUDIT is identical to ALLOWLIST on the client; the only difference
+        // is that the resulting event is flagged so the sync server can
+        // distinguish audit-only matches.
+        cd.auditReturn = YES;
```
mlw, line 298: Should AUDIT respect `cacheable`?

mlw (approved):

Source/santad/SNTPolicyProcessor.mm
russellhancox: It's a good question. If we assume the point of `AUDIT` is to safely test branches of a CEL expression to see if it catches more than intended, I'd argue that the performance impact of not caching would also be an unintended consequence. You'll still get the audit event for the first execution, giving the signal that you were looking for.

PR #973, opened 2026-05-26
santad: extend control XPC for network flow rules

mlw:
Extends the existing daemon control XPC surface to carry network flow rules on the daemon side. XPC delivery to santanetd and sync-server proto-field population (Phase 5 / Task 10) are out of scope.

- databaseRuleAddExecutionRules:fileAccessRules:… gains a networkFlowRules: parameter, forwarded to SNTRuleTable's combined 5-arg add method.
- struct RuleCounts gains networkFlow, populated from [ruleTable networkFlowRuleCount].
- databaseRulesHash: reply widens to three strings (execution / FAA / network-flow), reading the third from SNTRuleTableRulesHash.
- santactl status: reports the network flow rule count in the Network Extension section and the network flow rules hash (both gated on syncv2), in JSON and human-readable output.
- All callers updated to the new signatures: santactl rule/status, syncservice preflight/postflight/rule-download, and OCMock matchers. Preflight/postflight receive but do not yet populate the new proto fields.


Source/common/SNTXPCControlInterface.mm
```diff
@@ -59,20 +59,26 @@ + (void)initializeControlInterface:(NSXPCInterface*)r {
...
+      argumentIndex:2
+            ofReply:NO];
+
   [r setClasses:[NSSet setWithObjects:[NSArray class], [NSError class], nil]
-        forSelector:@selector(databaseRuleAddExecutionRules:
-                                            fileAccessRules:ruleCleanup:source:reply:)
+        forSelector:@selector
+        (databaseRuleAddExecutionRules:fileAccessRules:networkFlowRules:ruleCleanup:source:reply:)
       argumentIndex:1
             ofReply:YES];
```
russellhancox, line 83: This call is out of order with the argumentIndex. Is this call still needed? Could it be combined with the one now on line 67? Does the call on line 73 also need NSError?

russellhancox (approved):

Source/common/SNTXPCControlInterface.mm
mlw: Spoke offline - this is referring to the index of the reply block.

PR #977, opened 2026-05-28
santad: push network rules + settings to santanetd via combined XPC

mlw:
Lands the santad-side machinery for delivering server-pushed network flow rules and NetworkExtension settings to santanetd; the santanetd receiver and engine integration land separately in the santanetd repo.

- New SNTNetworkExtensionConfig (Source/common/ne/): NSSecureCoding composite of settings + networkFlowRules, used as the single payload for both the santanetd registration reply and the runtime push (nil rules == unchanged).
- SNDNetworkExtensionXPC collapses to one method updateNetworkExtensionConfig:; registration reply widened to return a config so santanetd applies settings and the full ruleset atomically on startup.
- SNTNetworkFlowDefaultAction enum on SNTNetworkExtensionSettings / SNTSyncNetworkExtensionSettings; preflight reads flow_default_action from the NetworkExtension config message.
- SNTConfigBundle gains a reconcileNetworkExtension marker set only by PostflightConfigBundle. SNTDaemonControlController.updateSyncSettings: triggers SNTNetworkExtensionQueue reconcileNetworkExtensionConfig when set (after the performSyncStateBatch commit), mirroring the CEL post-commit pattern.
- Reconcile is the single push path: detects deltas against last-pushed state the queue owns, sends settings always + rules only on hash change, or no message at all if neither changed. lastPushed is recorded only on a successful reply so failures retry naturally on the next reconcile. Body and the last-pushed writes in clearNetworkExtensionConnection and handleRegistration are wrapped in @synchronized(self).
- Atomic rules + hash read via SNTNetworkFlowRulesSnapshot returned from SNTRuleTable.retrieveAllNetworkFlowRulesSnapshot — single inDatabase: block, one iteration builds the array and computes the hash from the same blob bytes. Documents the rule-hash cache invariant (writes only inside DB blocks). retrieveAllNetworkFlowRules removed (sole caller migrated).
- Drop the redundant syncNetworkExtensionSettings KVO watcher; the syncBaseURL revoke watcher now triggers reconcile directly after clearing.
- Index-based FMDB column access for execution + file access hash compute loops (matches the network flow hash compute pattern).
- Tests added: SNTNetworkExtensionConfigTest (round-trip incl. nil rules), SNTNetworkExtensionQueueTest (reconcile branches, registration seed, success/failure last-pushed semantics), SNTRuleTable snapshot tests, marker round-trip + getter coverage in SNTConfigBundle/SNTSyncConfigBundle tests.

### DIFF
--- a/Source/common/ne/SNTNetworkExtensionSettings.h
+++ b/Source/common/ne/SNTNetworkExtensionSettings.h
@@ -14,9 +14,10 @@
 
 #import <Foundation/Foundation.h>
 
-/// Action applied to network flows that match no NetworkFlowRule.
+/// Action applied to network flows that match no NetworkFlowRule. Values mirror
+/// pbv2::NetworkFlowDefaultAction (see NetworkFlowDefaultActionFromProto in SNTSyncPreflight).
 typedef NS_ENUM(NSInteger, SNTNetworkFlowDefaultAction) {
-  SNTNetworkFlowDefaultActionUnspecified = 0,  // treated as Allow
+  SNTNetworkFlowDefaultActionUnspecified = 0,
   SNTNetworkFlowDefaultActionAllow = 1,
   SNTNetworkFlowDefaultActionDeny = 2,
 };
--- a/Source/common/ne/SNTNetworkExtensionSettings.h
+++ b/Source/common/ne/SNTNetworkExtensionSettings.h
@@ -14,6 +14,13 @@
 
 #import <Foundation/Foundation.h>
 
+/// Action applied to network flows that match no NetworkFlowRule.
+typedef NS_ENUM(NSInteger, SNTNetworkFlowDefaultAction) {
+  SNTNetworkFlowDefaultActionUnspecified = 0,  // treated as Allow
+  SNTNetworkFlowDefaultActionAllow = 1,
+  SNTNetworkFlowDefaultActionDeny = 2,
+};
+
 /// Settings passed from the daemon (santa) to the network extension (santanetd) over XPC.
 ///
 /// This class conforms to NSSecureCoding, allowing it to be passed directly as a typed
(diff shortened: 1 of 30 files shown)


Source/common/ne/SNTNetworkExtensionSettings.h
```diff
@@ -14,6 +14,13 @@
 
 #import <Foundation/Foundation.h>
 
+/// Action applied to network flows that match no NetworkFlowRule.
+typedef NS_ENUM(NSInteger, SNTNetworkFlowDefaultAction) {
```
russellhancox, line 19: Could we re-use the `NetworkFlowDefaultAction` enum from the proto?

Source/santad/SNTNetworkExtensionQueue.mm
```diff
@@ -94,43 +92,74 @@ - (instancetype)initWithNotifierQueue:(SNTNotificationQueue*)notifierQueue
...
+    SNTNetworkExtensionSettings* settings =
+        [self generateSettingsForProtocolVersion:self.connectedProtocolVersion];
+    if (!settings) {
+      LOGW(@"Failed to generate settings for protocol version %@", self.connectedProtocolVersion);
+      return;
+    }
+
+    // Fast-path hash check: cheap (cached) and avoids materializing the ruleset when nothing
+    // changed.
+    NSString* currentHash = [[self.ruleTable hashOfHashes] networkFlowRulesHash];
```
russellhancox: Could we export `networkFlowRulesHash` to avoid retrieving all 3 hashes only to use one?

russellhancox (dismissed):

Source/common/ne/SNTNetworkExtensionSettings.h
mlw: We could but I'd rather not. It couples the XPC wire format with the proto wire format and I'd like to keep the flexibility. There'd also be a minor annoyance with the ENCODE/DECODE macros that we could get around, but would make this wrapper object look different than the others we have that perform similar functions.

Source/santad/SNTNetworkExtensionQueue.mm
mlw: Yes.  also combined the rules hash impl between `networkFlowRulesHash:` and `retrieveAllNetworkFlowRulesSnapshot`

russellhancox (approved):

PR #981, opened 2026-06-01
santad: Make sandbox executions allow transitive executions of itself

russellhancox:


Source/santad/SNTExecutionController.mm
```diff
@@ -131,6 +154,7 @@ - (instancetype)initWithRuleTable:(SNTRuleTable*)ruleTable
     _policyProcessor = policyProcessor;
     _procSignalCache = std::make_unique<SantaCache<std::pair<pid_t, int>, bool>>(100000);
     _touchIDApprovalCache = std::make_unique<SantaCache<std::string, uint64_t>>(100);
+    _sandboxedSeatbeltProcs = std::make_unique<SantaCache<std::pair<pid_t, int>, bool>>(20000);
```
sharvilshah: out of curiosity: what happens when `_sanboxedSeatbeltProcs` cache reaches its capacity, and also why not make it `100000` like the `_procSignalCache`

sharvilshah (dismissed):

Source/santad/SNTExecutionController.mm
mlw: Now that the authorizer is a tree-aware client, we could vend EXIT events thru to remove entries from this cache.

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.h
```diff
@@ -17,24 +17,35 @@
...
+/// This is a tree-aware client so that, by ES's within-client ordering
+/// guarantees, the process tree is kept consistent for the lineage of every
+/// process being authorized: a process's NOTIFY_FORK is folded into the tree
+/// before its AUTH_EXEC is processed here, and the instigator/target nodes are
+/// retained for the duration of authorization. SNTExecutionController relies on
+/// this for its CEL activations and for the seatbelt self-exec relaxation, both
+/// of which walk the tree's ancestry of the exec'ing (already-forked) process.
 @interface SNTEndpointSecurityAuthorizer
-    : SNTEndpointSecurityClient <SNTEndpointSecurityEventHandler>
+    : SNTEndpointSecurityTreeAwareClient <SNTEndpointSecurityEventHandler>
```
mlw, line 39: Is is possible to stop making the Recorder a tree aware client now?

Source/santad/SNTExecutionController.mm
```diff
@@ -93,6 +95,19 @@ static SNTEventState BlockToAllowDecision(SNTEventState blockDecision) {
...
+// Returns true if two processes execute the same binary, using the same two
+// verification modes as the seatbelt expectation check below: strict cdhash
+// comparison when both are CdhashStrictlyEnforced (the kernel guarantees the
+// cdhash binds to executed content), otherwise a (dev, ino) vnode comparison.
+static bool SameBinary(const es_process_t* a, const es_process_t* b) {
+  if (santa::CdhashStrictlyEnforced(a->codesigning_flags) &&
+      santa::CdhashStrictlyEnforced(b->codesigning_flags)) {
+    return memcmp(a->cdhash, b->cdhash, CS_CDHASH_LEN) == 0;
+  }
+  return a->executable->stat.st_dev == b->executable->stat.st_dev &&
```
mlw, line 115: We should have hash info for both the source and target at this point (from the SNTDecisionCache for the source - tho we may have to "hydrate" if cache was flushed). Would make for a stronger comparison here

Source/santad/EventProviders/SNTEndpointSecurityAuthorizer.h
russellhancox: As discussed, no.

Source/santad/SNTExecutionController.mm
russellhancox: Further child executions would fail unless run under `santactl sb`. I've upped the limit to make this much less likely.

sharvilshah (approved):
LGTM

PR #982, opened 2026-06-01
syncservice: Update reachability to avoid retry loop

russellhancox:
#970 made startReachability reset `_reachable = NO` on every arm. Since nw_path_monitor reports the current path state once on start, an online host's first update (satisfied) looked like a down→up transition and triggered a sync → preflight fails → re-arm → repeat, with no backoff.

Fix: track the monitor's first update as a baseline (hasInitialPathState) and only trigger a sync on a genuine down→up transition. An already-online host records its baseline and stops; the full-sync timer handles retries. Also folded the trigger logic into handlePathReachable: and dropped the side-effecting setReachable: setter.

Fixes SNT-436


pmarkowsky (approved):
LGTM.

PR #984, opened 2026-06-02
santad: fix broken and over-firing config KVO watchers

mlw:
Several SNTConfigurator KVO watchers handled state changes incorrectly:

- allowedPathRegex/blockedPathRegex never fired on config reload because their keyPathsForValuesAffecting methods kept the pre-rename allowlist/blocklist names, so runtime changes never flushed the cache.
- enableSilentTTYMode, exportMetrics, and metricExportInterval had no keyPathsForValuesAffecting method at all, so changes took effect only after a daemon restart.
- staticRules and metricExportInterval lacked a no-op guard, flushing caches / restarting the export timer on every unrelated config reload.
- celFallbackRules recompiled all CEL fallback rules on every syncState change; its watcher now skips no-op updates. This requires SNTCELFallbackRule value equality, which it was missing.

Add SNTConfigurator and SNTCELFallbackRule tests covering the dependency wiring and value equality.


Source/common/SNTConfiguratorTest.mm
```diff
@@ -453,4 +474,54 @@ - (void)testClearSyncStateBypassesSyncStateAccessAuthorizer {
...
+  for (NSString* keyPath in keyPaths) {
+    [cfg removeObserver:observer forKeyPath:keyPath context:NULL];
+  }
+
+  for (NSString* keyPath in keyPaths) {
+    XCTAssertEqual(
+        [observer.firedKeyPaths countForObject:keyPath], (NSUInteger)1,
+        @"configState change must notify observers of '%@' (got %lu). Check "
+        @"keyPathsForValuesAffecting<Key> in SNTConfigurator.mm.",
+        keyPath, (unsigned long)[observer.firedKeyPaths countForObject:keyPath]);
```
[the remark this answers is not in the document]
mlw: Fixed

pmarkowsky (dismissed):
LGTM.

Source/common/SNTCELFallbackRule.mm
```diff
@@ -37,6 +37,24 @@ - (instancetype)initWithCELExpr:(NSString*)celExpr
   return self;
 }
 
+- (BOOL)isEqual:(id)other {
+  if (other == self) return YES;
+  if (![other isKindOfClass:[SNTCELFallbackRule class]]) return NO;
+  SNTCELFallbackRule* o = other;
+  return ((self.celExpr == o.celExpr || [self.celExpr isEqual:o.celExpr]) &&
+          (self.customMsg == o.customMsg || [self.customMsg isEqual:o.customMsg]) &&
+          (self.customURL == o.customURL || [self.customURL isEqual:o.customURL]));
```
russellhancox: FYI the pointer equality checks aren't necessary, NSString's isEqualToString: (which isEqual: calls) has a pointer equality short-circuit.

russellhancox (dismissed):

Source/common/SNTCELFallbackRule.mm
mlw: Kept them for customMsg/customURL since if both sides were `nil`, we'd return `NO`, but they're optional fields.

russellhancox (approved):

PR #991, opened 2026-06-03
Add running process force-kill rule action

hydazz:
## Summary
- add a `running_process_action` execution-rule field with unset/default, `NONE`, and `FORCE_KILL` values
- persist the field in the rules DB while keeping existing hashes unchanged for unset rules
- re-evaluate running processes after execution-rule and StaticRules updates, using the existing kill-command path for matching block/silent block rules marked `FORCE_KILL`
- add focused coverage for parsing, persistence, hashing, pidversion guarding, and rule-change triggers

## Notes
- This is intentionally the force-kill primitive only. It does not implement soft stop, grace periods, or camera/microphone/screen-sharing special handling.
- The pinned sync protobuf schema does not yet carry this field. A protos update is needed before sync servers can deliver it through RuleDownload. Static rules and local rule dictionaries can exercise the field now.

## Testing
- `git diff --check origin/main...HEAD`
- `git diff --name-only origin/main...HEAD -- '*.h' '*.m' '*.mm' '*.cc' | xargs xcrun clang-format --Werror --dry-run`
- `swift format lint -s -r /Users/ahyde/GitHub/hydazz/santa`
- `make test` fails locally because `bazel` is not installed

### DIFF
--- a/Source/common/SNTCommonEnums.h
+++ b/Source/common/SNTCommonEnums.h
@@ -83,6 +83,12 @@ typedef NS_ENUM(NSInteger, SNTRuleState) {
   SNTRuleStateSeatbelt = 11,
 };
 
+typedef NS_ENUM(NSInteger, SNTRuleRunningProcessAction) {
+  SNTRuleRunningProcessActionUnset = 0,
+  SNTRuleRunningProcessActionNone = 1,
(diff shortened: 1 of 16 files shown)


hydazz:
Closing for now; this needs local build/test validation before being proposed.

PR #996, opened 2026-06-05
compare hashes against previously-saved values instead of hard-coding them

lexi-nadia:


mlw:
Thank you for your contribution, we appreciate it!

The robot that normally checks & asks you to sign our CLA is on vacation - would you mind signing the CLA if you have not already: https://cla-assistant.io/northpolesec/santa, then respond back here once it's signed so that we can review.

lexi-nadia:
Signed :)

mlw:
I'm going to pass on this one. The hard-coded hashes are intentional. santad sends these values to the sync server, so the rule-hash computation needs to stay stable across versions and machines. The literals act as a regression guard: if a change alters what feeds into the hash, the test fails and we notice. Comparing against previously-saved values removes that guard, since the test then stays green even if the computation changes (it only checks the length and that the value moves after a remove).

The transitive/remove relational checks are a reasonable idea, but I'd want them alongside the golden values rather than in place of them.

Appreciate you taking a look.

lexi-nadia:
Got it, thanks!

For context, the problem I was trying to solve is that the hashes seem to change depending on whether `-fobjc-constant-literals` is passed. Just FYI in case you run into this later. :)

mlw:
@lexi-nadia This is interesting info!

I dug into this a bit and was able to repro. The root cause is how FAA rules are stored in the local database as serialized bytes via `NSKeyedArchiver`. When we build the FAA rules hash, the digest is updated directly from those serialized bytes. Apple makes no guarantee of a stable binary layout for `NSKeyedArchiver`, and the byte order depends on the dictionary's key ordering and internal layout. That matches what you saw. The triggers are the clang flags `-fno-constant-nsdictionary-literals` and `-fno-constant-nsarray-literals`, the behavior behind the `-fobjc-constant-literals` flag you mentioned.

The purpose of these hashes is to inform the sync server of the rule state so it can decide whether a clean sync is necessary. Two bits of good news. First, execution rules are unaffected. Second, for a given Santa build on a given macOS version the values are stable. At runtime the rule dictionaries are built dynamically from the sync payload, so the realistic drift vector is an OS version that changes `NSKeyedArchiver`'s layout. Worst case depends on what the sync server does with it, but it's likely just one extra clean sync, after which the values are once again internally consistent.

So while this can manifest at runtime, the fallout is pretty isolated. It'd still be good to fix. I think the right approach is a minor schema migration to add a column that stores a hash computed over a canonical representation of the bytes at insert time. Computing the rules hash then only has to fold in the name and precomputed hash fields, with no dependence on non-canonical representations.

Thank you for the extra details here! I've filed a ticket for us to track this. It's a fair bit more involved than the test change (schema migration plus a backfill), so no pressure at all, but you're very welcome to take a stab if you're interested. Otherwise we'll get to it soon.

PR #998, opened 2026-06-08
Proactively Lockdown permissions on labeler.yml

pmarkowsky:
This locks down the permissions on the labeler workflow.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,11 @@
 name: Process PRs
 
 on:
-  - pull_request_target
+  - pull_request:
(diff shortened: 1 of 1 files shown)


russellhancox (dismissed):

.github/workflows/labeler.yml
```diff
@@ -1,7 +1,11 @@
 name: Process PRs
 
 on:
-  - pull_request_target
+  - pull_request:
```
russellhancox:
```suggestion
  pull_request:
```

russellhancox (approved):

PR #1014, opened 2026-06-18
santad: Add `today()` and `days()` functions to CELv2 context

russellhancox:


Source/common/cel/RelativeTimeFunction.mm
```diff
@@ -0,0 +1,103 @@
...
+namespace cel {
+
+namespace {
+constexpr int64_t kSecondsPerDay = 24 * 60 * 60;
+
+// days(n) -> duration. Pure: n*24h. Sugar because CEL's native duration() only
+// parses units up to hours.
+absl::Duration DaysImpl(int64_t n, const google::protobuf::DescriptorPool*,
+                        google::protobuf::MessageFactory*, google::protobuf::Arena*) {
+  return absl::Hours(24) * n;
```
pmarkowsky, line 45: nit: do we need to clamp or otherwise check for wrapping here?

pmarkowsky (approved):
LGTM.

Source/common/cel/RelativeTimeFunction.mm
russellhancox: No, absl::Duration multiplication saturates and CEL timestamp comparisons range-check and anything out-of-range returns a sensible error (which would fail-closed in rules)

PR #1020, opened 2026-06-22
santad: Evict oldest spool files when full & add spool metrics

russellhancox:
When the telemetry spool reaches its size limit, erase the oldest spool files to make room rather than dropping all new writes (which also starved downstream processing). Raise the default spool size limit from 100MB to 250MB.

Also move from directory-walking and calling `stat()` once per-file when calculating the directory size and files to evict into batched calls to `getattrlistbulk()`, which dramatically reduces the number of syscalls when the spool is full (with no negative impact when the spool has few files in).

Add two metrics:
  * `/santa/spool/eviction_count` - spool files erased due to the size limit
  * `/santa/spool/size_bytes`     - current on-disk spool directory size

Part of WS-787

### DIFF
--- a/Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool.h
+++ b/Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool.h
@@ -125,8 +125,16 @@ class FsSpoolWriter {
       // Over the limit: rather than dropping all new telemetry, erase the
       // oldest spool files to make room. This keeps the freshest data flowing
       // (and lets signal processing continue) at the cost of the oldest,
-      // not-yet-exported files.
+      // not-yet-exported files. Updates spool_size_estimate_ to the
+      // post-eviction total.
       EvictOldestSpoolFiles(files, *estimate);
+
+      // If eviction couldn't free enough space (e.g. unlinks failing on a
+      // read-only remount or immutable files), keep the spool bounded by
+      // refusing the write rather than letting it grow without limit.
+      if (spool_size_estimate_ > max_spool_size_) {
+        return absl::ResourceExhaustedError("No space left in spool directory");
+      }
     }
 
     return absl::OkStatus();
@@ -279,14 +287,14 @@ class FsSpoolWriter {
   }
 
   // Deletes the oldest spool files until the estimated on-disk size of the
-  // spool directory drops below the high-water mark (90% of max_spool_size_).
-  // Evicting to a high-water mark rather than to exactly the limit frees ~10%
+  // spool directory drops below the low-water mark (90% of max_spool_size_).
+  // Evicting to a low-water mark rather than to exactly the limit frees ~10%
   // of headroom so eviction doesn't immediately recur on the next few writes.
   // `files`/`total` come from the caller's directory walk, so no second walk is
   // needed. Updates spool_size_estimate_ to the post-eviction total.
   // Best-effort: a file that fails to unlink is left in place and skipped.
   void EvictOldestSpoolFiles(std::vector<DirEntryInfo>& files, size_t total) {
-    const size_t high_water_mark = max_spool_size_ - max_spool_size_ / 10;
+    const size_t low_water_mark = max_spool_size_ - max_spool_size_ / 10;
 
     // Oldest first.
     std::sort(files.begin(), files.end(),
@@ -296,7 +304,7 @@ class FsSpoolWriter {
 
     size_t evicted = 0;
     for (const DirEntryInfo& file : files) {
-      if (total <= high_water_mark) {
+      if (total <= low_water_mark) {
         break;
       }
       if (Unlink(file.path.c_str()) == 0) {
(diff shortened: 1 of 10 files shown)


Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool.h
```diff
@@ -254,25 +269,46 @@ class FsSpoolWriter {
...
+
+  // Deletes the oldest spool files until the estimated on-disk size of the
+  // spool directory drops below the high-water mark (90% of max_spool_size_).
+  // Evicting to a high-water mark rather than to exactly the limit frees ~10%
+  // of headroom so eviction doesn't immediately recur on the next few writes.
+  // `files`/`total` come from the caller's directory walk, so no second walk is
+  // needed. Updates spool_size_estimate_ to the post-eviction total.
+  // Best-effort: a file that fails to unlink is left in place and skipped.
+  void EvictOldestSpoolFiles(std::vector<DirEntryInfo>& files, size_t total) {
+    const size_t high_water_mark = max_spool_size_ - max_spool_size_ / 10;
```
mlw: nit: this is a low water mark

Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool_nowindows.cc
```diff
@@ -124,41 +129,95 @@ absl::Status IterateDirectory(
...
+  memset(&attr_list, 0, sizeof(attr_list));
+  attr_list.bitmapcount = ATTR_BIT_MAP_COUNT;
+  attr_list.commonattr = ATTR_CMN_RETURNED_ATTRS | ATTR_CMN_NAME |
+                         ATTR_CMN_OBJTYPE | ATTR_CMN_MODTIME;
+  attr_list.fileattr = ATTR_FILE_ALLOCSIZE;
+
+  size_t total = 0;
+  // 64KiB holds ~580 entries/call, so even a 15k-file spool is a couple dozen
+  // syscalls rather than one stat() per file.
+  char buf[64 * 1024];
```
mlw: While probably not a major issue, this is a pretty massive stack buffer. We could do something like:
```
  auto buf = std::make_unique_for_overwrite<char[]>(64 * 1024);
  // ... getattrlistbulk(fd, &attr_list, buf.get(), 64 * 1024, 0) ...
```
This allocates a buffer and leaves it uninitialized. (We also have now `santa::UninitBuffer`, but this lives in VerifyingHasher. It should probably be moved up to `common/`, not sure i'd reach for it right now...)

Source/santad/Logs/EndpointSecurity/Writers/FSSpool/fsspool.h
```diff
@@ -98,23 +107,29 @@ class FsSpoolWriter {
...
+
+    if (spool_size_estimate_ > max_spool_size_) {
+      // Over the limit: rather than dropping all new telemetry, erase the
+      // oldest spool files to make room. This keeps the freshest data flowing
+      // (and lets signal processing continue) at the cost of the oldest,
+      // not-yet-exported files.
+      EvictOldestSpoolFiles(files, *estimate);
+    }
+
+    return absl::OkStatus();
```
mlw, line 140: From the robut:

With eviction in place, `SpaceAvailable()` now only returns non-OK on a scan failure; it can never return `ResourceExhausted` again. Two consequences worth deciding on explicitly, since they're linked:

**Dead code.** Because `ResourceExhausted` no longer propagates out of `InitializeCurrentSpoolStateIfNeeded()`, this is all now unreachable:
- `Write()` (lines 206-216): `space_check_failure_since_last_flush_` is pinned to `false`, so the `IsResourceExhausted` set and the `DataLossError("No space exists")` early-out never fire.
- `Spool::Write()` (`Spool.h:136`): the `absl::IsDataLoss` branch is dead as a result.
- The `Write()` doc comment (196-200) and the member comment (331-337) describe behavior that no longer happens.

It's logically dead rather than unused, so the compiler won't flag it, which makes it a trap for the next person debugging telemetry drops.

**Soft cap.** Eviction only decrements `total` on a successful `Unlink`, and we return OK regardless of the result. So if unlinks persistently fail (read-only remount / `EROFS`, immutable flags, FS errors), the spool grows past `max_spool_size_` with no backstop, whereas the old path refused writes and stayed bounded. Unlikely since we run as root and own the dir, but it's exactly the disk-pressure tail where overgrowth is worst.

Two ways out, and they fix different amounts:
1. **Delete** the dead machinery above + correct the comments, and accept the soft cap as out of scope.
2. **Rewire**: return `ResourceExhausted` here when `spool_size_estimate_ > max_spool_size_` *after* eviction. That restores the hard bound for the can't-free-space case and gives the existing `space_check_failure`/`DataLoss` code a real trigger again, so it stops being dead.

I lean toward (2): it's a few lines and makes the existing code honest. (1) is fine if we'd rather keep it minimal. Either way, let's not leave guard code that reads as live but can't execute.

mlw (changes requested):

mlw (approved):

PR #969, opened 2026-05-26
santad: persist network flow rules in SNTRuleTable

mlw:
## Summary
- Adds a third sub-table `network_flow_rules` (migration v12 → v13) holding raw `NetworkFlowRule.Add` proto blobs keyed by `rule_id`.
- Extends `addExecutionRules:fileAccessRules:…` to also take `networkFlowRules:`, so all three rule types from a single `RuleDownloadResponse` apply in one transaction. The 3-arg legacy wrapper is preserved for the compiler/standalone callers.
- Adds `networkFlowRulesHash` to `SNTRuleTableRulesHash`, populated via `hashOfHashes`. Xxhash128 over the blobs in `rule_id ASC` order — `rule_id` is field 1 of the proto so it's not hashed separately.
- Retrieve and hash loops use indexed-column FMDB access; enterprise blocklists are expected to reach 100k+ rules.

## Dependencies
**Stacked on #965** (`mlw/network-rules-common`, currently in review). This PR's base is that branch — please merge #965 first; GitHub will retarget this PR to `main` automatically. The two are intended to stand independently in review.

XPC delivery to `santanetd` and the corresponding `SNTDaemonControlController` entry are out of scope here and will follow in a later PR.

## Test plan
- [x] `bazel test //Source/santad:SNTRuleTableTest` — 16 new tests covering persist / upsert / remove / remove-nonexistent / CleanAll / CleanNonTransitive / CleanupExecutionRules-preserves / empty-array rejection / network-only non-empty guard / hash stability / hex length / blob sensitivity / order independence / empty-table.
- [x] `bazel test //Source/santad:SNTDaemonControlControllerTest //Source/santad:SNTCompilerControllerTest //Source/santad:SNTExecutionControllerTest` — existing tests still pass with the 5-arg signature.
- [x] `bazel build //Source/santad:Santad`
- [x] `Testing/check_test_suites.sh` clean.

PR #1022, opened 2026-06-22
Use `name` as the primary key for network flow rules

tburgin:
## Use `name` as the primary key for network flow rules

  Network flow rules now key on the sync server's `name` (the rule's stable slot
  identity) instead of the integer `rule_id`. `rule_id` is retained as the version
  field — it still rides on every Add and is still what gets reported when a rule
  trips.

  This matches the upstreamed proto change ([protos #130](https://github.com/northpolesec/protos/pull/130)):
  `NetworkFlowRule.Add` now carries `name` (field 1) + `rule_id` (field 2), and
  `Remove` carries only `name`.

  ### Notes

  - Because `rule_id` carries a `UNIQUE` constraint, an Add reusing an existing
    `rule_id` under a different name evicts the prior row via `INSERT OR REPLACE`.
    A correct server never reuses a `rule_id` across names; documented in-code.
  - Name length/pattern validation (≤63 bytes, C-identifier) lives in santanetd's
    `ValidateNetworkFlowRule`.

### DIFF
--- a/Source/common/SNTNetworkFlowRule.mm
+++ b/Source/common/SNTNetworkFlowRule.mm
@@ -58,7 +58,10 @@
   if (ruleName.length == 0) {
     return nil;
   }
-  return [self initWithState:SNTNetworkFlowRuleStateRemove ruleName:ruleName ruleId:0 protoBlob:nil];
+  return [self initWithState:SNTNetworkFlowRuleStateRemove
+                    ruleName:ruleName
+                      ruleId:0
+                   protoBlob:nil];
 }
 
 + (BOOL)supportsSecureCoding {
--- a/Source/common/SNTNetworkFlowRule.mm
+++ b/Source/common/SNTNetworkFlowRule.mm
@@ -17,6 +17,7 @@
 #import "Source/common/CoderMacros.h"
 
 @interface SNTNetworkFlowRule ()
+@property(readwrite, copy) NSString* ruleName;
 @property(readwrite) int64_t ruleId;
 @property(readwrite) SNTNetworkFlowRuleState state;
 @property(readwrite, copy) NSData* protoBlob;
@@ -25,6 +26,7 @@
 @implementation SNTNetworkFlowRule
 
 - (instancetype)initWithState:(SNTNetworkFlowRuleState)state
+                     ruleName:(NSString*)ruleName
                        ruleId:(int64_t)ruleId
                     protoBlob:(NSData*)protoBlob {
   if (state != SNTNetworkFlowRuleStateAdd && state != SNTNetworkFlowRuleStateRemove) {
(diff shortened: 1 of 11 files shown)


Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -339,8 +339,13 @@ - (uint32_t)initializeDatabase:(FMDatabase*)db fromVersion:(uint32_t)version {
...
+    // `rule_id` is the rule's version; it is kept UNIQUE because santanetd still
+    // uses it as the unique in-memory identity / precedence tiebreak, reading it
+    // out of the serialized Add blob.
     [db executeUpdate:@"CREATE TABLE 'network_flow_rules' ("
-                      @"'rule_id' INTEGER PRIMARY KEY, "
+                      @"'name' TEXT NOT NULL PRIMARY KEY, "
+                      @"'rule_id' INTEGER NOT NULL UNIQUE, "
                       @"'rule_blob' BLOB NOT NULL"
                       @")"];
     newVersion = 13;
```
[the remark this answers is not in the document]
tburgin: Nope, we have not shipped this yet. Using the same version is fine. Devs will need to clear their db or migrate manually but that is fine.

Source/common/SNTNetworkFlowRule.mm
```diff
@@ -33,28 +35,41 @@ - (instancetype)initWithState:(SNTNetworkFlowRuleState)state
...
+                             ruleId:(int64_t)ruleId
+                          protoBlob:(NSData*)protoBlob {
+  if (ruleName.length == 0 || !protoBlob) {
     return nil;
   }
-  return [self initWithState:SNTNetworkFlowRuleStateAdd ruleId:ruleId protoBlob:protoBlob];
+  return [self initWithState:SNTNetworkFlowRuleStateAdd
+                    ruleName:ruleName
+                      ruleId:ruleId
+                   protoBlob:protoBlob];
```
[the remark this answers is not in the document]
tburgin: The validation is done by the rule downloader.

tburgin:
This is ready for review.

Source/santasyncservice/SNTSyncRuleDownload.mm
```diff
@@ -335,10 +344,18 @@ SNTRuleCleanup SyncTypeToRuleCleanup(SNTSyncType syncType) {
       std::string serialized;
       add.SerializeToString(&serialized);
       NSData* blob = [NSData dataWithBytes:serialized.data() length:serialized.size()];
-      return [[SNTNetworkFlowRule alloc] initAddRuleWithId:add.rule_id() protoBlob:blob];
+      return [[SNTNetworkFlowRule alloc] initAddRuleWithName:StringToNSString(add.name())
+                                                      ruleId:add.rule_id()
+                                                   protoBlob:blob];
+    }
+    case ::pbv2::NetworkFlowRule::kRemove: {
+      const std::string& name = nr.remove().name();
```
mlw, line 352: We should define the validation we want to perform on names

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -339,8 +339,13 @@ - (uint32_t)initializeDatabase:(FMDatabase*)db fromVersion:(uint32_t)version {
   }
 
   if (version < 13) {
+    // `name` is the primary key (the sync server's identity for the rule).
+    // `rule_id` is the rule's version; it is kept UNIQUE because santanetd still
+    // uses it as the unique in-memory identity / precedence tiebreak, reading it
+    // out of the serialized Add blob.
     [db executeUpdate:@"CREATE TABLE 'network_flow_rules' ("
-                      @"'rule_id' INTEGER PRIMARY KEY, "
+                      @"'name' TEXT NOT NULL PRIMARY KEY, "
```
mlw, line 360: Saw this was flagged by CR - but we will need to bump versions here. The table creation landed in #969 which was part of 2026.5.

mlw (changes requested):

Source/santad/DataLayer/SNTRuleTable.mm
tburgin: Ah, thanks for pointing that out. Done.

Source/santasyncservice/SNTSyncRuleDownload.mm
tburgin: Currently I have validation being done in santanetd's ValidateNetworkFlowRule. Seems like that is the canonical place, but do you want to move that to Santa?

mlw: No, that's fine. Think I was just calling this out because it's mentioned in the PR description but the work isn't done in this PR.

mlw (approved):

PR #1023, opened 2026-06-23
test: Significantly speed-up some slower tests

russellhancox:
* SNTSyncTest: >60s -> 0.2s. Stub `updateSyncSettings:reply:` to bypass the semaphore timeout, allow overriding the XSRF retry backoff
* SantadTest: ~60s -> 1.5s. Wrap the deadline's Message copy in a `std::optional<Message>` shared by both blocks so that the handler can reset it once it wins the `processingSema` race. This speeds up the test (because the message is not kept alive for the full 1.5s after the handler responds) and also releases the message faster in normal execution
* SNTRuleTableTest: ~15s -> 1.2s. Every tests eagerly hashed + codesign-validated all critical system binaries even though most tests never used it. Made `criticalSystemBinaries` lazy using dispatch_once.

### DIFF
--- a/Source/common/es/SNTEndpointSecurityClient.mm
+++ b/Source/common/es/SNTEndpointSecurityClient.mm
@@ -23,6 +23,8 @@
 #include <sys/qos.h>
 
 #include <algorithm>
+#include <memory>
+#include <optional>
 #include <set>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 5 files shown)


tburgin (dismissed):

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -364,8 +375,8 @@ - (uint32_t)initializeDatabase:(FMDatabase*)db fromVersion:(uint32_t)version {
     LOGE(@"Failed to create CELv2 evaluator: %s", std::string(celEval.status().message()).c_str());
   }
 
-  // Setup critical system binaries
-  [self setupSystemCriticalBinaries];
+  // Critical system binaries (hashing + codesigning real binaries) are computed
+  // lazily on first access via the criticalSystemBinaries getter, keeping init cheap.
```
mlw, line 379: Right now this first access would be on the hot/sensitive AUTH EXEC path. We should prime this call immediately after initialization in SantadDeps.mm. Seems like the test will still trigger this, but should only happen once since protected with the dispatch once.

Source/common/es/SNTEndpointSecurityClient.mm
```diff
@@ -338,15 +340,20 @@ - (void)processMessage:(Message&&)msg handler:(void (^)(Message))messageHandler
   int64_t processingBudget = [self computeBudgetForDeadline:msg->deadline
                                                 currentTime:mach_absolute_time()];
 
-  // Copy the original message and then move into the auto-responder block
-  __block Message tmpDeadlineMsg = msg;
+  // Copy the original message into a heap holder shared with the auto-responder
+  // block. The holder lets the handler release the copy as soon as it wins the
+  // race (see below) instead of keeping the Message (and its underlying
+  // es_message_t) alive until the budget elapses.
+  auto deadlineMsgHolder = std::make_shared<std::optional<Message>>(msg);
```
mlw, line 347: This is great!

mlw (changes requested):

Source/santad/DataLayer/SNTRuleTable.mm
russellhancox: Good call, done

mlw (approved):

PR #1026, opened 2026-06-25
santad: Store signal reports and upload them to the sync server

russellhancox:
Add the SignalScanner that runs a Sleigh signal scan over each closed spool file, parses the SleighSignalScanResponse, and stores each SignalReport in a new signal_reports table in the events database. During sync (v2 only), pending reports are uploaded to the /signalupload handler, and an immediate upload is attempted after each scan with fallback to the next sync on failure.


Source/santad/SignalScanner.mm
```diff
@@ -0,0 +1,108 @@
...
+    NSData* data = signal.data;
+    // Guard against a nil/empty data: std::string(nullptr, 0) is undefined, and an empty
+    // signal is useless to Sleigh anyway.
+    if (data.length == 0) {
+      continue;
+    }
+    v.emplace_back(static_cast<const char*>(data.bytes), data.length);
+  }
+
+  absl::MutexLock lock(&lock_);
```
mlw: nit: Use the new constructor over the deprecated one.

Source/santad/SignalScanner.mm
```diff
@@ -0,0 +1,108 @@
...
+  signals_ = std::move(v);
+}
+
+void SignalScanner::ScanFile(std::string path) {
+  auto shared_this = shared_from_this();
+  dispatch_async(scan_q_, ^{
+    std::vector<std::string> signals;
+    {
+      absl::MutexLock lock(&shared_this->lock_);
```
mlw: nit: Use the new constructor over the deprecated one.

Source/santad/SantadDeps.mm
```diff
@@ -155,13 +159,41 @@
...
+      SleighLauncher::Create(std::string(SleighLauncher::kDefaultSleighPath)),
+      [configurator telemetryExportTimeoutSec], ^(NSArray<SNTStoredSignalReport*>* reports) {
+        // Persist first so the reports survive an upload failure (and a later sync re-tries).
+        // Storage deduplicates repeated firings of the same signal; only the stored subset is
+        // uploaded so a misconfigured signal can't flood the server with identical reports.
+        reports = [[SNTDatabaseController eventTable] addStoredSignalReports:reports];
+        if (reports.count == 0) return;
+        // Attempt an immediate out-of-band upload. The sync service deletes the reports on
+        // success and otherwise leaves them for the next sync.
+        MOLXPCConnection* conn = [SNTXPCSyncServiceInterface configuredConnection];
```
mlw: Could we re-use `syncd_queue` here? Would be a way to not have to create a new XPC connection per finding.

mlw (approved):

PR #1029, opened 2026-06-25
common: Add SNTStoredNetworkFlowEvent

mlw:
Adds `SNTStoredNetworkFlowEvent`, a stored-event model mirroring the v2 sync proto `NetworkFlowEvent`, with three proto-value-aligned `NS_ENUM`s (decision, direction, socket family) and a santad-local `silent` bit. NSSecureCoding round-trip plus uniqueID/unactionable coverage; builds standalone.

### DIFF
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -918,6 +918,27 @@ santa_unit_test(
 )
 
 objc_library(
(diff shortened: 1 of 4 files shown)


tburgin (approved):
love it

russellhancox (approved):

PR #1035, opened 2026-06-30
santad: extract shared TimedSyncSession base and refactor TMM onto it

sharvilshah:
Part of WS-1128. Stacked on #1034, ~review/merge that first~ which has been merged, so the base branch for this is `main`

- Extract a shared TimedSyncSession base class: timer, on-disk state
  persistence, and KVO lifecycle common to timed sync sessions.
- Refactor Temporary Monitor Mode onto it; drop its bespoke timer/state code.
- Replace TMM-specific SNTConfigurator accessors with generic timed-session
  accessors (persistTimedSessionState:forKey: / savedTimedSessionStateForKey:).
- No behavior change.

### DIFF
--- a/Source/santad/BUILD
+++ b/Source/santad/BUILD
@@ -316,6 +316,25 @@ santa_unit_test(
 )
 
 objc_library(
+    name = "TimedSyncSession",
+    srcs = ["TimedSyncSession.mm"],
+    hdrs = ["TimedSyncSession.h"],
+    sdk_dylibs = [
+        "EndpointSecurity",
+    ],
+    deps = [
+        ":SNTNotificationQueue",
+        "//Source/common:SNTConfigurator",
+        "//Source/common:SNTKVOManager",
+        "//Source/common:SNTLogging",
+        "//Source/common:SNTSystemInfo",
+        "//Source/common:SystemResources",
+        "//Source/common:Timer",
+        "@abseil-cpp//absl/synchronization",
+    ],
+)
+
+objc_library(
     name = "TemporaryMonitorMode",
     srcs = ["TemporaryMonitorMode.mm"],
     hdrs = ["TemporaryMonitorMode.h"],
@@ -324,16 +343,12 @@ objc_library(
     ],
     deps = [
         ":SNTNotificationQueue",
+        ":TimedSyncSession",
         "//Source/common:PassKey",
-        "//Source/common:Pinning",
         "//Source/common:SNTConfigurator",
         "//Source/common:SNTError",
-        "//Source/common:SNTKVOManager",
         "//Source/common:SNTModeTransition",
         "//Source/common:SNTStoredTemporaryMonitorModeAuditEvent",
-        "//Source/common:SNTSystemInfo",
-        "//Source/common:SystemResources",
-        "//Source/common:Timer",
         "@abseil-cpp//absl/synchronization",
     ],
 )
@@ -344,6 +359,7 @@ santa_unit_test(
     deps = [
         ":SNTNotificationQueue",
         ":TemporaryMonitorMode",
+        ":TimedSyncSession",
         "//Source/common:SNTModeTransition",
         "//Source/common:SystemResources",
         "@OCMock",
--- a/Source/santad/BUILD
+++ b/Source/santad/BUILD
@@ -347,6 +347,7 @@ objc_library(
         "//Source/common:PassKey",
         "//Source/common:SNTConfigurator",
         "//Source/common:SNTError",
+        "//Source/common:SNTLogging",
         "//Source/common:SNTModeTransition",
         "//Source/common:SNTStoredTemporaryMonitorModeAuditEvent",
         "@abseil-cpp//absl/synchronization",
--- a/Source/santad/TemporaryMonitorMode.mm
+++ b/Source/santad/TemporaryMonitorMode.mm
@@ -102,8 +102,10 @@ bool TemporaryMonitorMode::ApplyEffect(NSError** err) {
   return true;
 }
 
-void TemporaryMonitorMode::RevertEffect() {
+bool TemporaryMonitorMode::RevertEffect() {
+  // Clearing an in-memory flag cannot fail, so the revert always succeeds.
   [configurator_ setInTemporaryMonitorMode:NO];
+  return true;
 }
 
 bool TemporaryMonitorMode::ReapplyEffectOnRestart() {
--- a/Source/santad/TemporaryMonitorMode.mm
+++ b/Source/santad/TemporaryMonitorMode.mm
@@ -14,21 +14,13 @@
 
 #include "Source/santad/TemporaryMonitorMode.h"
 
-#include <mach/mach_time.h>
-
 #import "Source/common/MOLXPCConnection.h"
-#include "Source/common/Pinning.h"
 #import "Source/common/SNTConfigurator.h"
 #import "Source/common/SNTError.h"
+#import "Source/common/SNTLogging.h"
+#import "Source/common/SNTModeTransition.h"
 #import "Source/common/SNTStoredTemporaryMonitorModeAuditEvent.h"
-#import "Source/common/SNTSystemInfo.h"
 #import "Source/common/SNTXPCNotifierInterface.h"
-#include "Source/common/SystemResources.h"
-
-NSString* const kStateTempMonitorModeBootUUIDKey = @"BootUUID";
-NSString* const kStateTempMonitorModeDeadlineKey = @"Deadline";
-NSString* const kStateTempMonitorModeSavedSyncURLKey = @"SyncURL";
-NSString* const kStateTempMonitorModeSessionUUIDKey = @"SessionUUID";
 
 namespace santa {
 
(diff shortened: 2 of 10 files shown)


sharvilshah:
@coderabbitai full review

sharvilshah:
@coderabbitai full review

Source/santad/TimedSyncSession.mm
```diff
@@ -0,0 +1,349 @@
...
+  if (secs_remaining < kMinAllowedStateRemainingSeconds) {
+    // Not resumable. The reboot/expiry leave audit (and sync-change revoke policy)
+    // were already produced by GetSecondsRemainingFromStateLocked. Actively revert
+    // the effect (TMM: clear the flag, a no-op on a fresh boot; TAM: remove the
+    // user from the admin group), clear persisted + extra state, notify the GUI.
+    RevertEffect();
+    [configurator_ persistTimedSessionState:nil forKey:StateKey()];
+    NotifyLeave();
+    ClearExtraState();
+    return;
```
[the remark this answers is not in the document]
sharvilshah: resolved in [40de624](https://github.com/northpolesec/santa/pull/1035/commits/40de624a34b2f1e4d1a41ec3cdc61f8ebf512ec9)

Source/santad/TimedSyncSession.mm
```diff
@@ -0,0 +1,349 @@
...
+              if ((!newValue && !oldValue) ||
+                  ([newValue.absoluteString isEqualToString:oldValue.absoluteString])) {
+                return;
+              }
+              if (auto strong_self = weak_self.lock()) {
+                if (strong_self->Revoke(strong_self->LeaveReasonSyncServerChanged())) {
+                  LOGI(@"Timed sync session revoked due to SyncBaseURL changing.");
+                }
+              }
+            }] ];
```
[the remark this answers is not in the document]
sharvilshah: resolved in [40de624](https://github.com/northpolesec/santa/pull/1035/commits/40de624a34b2f1e4d1a41ec3cdc61f8ebf512ec9)

Source/santad/BUILD
```diff
@@ -324,16 +343,12 @@ objc_library(
...
-        "//Source/common:Pinning",
         "//Source/common:SNTConfigurator",
         "//Source/common:SNTError",
-        "//Source/common:SNTKVOManager",
         "//Source/common:SNTModeTransition",
         "//Source/common:SNTStoredTemporaryMonitorModeAuditEvent",
-        "//Source/common:SNTSystemInfo",
-        "//Source/common:SystemResources",
-        "//Source/common:Timer",
         "@abseil-cpp//absl/synchronization",
```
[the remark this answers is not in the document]
sharvilshah: resolved in [40de624](https://github.com/northpolesec/santa/pull/1035/commits/40de624a34b2f1e4d1a41ec3cdc61f8ebf512ec9)

Source/santad/TemporaryMonitorModeTest.mm
```diff
@@ -88,66 +90,196 @@ TemporaryMonitorModePeer tmm([SNTConfigurator configurator], self.mockNotQueue,
...
   testState = [goodTestState mutableCopy];
-  testState[kStateTempMonitorModeSavedSyncURLKey] = pinnedURL2.host;
+  testState[kTimedSessionSyncURLKey] = pinnedURL2.host;
   OCMExpect([self.mockConfigurator
       setSyncServerModeTransition:[OCMArg checkWithBlock:^BOOL(SNTModeTransition* mt) {
         return mt.type == SNTModeTransitionTypeRevoke;
       }]]);
-  XCTAssertEqual(tmm.GetSecondsRemainingFromInitialStateLocked(testState, testBootUUID, pinnedURL),
-                 0);
+  XCTAssertEqual(tmm.GetSecondsRemainingFromStateLocked(testState, testBootUUID, pinnedURL), 0);
```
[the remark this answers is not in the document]
sharvilshah: resolved in [40de624](https://github.com/northpolesec/santa/pull/1035/commits/40de624a34b2f1e4d1a41ec3cdc61f8ebf512ec9)

sharvilshah:
Heads up before you dig in: this looks bigger than it is. It's ~960 added / ~390 removed, but most of the removed lines from `TemporaryMonitorMode.mm` just moved into the new `TimedSyncSession.mm` — very little is actually new logic.

The point of the PR is to stop TMM from owning all its own plumbing. The timer, the persisted session state, the locking, and the whole sync/enter/expire/restart lifecycle get pulled out into a base class (`TimedSyncSession`), and TMM becomes a thin subclass that just fills in the parts that are specific to it. Nothing about how TMM behaves for a user changes — I mainly did this so Temporary Admin Mode (next PR) can sit on the same base instead of copy-pasting all of it.

Wherever you see a chunk deleted from TMM, it almost certainly reappears in `TimedSyncSession`.

If it helps, I'd read it in this order:

1. `TimedSyncSession.h` first — that's the whole design. The base owns the timer, the common state (session UUID + expiration), and the lifecycle (`SetupFromState` for restart recovery, `RequestMinutes`, `Revoke`, expiry), and everything feature-specific goes through the virtual hooks near the bottom (apply/revert the effect, build the audit events, the policy predicates, the enter/leave reason enums).
2. `TimedSyncSession.mm` — the moved lifecycle/timer/lock code.
3. `TemporaryMonitorMode.{h,mm}` — now just implements those hooks. This is where you'd confirm the TMM-specific behavior actually survived the move.
4. `SNTConfigurator` — the old TMM-only state accessors got generalized to keyed ones (persistTimedSessionState:forKey: / savedTimedSessionStateForKey:) so more than one feature can share them. The SNTDaemonControlController change is a one-liner from that.

**On testing**: I expanded `TemporaryMonitorModeTest` and the full suite is green. I also ran this branch's build against a real workshop on a local build end to end, synced the policy down, entered monitor mode, watched the enter/leave audit events upload, and confirmed it auto-reverted on the timer.

Source/santad/TimedSyncSession.mm
```diff
@@ -0,0 +1,379 @@
...
+    // token chain, so a push-token-only deployment can hold a session with a nil
+    // syncBaseURL. Watching syncBaseURL alone would never revoke such a session when
+    // its token chain is lost. Watch the chain too and revoke as soon as the gate is
+    // no longer satisfied.
+    [[SNTKVOManager alloc]
+        initWithObject:configurator_
+              selector:@selector(pushTokenChain)
+                  type:[NSArray class]
+              callback:^(NSArray* oldValue, NSArray* newValue) {
+                if (auto strong_self = weak_self.lock()) {
```
mlw, line 97: There is a potential deadlock here. This KVO handler will get fired on changes to anything in SNTConfigurator's `syncStateSet`. The call down below to `Revoke` can trigger this callback again, and can deadlock hitting the mutex in `Revoke`.

Minimally, we can break the cycle by ensuring 
```suggestion
                if ((!newValue && !oldValue) || [newValue isEqualToArray:oldValue]) {
                  return;
                }
                if (auto strong_self = weak_self.lock()) {
```

A more thorough fix would be to revoke asynchronously on some internal dispatch queue.

Source/santad/TimedSyncSession.mm
```diff
@@ -0,0 +1,379 @@
...
+  // fail-closed timeout so a hung GUI cannot hold grant_mutex_ indefinitely.
+  __block BOOL authenticated = NO;
+  __block NSString* reason = nil;
+  dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+  RequestAuthorization(^(BOOL ok, NSString* r) {
+    authenticated = ok;
+    reason = r;
+    dispatch_semaphore_signal(sema);
+  });
+  if (dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, kAuthTimeoutNSec)) != 0) {
```
mlw, line 247: `RequestAuthorization` wraps a call to `authorizeTemporaryMonitorMode`, which wraps a synchronous XPC call to the GUI.

If the XPC connection in `SNTNotificationQueue` (`self.notifierConnection`) is not setup (or maybe dies), the semaphore is never signaled and we hang for 90s. Previously this would've been an instant return.

We should look into making `authorizeTemporaryMonitorMode:` ensure the block gets called.

Source/santad/TemporaryMonitorMode.mm
```diff
@@ -38,310 +30,188 @@
...
-    // Revoke the mode transition authorization as well so the machine is no longer eligible.
-    RevokeLocked(SNTTemporaryMonitorModeLeaveReasonSyncServerChanged);
-    return 0;
-  }
-
-  NSNumber* deadline = tmm[kStateTempMonitorModeDeadlineKey];
-  if (!deadline) {
-    return 0;
-  }
+    : TimedSyncSession(kMinTemporaryMonitorModeMinutes, kMaxTemporaryMonitorModeMinutes,
```
mlw, line 44: This was an existing bug - but we're passing "minutes" here thru to the `Timer` constructor, but the `Timer` is expecting seconds. This only matters if someone were to set a TMM max time more than `kMaxTemporaryMonitorModeMinutes` (12 hours). But we should fix this up.

mlw (changes requested):

Source/santad/TemporaryMonitorMode.mm
sharvilshah: Fixed. Moved it to the base ctor so it multiplies by 60 before handing off to Timer. Did it there instead of in TMM so TAM gets it too

Source/santad/TimedSyncSession.mm
sharvilshah: Yeah, nice find! Went with the guard you suggested. The re-entrant fire is always coming from the modeTransition write, and that doesn't touch the chain, so old==new catches it before we get to Revoke(). Left a comment above it explaining why it's there, since the syncBaseURL watcher right above was basically doing the same thing by accident and we didn't carry it over here.

Source/santad/TimedSyncSession.mm
sharvilshah: fixed, thanks!

mlw (approved):

PR #1036, opened 2026-07-01
sync: Temporary Admin Mode policy and upload audit events

sharvilshah:
Part of WS-1128

Stacked on #1034 

Wires Temporary Admin Mode (TAM) into the sync flow:

- Preflight parses the TAM policy from the sync response into SNTSyncState and
  commits it via the config bundle, so the daemon picks up on-demand / revoke
  TAM policy the same way it already does the mode transition.
- Event upload maps the TAM audit events (enter / leave / denied), including
  their reason enums, to the santa.sync.v2 proto messages.


sharvilshah:
@coderabbitai full review

russellhancox (approved):

PR #1037, opened 2026-07-01
santad: add Temporary Admin Mode

sharvilshah:
Part of WS-1128, stacked on #1035

Adds Temporary Admin Mode (TAM): a console user can self-elevate to admin for a bounded, auto-reverting window, modeled on Temporary Monitor Mode and built on the shared TimedSyncSession base from the prior PR #1035.

- daemon: (TemporaryAdminMode, AdminGroupMembership): add/revert with auto-expiry, audit-event emission, and the unprivileged-control XPC surface. The elevation target is resolved from the XPC peer's audit token, never trusted from the GUI payload.

- gui: status-bar menu entry to request/observe TAM, with justification prompting driven by policy, plus the notifier-XPC methods the daemon calls to drive the UI.
 
- santactl: santactl status reports TAM availability and seconds remaining.

PR #1039, opened 2026-07-01
santad: revoke Temporary Admin Mode on console-session leave

sharvilshah:
Part of WS-1128, stacked on #1037

- Ends an active Temporary Admin Mode elevation when the console user's login session leaves, so a grant cannot outlive its session.
- SNTLoginWindowSessionHandler subscribes to Endpoint Security login-window session events; on a leave it resolves the departing uid and asks TemporaryAdminMode to end that user's session

### DIFF
--- a/Source/santad/SNTLoginWindowSessionHandler.mm
+++ b/Source/santad/SNTLoginWindowSessionHandler.mm
@@ -70,6 +70,11 @@
     if (!tam->SecondsRemaining().has_value()) {
       return;
     }
+    // Best-effort resolve the login username to a uid. Resolution is NOT required: EndForUserEvent
+    // matches on the uid OR the stored username, so a transient directory failure or a
+    // local-vs-directory uid collision still revokes via the username. A uid of 0 (unresolved, or
+    // the root/loginwindow pseudo-user) never matches a real target uid.
+    uid_t uid = 0;
     struct passwd pwd;
     struct passwd* result = NULL;
     long bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
@@ -86,12 +91,10 @@
            buf.size() < (1 << 20)) {
       buf.resize(buf.size() * 2);
     }
-    if (rc != 0 || result == NULL || pwd.pw_uid == 0) {
-      // Unresolved, or the loginwindow/root pseudo-user — never a TAM target. The timer remains
-      // the backstop if a real revoke was somehow missed.
-      return;
+    if (rc == 0 && result != NULL && pwd.pw_uid != 0) {
+      uid = pwd.pw_uid;
     }
-    tam->EndForUserEvent(pwd.pw_uid, reason);
+    tam->EndForUserEvent(uid, username, reason);
   });
 }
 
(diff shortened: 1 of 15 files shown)


sharvilshah:
@coderabbitai full review

Source/santad/SNTLoginWindowSessionHandler.mm
```diff
@@ -0,0 +1,75 @@
...
+  }
+  // Resolve the uid and run the revoke (RemoveMember can be a slow OpenDirectory write) off the
+  // ES handler thread. `username` is retained by the block; UTF8String is read on the queue.
+  auto tam = _temporaryAdminMode;
+  dispatch_async(_queue, ^{
+    struct passwd pwd;
+    struct passwd* result = NULL;
+    char buf[1024];
+    if (getpwnam_r(username.UTF8String, &pwd, buf, sizeof(buf), &result) != 0 || result == NULL ||
+        pwd.pw_uid == 0) {
```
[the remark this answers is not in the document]
sharvilshah: fixed

Source/santad/SNTLoginWindowSessionHandler.mm
```diff
@@ -0,0 +1,91 @@
...
+               ERANGE &&
+           buf.size() < (1 << 20)) {
+      buf.resize(buf.size() * 2);
+    }
+    if (rc != 0 || result == NULL || pwd.pw_uid == 0) {
+      // Unresolved, or the loginwindow/root pseudo-user — never a TAM target. The timer remains
+      // the backstop if a real revoke was somehow missed.
+      return;
+    }
+    tam->EndForUserEvent(pwd.pw_uid, reason);
```
mlw: We bail early here, but I wonder if this could be problematic. I'm not well versed on all the ways local vs. directory collision can happen, or UID lookup failures could happen... But the `TemporaryAdminMode` class stores both the uid and username. If `EndForUserEvent` took both the resolved UID (or 0 on failure) and username, it could match on either.

Source/santad/SNTLoginWindowSessionHandler.mm
```diff
@@ -0,0 +1,91 @@
...
+  }
+  // Guard against an empty/nil username (invalid UTF-8, or the loginwindow pseudo-user).
+  if (username.length == 0) {
+    return;
+  }
+  // Resolve the uid and run the revoke (RemoveMember can be a slow OpenDirectory write) off the
+  // ES handler thread. `username` is retained by the block; UTF8String is read on the queue.
+  auto tam = _temporaryAdminMode;
+  dispatch_async(_queue, ^{
+    struct passwd pwd;
```
mlw, line 78: We should short circuit doing work here when no TAM session is active. Checking `tam->SecondsRemaining()` takes a lock tho, so we should do the check in the async block here.

mlw (dismissed):

Source/santad/SNTLoginWindowSessionHandler.mm
sharvilshah: yeah, makes sense. Updated

Source/santad/SNTLoginWindowSessionHandler.mm
sharvilshah: yep! updated

mlw (approved):

PR #1043, opened 2026-07-04
common: add token-validated csops helpers (csops_audittoken)

mlw:
## What

Adds `audit_token_t`-taking overloads of the four `CSOpsHelper` getters
(`CSOpsStatusFlags`, `CSOpsGetCDHash`, `CSOpsGetTeamID`, `CSOpsGetSigningID`),
each defaulting to `csops_audittoken(2)`.

## Why

`csops(2)` reads by raw pid, so a pid recycled between the time an identity is
captured and the time it's resolved can be misattributed. `csops_audittoken`
has the kernel verify the token's pid+pidversion against the live process
atomically with each read (ESRCH on mismatch), closing that race.

The overloads bind the token into the existing `CSOpsFunc` injection seam via a
small `BindToken` adapter, so the pid-based implementations and blob parsing
remain the single implementation (no duplicated logic).

## Notes

- `Source/santad/KillingMachine.mm`: the getters are now overloaded, so passing
  a bare name into `make_unique<StringMatcher>(...)` became ambiguous
  (forwarding-reference deduction has no target type to resolve against). Cast
  to the pid-based overload's function-pointer type at the three call sites
  (behavior-preserving) to keep the build green.

## Testing

- New `CSOpsHelperTest` (3 tests), including a live-kernel check that a bumped
  pidversion yields ESRCH on all four reads.
- Broader sweep: `//Source/common:unit_tests` (62 targets) + `KillingMachineTest`
  + `process_tree_test` all pass.

## Context

First of a series: this is the prerequisite for santanetd resolving its
fallback process identity via `csops_audittoken` (separate PR).

### DIFF
--- a/Source/common/BUILD
+++ b/Source/common/BUILD
@@ -262,8 +262,18 @@ objc_library(
     name = "CSOpsHelper",
     srcs = ["CSOpsHelper.mm"],
     hdrs = ["CSOpsHelper.h"],
+    sdk_dylibs = ["bsm"],
     deps = [
         ":String",
+    ],
+)
+
+santa_unit_test(
+    name = "CSOpsHelperTest",
+    srcs = ["CSOpsHelperTest.mm"],
+    deps = [
+        ":AuditUtilities",
+        ":CSOpsHelper",
     ],
 )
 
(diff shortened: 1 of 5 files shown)


sharvilshah (approved):
lgtm

PR #1044, opened 2026-07-05
santactl: add adminmode command and report user type in status

sharvilshah:
Part of WS-1128

- Add `santactl adminmode` (alias `am`) to request or cancel a Temporary Admin Mode
  grant for the current user; drives the existing TAM XPC surface, no daemon changes
- Non-root command: the daemon resolves the target uid from the XPC peer audit token
  and handles the GUI authorization (and optional justification) prompt
- A request with no `--duration` resolves to the policy-configured default
- Lift the `--duration` parser from `monitormode` onto the shared `SNTCommand` base and
  add `SNTCommandTest` coverage

### DIFF
--- a/Source/santactl/Commands/SNTCommandMonitorMode.mm
+++ b/Source/santactl/Commands/SNTCommandMonitorMode.mm
@@ -53,47 +53,6 @@ REGISTER_COMMAND_NAME(@"monitormode")
   return [NSSet setWithArray:@[ @"mm" ]];
 }
 
-// Parse a time interval string into a number of minutes.
-// e.g. "10m" -> 10, "2h" -> 120, "3d" -> 3600
-- (NSTimeInterval)parseTimeInterval:(NSString*)duration {
-  NSScanner* scanner = [NSScanner scannerWithString:duration];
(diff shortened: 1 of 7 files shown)


Source/santactl/Commands/SNTCommandMonitorMode.mm
```diff
@@ -53,47 +53,6 @@ + (NSString*)longHelpText {
...
-    }
-    if ([unit isEqualToString:@"d"]) {
-      return intValue * (60 * 24);
-    }
-  }
-  return 0;
-}
-
 - (void)runWithArguments:(NSArray*)arguments {
   NSTimeInterval requestedDuration;
```
mlw: You set this to `0` in the new file, we should replicate here, feels like a latent bug.

Source/santactl/Commands/SNTCommandMonitorMode.mm
mlw: Similarly, let's initialize to `NO`

mlw (dismissed):

mlw (approved):

PR #1045, opened 2026-07-05
santad: demote and restore admins with Temporary Admin Mode

sharvilshah:
Part of WS-1128. Looks large, but ~65% are tests. 

- When the TAM policy turns on, santad snapshots direct group-80 admins
  (uid >= 500) into a `DemotedAdmins` record in the tamper-protected
  state.plist and makes them standard; on revoke it restores exactly those
  users and deletes the record.
- Record presence is the only edge detector, so reconciliation runs every
  sync and is idempotent: demote once per off-to-on edge, never re-demote,
  never fight other user-management tools.
- Persist-before-flip with rollback: no path to a demoted user with no
  record. Restore retries every sync until it converges.
- Also hardens the TAM grant path: availability re-check under the session
  lock, grants refused while a teardown residue is pending, and session
  state persisted before elevating.

### DIFF
--- a/Source/santad/AdminUserState.mm
+++ b/Source/santad/AdminUserState.mm
@@ -41,7 +41,7 @@ AdminUserState::AdminUserState(SNTConfigurator* configurator,
       revoke_tam_([revoke_tam copy]) {}
 
 void AdminUserState::HandlePolicy(SNTTemporaryAdminPolicy* policy) {
-  absl::MutexLock lock(&lock_);
+  absl::MutexLock lock(lock_);
   bool have_record = [configurator_ savedDemotedAdmins] != nil;
   // A record that survives a revoke (kept to retry a failed restore) also
   // suppresses capture for the whole next enabled window: restored users keep
@@ -104,7 +104,7 @@ void AdminUserState::SetupFromState() {
 }
 
 void AdminUserState::HandleSyncServerChange() {
-  absl::MutexLock lock(&lock_);
+  absl::MutexLock lock(lock_);
   if ([configurator_ savedDemotedAdmins] == nil) {
     return;
   }
--- a/Source/santad/AdminUserState.mm
+++ b/Source/santad/AdminUserState.mm
@@ -1,0 +1,195 @@
+/// Copyright 2026 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#include "Source/santad/AdminUserState.h"
+
+#import "Source/common/SNTError.h"
+#import "Source/common/SNTLogging.h"
+
+namespace santa {
+
+// Keys for one entry in the persisted DemotedAdmins record.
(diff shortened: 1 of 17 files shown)


Source/santad/AdminUserState.mm
```diff
@@ -0,0 +1,195 @@
...
+      LOGI(@"DemotedAdmins: demoted %@ (uid=%u) to standard", user[kDemotedAdminUsername], uid);
+    } else {
+      // Deliberately not retried: re-applying demotions would fight other
+      // user-management software. The user stays recorded and stays admin;
+      // remediation is toggling the policy off then on.
+      LOGE(@"DemotedAdmins: failed to demote %@ (uid=%u): %@", user[kDemotedAdminUsername], uid,
+           err.localizedDescription);
+    }
+  }
+}
```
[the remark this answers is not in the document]
sharvilshah: @coderabbitai Not fixing this, since this doesn't buy anything: `NewPolicyReceived` on the line above does the same directory-backed work synchronously, and `HandlePolicy` only touches the directory on policy transitions — steady-state syncs are a no-op. The only waiter is the sync service's own sequential thread, which intentionally wants commit-before-proceed. Making this async would also reintroduce the stranding race this PR closes: a queued revoke restore could run after a later sync re-enables the policy and clobber a live TAM session's state.

Source/santad/TemporaryAdminMode.mm
```diff
@@ -182,8 +181,25 @@
...
+bool TemporaryAdminMode::ApplyEffect(NSError** err) ABSL_EXCLUSIVE_LOCKS_REQUIRED(lock_) {
+  // Persist-before-flip: name the target on disk before the elevation, so a
+  // crash between the two leaves a deadline-0 record the next daemon start
+  // reverts — never an elevated user with no on-disk state. It also
+  // guarantees AdminUserState's capture exclusion sees the target: capture
+  // reads membership first and TAM state second, so member-visible implies
+  // state-visible. BeginSessionLocked overwrites this record immediately
+  // after a successful add.
+  PersistExpiredForRetryLocked();
+  if (!membership_->AddMember(target_uid_, err)) {
```
mlw: If `AddMember` fails, `PersistExpiredForRetryLocked` has cleared the grant. If this is a refresh (e.g., add more time), then if the daemon is restarted before the time runs out, the state management will be incorrect

mlw (dismissed):

Source/santad/TemporaryAdminMode.mm
sharvilshah: fixed

Source/santad/AdminUserState.mm
```diff
@@ -0,0 +1,265 @@
...
+
+AdminUserState::AdminUserState(SNTConfigurator* configurator,
+                               std::unique_ptr<AdminGroupMembership> membership,
+                               void (^revoke_tam)(void))
+    : configurator_(configurator),
+      membership_(std::move(membership)),
+      revoke_tam_([revoke_tam copy]) {}
+
+void AdminUserState::HandlePolicy(SNTTemporaryAdminPolicy* policy) {
+  absl::MutexLock lock(&lock_);
```
mlw: nit: This is using the older, deprecated form. Use the newer form that doesn't take the reference.

Source/santad/AdminUserState.mm
```diff
@@ -0,0 +1,265 @@
...
+
+  // The watchers only fire on changes: a sync server that went away while the
+  // daemon was not running is reconciled directly.
+  if (!configurator_.isSyncV2Enabled) {
+    HandleSyncServerChange();
+  }
+}
+
+void AdminUserState::HandleSyncServerChange() {
+  absl::MutexLock lock(&lock_);
```
mlw: nit: This is using the older, deprecated form. Use the newer form that doesn't take the reference.

mlw (changes requested):

mlw (approved):

PR #1047, opened 2026-07-05
Add remote_hostname field to NetworkActivity.Flow

mlw:
Adds an optional remote_hostname field (14) to NetworkActivity.Flow so santanetd can emit the remote hostname it already collects per flow. Field is placed with the other remote-endpoint fields; next ID: 15. Wire-compatible addition; no code changes.


russellhancox (approved):

pmarkowsky (approved):
LGTM.

PR #1066, opened 2026-07-14
common: reduce per-evaluation copies in CEL activations

mlw:
Two changes that reduce per-evaluation copying in the CEL activation path. When a fallback rule's result is non-cacheable (expressions touching `args`, `ancestors`, etc.) this cost is paid on every exec, so it scales with system exec rate.

1. `Memoizer<T>::operator()` returns `const T&` instead of copying the cached value on every access. Copy/move members are deleted so references can't outlive their storage; contracts documented and tested.
2. `Activation::FindValue` wraps the memoized `ancestors`/`fds` protos in place instead of `CopyFrom`-ing each onto the eval arena — the same lifetime contract `target` already uses. Ancestor data now materializes once per evaluation instead of three times.

**Reviewer note:** commit 2 is only memory-safe on top of commit 1 (the old by-value return would leave the wrapped pointers dangling mid-evaluation) — don't cherry-pick it alone. `MemoizerTest` pins the contract; new populated-`ancestors` CELTest coverage gives the nightly ASan run visibility into this path (previously empty fixtures only).

Tested: `common`, `cel`, and `santad` suites pass; the unsafe combination was verified to crash `testAncestors` under `MallocScribble`.

### DIFF
--- a/Source/common/MemoizerTest.mm
+++ b/Source/common/MemoizerTest.mm
@@ -1,0 +1,96 @@
+/// Copyright 2026 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import <XCTest/XCTest.h>
+
+#include <type_traits>
+
+#include "Source/common/Memoizer.h"
+
+namespace {
+
+// Counts copy constructions/assignments so tests can assert that accessing a
+// memoized value does not duplicate it.
+struct Counted {
+  static int copies;
+
+  int value = 0;
+
+  Counted() = default;
+  explicit Counted(int v) : value(v) {}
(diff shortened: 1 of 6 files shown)


Source/common/MemoizerTest.mm
```diff
@@ -0,0 +1,96 @@
...
+  santa::Memoizer<int> m([&calls] {
+    ++calls;
+    return 42;
+  });
+
+  XCTAssertFalse(m.HasValue());
+  XCTAssertEqual(m(), 42);
+  XCTAssertTrue(m.HasValue());
+  XCTAssertEqual(m(), 42);
+  XCTAssertEqual(calls, 1);
```
russellhancox, line 69: n.b there is a test for this in Test.mm already but it's in context and therefore not as clear as this. I think it's fine to have both but wanted to call it out

Source/common/cel/Test.mm
```diff
@@ -623,6 +623,104 @@ - (void)testFds {
   }
 }
 
+- (void)testAncestors {
```
russellhancox: We have tests that largely overlap this in SNTPolicyProcessorTest.mm. Not sure exactly how much duplication there is but we probably don't need both.

russellhancox (dismissed):

Source/common/MemoizerTest.mm
mlw: There is some overlap, but the one in Test.mm is focused on the Activation case. Changed the comment to make this clear.

Source/common/cel/Test.mm
mlw: Reworked things to reduce duplication.

russellhancox (approved):

PR #1077, opened 2026-07-16
Protect tamper-resistant paths from copyfile

mlw:
Tamper resistance doesn't cover `copyfile`, which can both copy protected data out and copy over/into protected paths.

**Fix:** Subscribe to `AUTH_COPYFILE` and pass it through the existing path-target validation. `Message::PathTargets` already extracts the source and destination (either the new path or the existing file to overwrite), so both directions are covered. Under inverted target-path muting ES delivers the event if any target path is protected.

Also simplifies the caching logic: rather than enumerating which events "don't support caching", only `OPEN` opts out (its all-or-nothing flags response can't be safely cached). Every other event's decision is a pure function of the target path(s) — which is what ES keys its cache on — and ES ignores the cache flag for events it can't cache. This flips the pre-existing CLONE/CREATE cases to cacheable; tests updated to match.


russellhancox (approved):

pmarkowsky (approved):
LGTM.

sharvilshah (approved):

PR #1082, opened 2026-07-16
TimedSyncSession: fix lock_ <-> timer-queue deadlock

sharvilshah:
Fixes SNT-468

- Holding lock_ while calling a blocking Timer method (StopTimer/StartTimerWithInterval) could deadlock: they dispatch_sync onto the timer queue, which may be running OnTimer(), which is waiting for lock_.
- Adds non-blocking StopTimerAsync/StartTimerWithIntervalAsync to the Timer<T> mixin (weak_ptr-guarded); session code enqueues them while lock_ is held so timer operations execute in lock order.
- active_/deadline_ are now the sole authority on session state; OnTimer validates against them, so a fire that loses a race to a cancel/refresh no-ops or re-arms instead of tearing down.

### DIFF
--- a/Source/common/SNTTimerTest.mm
+++ b/Source/common/SNTTimerTest.mm
@@ -15,8 +15,33 @@
 #import <Foundation/Foundation.h>
 #import <XCTest/XCTest.h>
 
+#include <memory>
+
 #import "Source/common/SNTTimer.h"
 #import "Source/common/TestUtils.h"
+#include "Source/common/Timer.h"
+
+namespace {
+
+// Minimal CRTP Timer subclass for exercising the async variants directly.
+// Must be owned by a shared_ptr: starting the timer takes shared_from_this().
+class AsyncTestTimer : public santa::Timer<AsyncTestTimer> {
+ public:
+  AsyncTestTimer()
+      : Timer(1, 60, OnStart::kWaitOneCycle, "AsyncTestTimer", RescheduleMode::kTrailingEdge,
+              QOS_CLASS_USER_INTERACTIVE) {}
+
+  bool OnTimer() {
+    if (on_timer_entered_) dispatch_semaphore_signal(on_timer_entered_);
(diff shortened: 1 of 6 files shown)


Source/common/SNTTimerTest.mm
```diff
@@ -15,8 +15,33 @@
...
+              QOS_CLASS_USER_INTERACTIVE) {}
+
+  bool OnTimer() {
+    if (on_timer_entered_) dispatch_semaphore_signal(on_timer_entered_);
+    if (on_timer_block_) dispatch_semaphore_wait(on_timer_block_, DISPATCH_TIME_FOREVER);
+    return false;
+  }
+
+  dispatch_semaphore_t on_timer_entered_;
+  dispatch_semaphore_t on_timer_block_;
```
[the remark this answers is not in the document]
sharvilshah: @coderabbitai SNTTimerTest.mm compiles with ARC. Under ARC, ObjC-pointer members of C++ classes (including dispatch_semaphore_t) are implicitly __strong and zero-initialized to nil by the default constructor, so OnTimer()'s existing nil checks are safe.

Source/santad/TimedSyncSession.h
```diff
@@ -207,24 +222,31 @@ class TimedSyncSession : public Timer<TimedSyncSession> {
...
-  // hide the base class Start/Stop methods
+  // Hide the base-class timer methods from subclasses and external callers.
+  // The blocking trio dispatch_syncs onto the timer queue: never call them
+  // while holding lock_ (see the class threading comment). Session code uses
+  // the *Async shims, always invoked with lock_ held.
   bool StartTimer();
   bool StopTimer();
   bool StartTimerWithInterval(uint32_t interval_seconds);
+  void StopTimerAsync();
+  void StartTimerWithIntervalAsync(uint32_t interval_seconds);
```
mlw, line 251: In the comment on the class above, you also call out `IsStarted` and `SetTimerInterval`, should those also be included here?

sharvilshah: Yep, updated

mlw (approved):

PR #1087, opened 2026-07-19
common: reduce per-exec copies and duplicate work in the process tree

mlw:
The exec ingest path deep-copied the `Program` (args vector + strings) two-plus times per delivery and rebuilt it even for deliveries the tree discards as duplicates. This moves the `Program` through `HandleExec` rather than copying it, moves the args vector out of the ES adapter, and hoists the arg count out of the loop.

It also skips constructing the new `Process`/`Program` entirely when a read-locked peek shows the exec was already applied. `StepLocked` remains the authoritative dedup gate; the peek is a best-effort fast path, behavior-preserving because a seen key has already advanced the removal cutoff past its timestamp. Adds a duplicate-delivery regression test; the existing tree tests still pass.

### DIFF
--- a/Source/common/processtree/process_tree.cc
+++ b/Source/common/processtree/process_tree.cc
@@ -122,26 +122,36 @@ void ProcessTree::HandleFork(uint64_t timestamp, const Process& parent,
 }
 
 void ProcessTree::HandleExec(uint64_t timestamp, const Process& p,
-                             const Pid new_pid, const Program prog,
-                             const Cred c) {
+                             const Pid new_pid, Program prog, const Cred c) {
   // TODO(nickmg): should struct pid be reworked and only pid_version be passed?
   assert(new_pid.pid == p.pid_.pid);
 
-  // Expected double-apply: a tree-aware auth client (the Authorizer) subscribes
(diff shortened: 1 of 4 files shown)


Source/common/processtree/process_tree.cc
```diff
@@ -122,26 +122,36 @@ void ProcessTree::HandleFork(uint64_t timestamp, const Process& parent,
...
+    absl::ReaderMutexLock lock(mtx_);
+    if (seen_.contains(key)) {
+      return;
+    }
+  }
 
-  // Construct the new process (copies program/args/signing) OUTSIDE the lock to
-  // keep the shared tree lock short on the serial ES handler path. A duplicate
-  // wastes this allocation, which is cheaper than lengthening the lock hold.
+  // Construct the new process (copies program/args/signing) OUTSIDE the write
```
russellhancox: "Copies"

russellhancox (dismissed):

Source/common/processtree/process_tree.cc
mlw: When fixing the comments, noticed that the check for "have we seen this exec before" was being done after copying things like args the first time. Bubbled that check up in addition to fixing the comments.

russellhancox (approved):

PR #1097, opened 2026-07-21
Add HTTP Santa commands, drained serially at the end of sync

tburgin:
Adds a new sync stage that drains Workshop's command queue after a successful postflight. Each `POST /commands/{machine_id}` exchange is strictly 1:1: the request posts the result of the previously executed command, the response carries at most one newly queued command, and the loop runs until the server is drained (capped at 50 per sync). Drain failures never fail the sync — unexecuted commands stay queued server-side and retry next sync. Nothing is persisted client-side.

Command execution is factored out of the NATS push client into a transport-agnostic `SNTSantaCommandHandler` (one file per command: `+Kill`, `+EventUpload`) that both NATS and the HTTP drain call. NATS wire behavior is unchanged; HMAC/timestamp/nonce verification stays NATS-only.

Details:
- Event uploads post an ack-only `DELIVERED` result before executing so the server shows in-flight state, then `COMPLETE` with the payload. Kill posts straight to `COMPLETE`.
- Each command logs `Running command <id>...` so a long-running command doesn't look like a hung sync.
- Bumps the protos pin for the serial `CommandsRequest`/`CommandsResponse` shapes; the new `inventory_scan` NATS variant is routed to `ERROR_UNKNOWN_REQUEST_TYPE` until implemented.
- New tests for the handler and the drain stage; existing NATS command tests pass unmodified.

### DIFF
--- a/Source/santasyncservice/SNTSantaCommandHandler+EventUpload.mm
+++ b/Source/santasyncservice/SNTSantaCommandHandler+EventUpload.mm
@@ -14,23 +14,12 @@
 
 #import "Source/santasyncservice/SNTSantaCommandHandler+EventUpload.h"
 
-#include <algorithm>
-
 #import "Source/common/SNTLogging.h"
 #include "Source/common/String.h"
 
 namespace pbv1 = ::santa::commands::v1;
 using santa::NSStringToUTF8String;
 using santa::StringToNSString;
-
-// Per-path budget when waiting for an event upload command to finish. Each path
-// is bounded by the delegate's bundle-service timeout (600s) plus upload time,
-// and paths are processed serially.
-static constexpr int64_t kEventUploadPerPathTimeoutSeconds = 660;
-
-// Absolute ceiling on how long a single event upload command is waited on,
-// regardless of how many paths it carries.
-static constexpr int64_t kEventUploadMaxWaitSeconds = 3600;
 
 @implementation SNTSantaCommandHandler (EventUpload)
 
@@ -103,14 +92,12 @@ static constexpr int64_t kEventUploadMaxWaitSeconds = 3600;
     return pbResponse;
   }
 
-  int64_t timeoutSeconds =
-      std::min(kEventUploadPerPathTimeoutSeconds * eventUploadRequest.paths_size(),
-               kEventUploadMaxWaitSeconds);
-  if (dispatch_semaphore_wait(
-          sema, dispatch_time(DISPATCH_TIME_NOW, timeoutSeconds * NSEC_PER_SEC)) != 0) {
-    pbResponse->set_error(::pbv1::EventUploadResponse::ERROR_INTERNAL);
-    if (errorMessage) *errorMessage = "timed out waiting for event upload to complete";
-  } else if (uploadError) {
+  // Each path is bounded by the delegate's per-path timeouts (bundle-service
+  // generation plus upload retries), so the aggregate completion is guaranteed
+  // to fire in bounded time. Wait for it and report the real outcome — there is
+  // no separate outer deadline for the serial per-path work to outlast.
+  dispatch_semaphore_wait(sema, DISPATCH_TIME_FOREVER);
+  if (uploadError) {
     pbResponse->set_error(::pbv1::EventUploadResponse::ERROR_INTERNAL);
     if (errorMessage) *errorMessage = NSStringToUTF8String(uploadError.localizedDescription);
   }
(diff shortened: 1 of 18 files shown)


Source/santasyncservice/SNTSantaCommandHandler+EventUpload.mm
```diff
@@ -0,0 +1,126 @@
...
+static constexpr int64_t kEventUploadPerPathTimeoutSeconds = 660;
+
+// Absolute ceiling on how long a single event upload command is waited on,
+// regardless of how many paths it carries.
+static constexpr int64_t kEventUploadMaxWaitSeconds = 3600;
+
+// Forward declaration of private interface to access private properties
+@interface SNTSantaCommandHandler ()
+@property(weak) id<SNTPushNotificationsSyncDelegate> syncDelegate;
+@end
```
russellhancox: If we're going to do this, we might as well make it public

russellhancox (dismissed):

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -759,13 +767,34 @@ - (SNTSyncStatusType)postflightWithSyncState:(SNTSyncState*)syncState {
...
   return SNTSyncStatusTypePostflightFailed;
 }
 
+// Drain any commands the server has queued for this host. Runs after a
+// successful postflight; a failure (or partial drain) must not affect the
+// sync result. Undrained commands stay queued server-side and are picked up
+// on the next sync.
+- (void)commandsWithSyncState:(SNTSyncState*)syncState {
+  // Queued commands are only supported in sync v2. Skip on v1.
+  if (!syncState.isSyncV2) {
```
mlw, line 798: If an org updates Santa before Workshop, will this result in a bunch of 404's?

Source/santasyncservice/SNTSyncManager.mm
```diff
@@ -759,13 +767,34 @@ - (SNTSyncStatusType)postflightWithSyncState:(SNTSyncState*)syncState {
...
+    return;
+  }
+
+  SLOGD(@"Commands starting");
+  SNTSyncCommands* p = [[SNTSyncCommands alloc] initWithState:syncState
+                                               commandHandler:self.commandHandler];
+  if ([p sync]) {
```
mlw: Running this synchronously might be problematic if there are a lot of commands. E.g., kill proc has a 90s timeout because it has to walk proc listing and grab info for each one, upload commands could potentially be problematic too(? tho not sure if that work might already be punted off the queue...).

This could interfere with timely syncs.

mlw (dismissed):

Source/santasyncservice/SNTSyncManager.mm
mlw: Update - the robot says:
```
For N ≥ ~6 paths the outer wait caps at 3600s (kEventUploadMaxWaitSeconds) but the inner work is serial at up to 600s/path, so it can exceed the cap. On outer timeout, handleEventUploadRequestAndWait returns COMPLETE + ERROR_INTERNAL "timed out" (+EventUpload.mm:109-112) while the upload is still running on eventUploadQueue. Consequences: the next command's upload queues behind the abandoned one, and an upload that later succeeds gets reported as timed-out → redelivered/re-run next sync.
```

Source/santasyncservice/SNTSyncManager.mm
tburgin: Good callout, made a change to ignore 404s for this path.

Source/santasyncservice/SNTSyncManager.mm
tburgin: Yeah, I was trying to keep it simple but blocking the sync isn't good. I have moved commands to a serial async queue and have removed the timeouts. This lets the current command timeouts drive.

Any new commands requests will pile up in the serial queue and will checkin and find an empty server side queue, perfect!

Source/santasyncservice/SNTSyncStage.mm
```diff
@@ -186,6 +187,11 @@ - (NSData*)dataFromRequest:(NSURLRequest*)request
     }
   }
 
+  // Report the final HTTP status (0 if no HTTP response was received) so callers
+  // can special-case specific codes, e.g. a 404 for an endpoint an older server
+  // does not implement.
+  if (statusCode) *statusCode = response.statusCode;
+
```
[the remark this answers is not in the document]
tburgin: This is okay, these are background logs.

russellhancox (approved):

PR #1101, opened 2026-07-24
gui: Add `event_detail_button_text` support

russellhancox:


Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -739,12 +745,13 @@ - (BOOL)addExecutionRules:(NSArray<SNTRule*>*)executionRules
...
+                             @"(identifier, state, type, custommsg, customurl, "
+                             @"event_detail_button_text, timestamp, "
                              @"comment, cel_expr, seatbelt_policy, rule_id) "
-                             @"VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?);",
+                             @"VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);",
                              rule.identifier, @(rule.state), @(rule.type), rule.customMsg,
-                             rule.customURL, @(rule.timestamp), rule.comment, rule.celExpr,
-                             rule.seatbeltPolicy, @(rule.ruleId)]) {
+                             rule.customURL, rule.eventDetailButtonText, @(rule.timestamp),
+                             rule.comment, rule.celExpr, rule.seatbeltPolicy, @(rule.ruleId)]) {
```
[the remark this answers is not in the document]
russellhancox: The rule hash deliberately only includes decision-making fields - it already ignores customMsg, customURL, etc. This new field is also presentation-only and so should not be included.

Source/common/SNTCachedDecision.h
```diff
@@ -63,6 +63,7 @@
 
 @property NSString* customMsg;
 @property NSString* customURL;
+@property NSString* eventDetailButtonText;
```
mlw, line 66: Need to add to `copyWithZone:`

Source/gui/SNTNetworkFlowMessageWindowView.swift
```diff
@@ -286,7 +286,7 @@ struct SNTNetworkFlowMessageWindowView: View {
 
       HStack {
         if shouldAddOpenButton() {
-          OpenEventButton(action: openButton)
+          OpenEventButton(customText: event?.eventDetailButtonText, action: openButton)
```
mlw, line 289: I don't believe the network rule button text gets capped - this gets ulled down in Rule Download as proto. but not parsed out / verified

Source/santasyncservice/SNTSyncPreflight.mm
```diff
@@ -585,9 +585,18 @@ void HandleV2Responses(const ::pbv2::PreflightResponse& resp, SNTSyncState* sync
...
       NSString* customMsg = !rule.custom_msg().empty() ? StringToNSString(rule.custom_msg()) : nil;
       NSString* customURL = !rule.custom_url().empty() ? StringToNSString(rule.custom_url()) : nil;
+      // The button label has to fit on a button, so cap it (kEventDetailButtonTextMaxLength).
+      NSString* eventDetailButtonText = !rule.event_detail_button_label().empty()
+                                            ? StringToNSString(rule.event_detail_button_label())
+                                            : nil;
+      if (eventDetailButtonText.length > kEventDetailButtonTextMaxLength) {
+        eventDetailButtonText =
+            [eventDetailButtonText substringToIndex:kEventDetailButtonTextMaxLength];
+      }
```
mlw: The robot spotted an issue here - Since NSString operates on UTF-16, the truncation could occur in the middle of a multi-unit char. This also affects the code in `SNTSyncRuleDownload`.

I liked the suggestion to add a new function to `Source/common/String.h`, something like:
```
  // Returns `str` capped at `maxLength` UTF-16 code units, cutting only on a
  // character boundary. -substringToIndex: slices on a raw UTF-16 index, so it
  // will happily split a surrogate pair (emoji), an emoji ZWJ sequence, or a base
  // character and its combining marks; the orphaned half becomes U+FFFD the next
  // time the string is encoded as UTF-8. When the cut point lands inside such a
  // sequence the whole sequence is dropped, so the result may be shorter than
  // `maxLength`. Strings already within the limit (including nil) are returned
  // unchanged.
  static inline NSString* TruncateNSStringToLength(NSString* str, NSUInteger maxLength) {
    if (str.length <= maxLength) {
      return str;
    }

    // `maxLength` is a valid index because the string is longer than it. If the
    // sequence covering that index starts earlier, cutting at `maxLength` would
    // split it, so cut at the start of the sequence instead.
    NSRange sequence = [str rangeOfComposedCharacterSequenceAtIndex:maxLength];
    return [str substringToIndex:MIN(sequence.location, maxLength)];
  }
```

mlw (changes requested):

Source/gui/SNTNetworkFlowMessageWindowView.swift
russellhancox: Fixed this in SNTNetworkExtensionQueue.mm. I'd prefer to do this during ingestion but as the network rules are stored as a whole blob it would mean deserializing, updating, and then reserializing, so I'm doing it at notification time instead.

Source/santasyncservice/SNTSyncPreflight.mm
russellhancox: Done

mlw (approved):

PR #1104, opened 2026-07-28
santasyncservice: run binary uploads from the HTTP command queue

tburgin:
Bumps the protos pin to pick up the additive `binary_upload`/`inventory_scan` oneof fields on `QueuedCommand` and `CommandResult`, and wires binary upload into the sync-pull command drain. Execution moves out of the NATS client into a shared `SNTSantaCommandHandler+BinaryUpload` (one file per command, like `+Kill` and `+EventUpload`) exposing an async variant — NATS publishes from the completion block, keeping its message queue unblocked — and a blocking variant for the drain, bounded at 15 minutes so a santad that dies without replying can't wedge the serial sync queue, with late replies writing to a `shared_ptr` owned by the block rather than the arena message.

Binary uploads ack `DELIVERED` before executing like event uploads do, so the server sees them in flight; the per-variant name mapping now lives in one place behind `+shouldPostDeliveredAckForCommand:`, and `inventory_scan` is routed to `HOST_STATUS_FAILED` until implemented, mirroring how NATS answers it with `ERROR_UNKNOWN_REQUEST_TYPE`. One NATS behavior change: a sync delegate present but disconnected from santad used to drop the request silently and never publish a reply, and now reports `DISPOSITION_INTERNAL_ERROR`.

`bazel test :unit_tests` passes 138/138, including new coverage for the request round-tripping to santad with the presigned POST intact, unparseable/absent replies, allow-list rejection never reaching santad, the ack predicate matrix, and a full HTTP drain asserting the `DELIVERED` → `COMPLETE` sequence. A real upload was not exercised end to end — that needs a live sync server plus santad, so the daemon-side path remains covered only by the existing `SNTBinaryUploadControllerTest`.

A part of WS-1271

### DIFF
--- a/Source/santasyncservice/BUILD
+++ b/Source/santasyncservice/BUILD
@@ -26,18 +26,21 @@ objc_library(
     name = "SNTSantaCommandHandler",
     srcs = [
         "SNTSantaCommandHandler.mm",
+        "SNTSantaCommandHandler+BinaryUpload.mm",
         "SNTSantaCommandHandler+EventUpload.mm",
         "SNTSantaCommandHandler+Kill.mm",
         "SNTSantaCommandHandler+PackageInventory.mm",
     ],
     hdrs = [
         "SNTSantaCommandHandler.h",
+        "SNTSantaCommandHandler+BinaryUpload.h",
         "SNTSantaCommandHandler+EventUpload.h",
         "SNTSantaCommandHandler+Kill.h",
         "SNTSantaCommandHandler+PackageInventory.h",
     ],
     deps = [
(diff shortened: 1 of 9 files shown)


sharvilshah (commented):
Can we add `binary_upload` and `package_inventory` to our docs here: https://github.com/northpolesec/santa/blob/9c33b2a68e9629c21ee31d86a5717492cf40144a/docs/src/lib/santaconfig.ts#L131-L141

sharvilshah (approved):

tburgin:
Will add the docs in a follow up

PR #1105, opened 2026-07-29
Revert "santactl: report build metadata from the embed label"

mlw:
Reverts #1102. The metadata reported there was added before the release build process was settled; that information is tracked in the build artifacts instead, so there is no need to carry it in the version string.

- `CommitHash.plist` again contains only `SNTCommitHash`.
- `santactl version` returns to `<version> (build N, commit <sha>)` in all cases.

Testing: rebuilt `//Source/common:CommitHash` both with and without an embed label carrying build metadata — only `SNTCommitHash` is written in either case. `bazel build //Source/santactl:SNTCommandVersion` is clean and `./Testing/lint.sh` exits 0.


Vpr99 (approved):

matt-bobrowski (approved):
LGTM.

russellhancox (approved):

PR #1106, opened 2026-07-29
workflows: Skip scheduled and deploy jobs outside this repo

mlw:
`continuous` and `sanitizers` run on a schedule, and `docs-deploy` publishes to GitHub Pages. In a fork these either consume runner time to no purpose or cannot succeed, so guard them on the repository name.

The docs `build` job is left unguarded on purpose, so documentation changes are still validated in a fork.

Verified the guard lands only on `continuous/preqs`, `sanitizers/test` and `docs-deploy/deploy`, and that all three files still parse.

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   preqs:
+    if: github.repository == 'northpolesec/santa'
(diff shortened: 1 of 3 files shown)


pmarkowsky (approved):
LGTM.

PR #1125, opened 2026-08-08
santactl: Make the duration parser generic

mlw:
`SNTCommand parseTimeInterval:` returned minutes because its only callers needed
minutes — the function was written inside `SNTCommandMonitorMode` and hoisted into
`SNTCommand` when `adminmode` became a second caller. It now takes the unitless
interpretation from the caller and returns checked `int64_t` seconds, with syntax
errors reported through `NSError`. The whole-minute requirement the two
temporary-mode commands share lives in `+[SNTCommand parseWholeMinutes:error:]`.

Parsing uses `strtoll` rather than `NSScanner`. `ERANGE` distinguishes a value too
large to represent from a genuine boundary value, which `NSScanner` cannot, so
there is no saturation guard and no unreachable switch arm.

Also fixes a duration-cap bypass found while doing this (first commit, separate so
it can be reviewed and backported on its own): both temporary-mode clamps tested
the limit with `-unsignedLongLongValue` and returned `-unsignedIntValue`, which
disagree for an `NSNumber` holding a double above 2^64 — a request could pass the
30-day check as a small value and then be granted as `UINT32_MAX` minutes.

Behavior changes to `--duration`:
- `120s` is accepted as 2 minutes. `s` was previously scanned as a unit character
  but had no branch, so it always failed.
- Invalid values say what was actually wrong instead of sharing one generic
  message. Leading whitespace, an unknown unit, and trailing content are distinct.
- Non-ASCII digits are rejected. `NSScanner` accepted them, so `١٠m` parsed as 10
  minutes; `strtoll` does not.

Help text is unchanged: `--duration {minutes}` and the `30m`/`2h` examples stay
accurate.


Source/santactl/Commands/SNTCommandAdminMode.mm
```diff
@@ -25,6 +29,31 @@
 @interface SNTCommandAdminMode : SNTCommand <SNTCommandProtocol>
 @end
 
+// Resolves a --duration argument to whole minutes; a bare integer means minutes,
+// matching the documented flag. SNTMonitorModeDurationMinutes is the deliberate
+// twin of this function, so changes here usually belong there too.
```
russellhancox: The two functions are identical apart from the name. Any reason not to have this be shared in SNTCommand (or a separate file, or part of parseTimeInterval:defaultUnit:error:)?

Source/santactl/SNTCommand.h
```diff
@@ -67,6 +70,18 @@
...
+///  What a duration string with no unit suffix means. `SNTDurationUnitNone`
+///  makes a bare integer invalid.
+///
+typedef NS_ENUM(NSInteger, SNTDurationUnit) {
+  SNTDurationUnitNone = 0,
+  SNTDurationUnitSeconds,
+  SNTDurationUnitMinutes,
+  SNTDurationUnitHours,
+  SNTDurationUnitDays,
+};
```
russellhancox, line 83: This seems like YAGNI. The two callers are using SNTDurationUnitMinutes, none of the others are used. The flag in both commands is documented as minutes, we can assume that is the case until we need this extra complexity

Source/santactl/SNTCommand.mm
```diff
@@ -44,45 +45,112 @@ - (void)printErrorUsageAndExit:(NSString*)error {
...
+      case 'h': unit = SNTDurationUnitHours; break;
+      case 'd': unit = SNTDurationUnitDays; break;
+      default:
+        // Unreachable: the charset scan above only ever admits 's', 'm', 'h', or 'd'.
+        [SNTError
+            populateError:error
+                 withCode:SNTErrorCodeInvalidDuration
+                   format:@"invalid duration \"%@\": unknown unit '%@' (expected s, m, h, or d)",
+                          duration, scannedUnit];
+        return std::nullopt;
```
russellhancox, line 106: Given that this is provably unreachable this seems unnecessary.

Source/santactl/SNTCommand.mm
```diff
@@ -44,45 +45,112 @@ - (void)printErrorUsageAndExit:(NSString*)error {
...
+                                defaultUnit:(SNTDurationUnit)defaultUnit
+                                      error:(NSError**)error {
+  if (duration.length == 0) {
+    [SNTError populateError:error
+                   withCode:SNTErrorCodeInvalidDuration
+                     format:@"invalid duration: empty string"];
+    return std::nullopt;
+  }
+
   NSScanner* scanner = [NSScanner scannerWithString:duration];
```
russellhancox: This is implementing a pretty large amount of `strtoll()` - one call would get you the integer (with overflow detection) and the trailing suffix.

mlw: Ditched the old `NSScanner`, moved to `strtoll`

Source/santactl/SNTCommand.h
mlw: This was added because I have an upcoming change that doesn't use bare values to mean "minutes"

Source/santactl/SNTCommand.mm
mlw: After moving to `strtoll` this is now needed.

Source/santactl/Commands/SNTCommandAdminMode.mm
mlw: Restructured things a bit, moved a helper into `SNTCommand`.

russellhancox (approved):

PR #1136, opened 2026-08-12
santad: Separate reusable code-signing validation results from authorization-result cacheability

dgollahon-plaid:
Context-dependent CEL decisions still require policy evaluation on every execution, but they no longer repeat static code-signing validation for an unchanged binary. The previous reuse condition effectively depended on `certSHA256`. This is incomplete because valid ad-hoc signatures—including linker-generated signatures—normally have no certificate. A non-cacheable CEL decision therefore retained the binary identity but repeated static signature analysis on every execution.

The cached validation state now distinguishes:

- Successful validation, including ad-hoc and linker-signed binaries without certificates
- Unsigned binaries
- Deterministic signature and executable-format failures
- Results that require validation because they are unknown or may depend on mutable system state

Each deterministic failure has a typed state that maps to its exact `OSStatus`. Mutable failures, such as filesystem, cancellation, trust, database, and internal errors, are retried.

The issue was observed with frequently executed command-line tools:

- Homebrew `mise`: valid ad-hoc, linker-signed binary with no signing certificate. Static validation succeeds, but the missing certificate previously caused repeated validation after non-cacheable policy decisions.
- GitHub `rg`: valid ad-hoc, linker-signed binary with no certificate.
- Homebrew `rg`: valid ad-hoc, non-linker-signed binary with no certificate.
- Unsigned or damaged `starship` variants returned `errSecCSUnsigned` and could also repeat the same analysis.

Rough local measurements for the installed `mise` binary:

- Repeated direct static validation: approximately 1.4–1.5 seconds per invocation in the instrumented sandbox.
- Reusing the cached validation result: below measurable millisecond resolution.
- End-to-end policy comparison under the same test environment:
  - Validation repeated: approximately 1.76 seconds median.
  - Validation reused: approximately 0.04 milliseconds median.

The sandbox substantially inflated absolute validation time. Interactive shell observations were a few hundred milliseconds per invocation. The important result is the repeated work and the relative difference, not the sandbox’s absolute latency.

Added a CEL regression test which verifies that a non-cacheable authorization retains completed signature validation while still re-evaluating policy for the next execution.

### DIFF
--- a/Source/common/SNTCachedDecision.h
+++ b/Source/common/SNTCachedDecision.h
@@ -21,6 +21,33 @@
 
 @class MOLCertificate;
 
+typedef NS_ENUM(NSInteger, SNTCodesignValidationState) {
+  SNTCodesignValidationStateNeedsValidation = 0,
+  SNTCodesignValidationStateSuccess,
+  SNTCodesignValidationStateUnsigned,
+  SNTCodesignValidationStateSignatureFailed,
+  SNTCodesignValidationStateSignatureUnsupported,
+  SNTCodesignValidationStateRequirementInvalid,
+  SNTCodesignValidationStateRequirementUnsupported,
+  SNTCodesignValidationStateBadObjectFormat,
+  SNTCodesignValidationStateSignatureInvalid,
+  SNTCodesignValidationStateTooBig,
+  SNTCodesignValidationStateUnsupportedDigestAlgorithm,
+  SNTCodesignValidationStateInvalidTeamIdentifier,
+  SNTCodesignValidationStateBadTeamIdentifier,
+  SNTCodesignValidationStateMultipleExecutableSegments,
+  SNTCodesignValidationStateInvalidEntitlements,
+  SNTCodesignValidationStateInvalidRuntimeVersion,
+};
+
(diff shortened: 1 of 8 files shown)


dgollahon-plaid:
Hi there 👋 

Feel free to treat this fix as a bug-report annotated with tests.

I (and other colleagues) have observed hundreds of milliseconds (or even seconds) of latency on very fast binary programs that I managed to trace back to Santa. This is especially notable for tools like `mise` or `starship` that are invoked on every shell prompt.

The root cause appears to be failing to cache signature validation. I've tried to produce an appropriate caching mechanism and regression tests with the help of Codex, though I am unfamiliar with both this codebase and this language and have no strong opinion on implementation beyond "I would like to not repeatedly pay this penalty on every binary invocation".

mlw:
Thank you for the report here. Without signing the CLA we cannot look at the diff. However, from the description I understand the issue and I think there is a simpler path than what is described here. I'm closing this PR and have opened #1139 to supersede it. If you believe this PR is not solving the same issue, please feel free to reach out!

dgollahon-plaid:
Thank you @mlw - I believe your approach will solve the problem as well.

PR #1138, opened 2026-08-12
santad: Don't present UI during signature invalidation

russellhancox:
On arm64 - where all binaries must be signed - a process with CS_KILL but not CS_VALID is going to be killed immediately. A process that has CS_KILLED set will also. These cases can occur when Santa is evaluating execution of a binary that was being written at the time of execution. Instead of blocking the execution and presenting UI and having Santa be blamed for an execution that would have failed anyway, suppress the UI in this case.

### DIFF
--- a/Source/common/CodeSigningIdentifierUtils.h
+++ b/Source/common/CodeSigningIdentifierUtils.h
@@ -37,7 +37,15 @@ static inline bool CdhashStrictlyEnforced(uint32_t csFlags) {
 // invalidity even if we allow the execution to proceed. Native
 // arm64 code must be signed, while unsigned x86_64 code may execute on Intel or
 // under Rosetta, so CS_KILL without CS_VALID is only terminal for arm64.
-bool KernelWillKillForInvalidSignature(uint32_t csFlags, cpu_type_t imageCPUType);
+//
+// Note that treating CS_KILL without CS_VALID on arm64 as invalid is technically
+// incorrect: to the kernel at least, such a binary is considered unsigned. This
+// is a deliberate trade-off. All arm64 binaries must be signed, so the kernel
+// would refuse to execute a genuinely unsigned one anyway, and creating an
+// unsigned arm64 binary is difficult in the first place. The far more common
+// case is a binary that is signed but damaged, so reporting invalid is the more
+// useful answer.
+bool KernelWillKillForCodeSigning(uint32_t csFlags, cpu_type_t imageCPUType);
 
 extern const NSUInteger kTeamIDLength;
 extern NSString* const kPlatformTeamID;
--- a/Source/common/CodeSigningIdentifierUtils.h
+++ b/Source/common/CodeSigningIdentifierUtils.h
@@ -17,6 +17,7 @@
 
 #import <Foundation/Foundation.h>
 #include <Kernel/kern/cs_blobs.h>
+#include <mach/machine.h>
 
 #include <cstdint>
 #include <string_view>
@@ -31,6 +32,12 @@ namespace santa {
 static inline bool CdhashStrictlyEnforced(uint32_t csFlags) {
   return (csFlags & CS_VALID) && (csFlags & (CS_HARD | CS_KILL));
 }
+
+// Returns true when the kernel will kill the process for code-signing
+// invalidity even if we allow the execution to proceed. Native
+// arm64 code must be signed, while unsigned x86_64 code may execute on Intel or
+// under Rosetta, so CS_KILL without CS_VALID is only terminal for arm64.
+bool KernelWillKillForInvalidSignature(uint32_t csFlags, cpu_type_t imageCPUType);
 
 extern const NSUInteger kTeamIDLength;
 extern NSString* const kPlatformTeamID;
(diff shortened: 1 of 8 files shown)


Source/santad/SNTPolicyProcessor.mm
```diff
@@ -781,7 +782,11 @@ - (nonnull SNTCachedDecision*)decisionForFileInfo:(nonnull SNTFileInfo*)fileInfo
       platformBinaryState:pbs
       signingStatusCallback:^SNTSigningStatus {
         uint32_t csFlags = targetProc->codesigning_flags;
-        if ((csFlags & CS_SIGNED) == 0) {
+        // CS_KILLED, and CS_KILL without CS_VALID on native arm64, identify
+        // images the kernel rejected even if CS_SIGNED has already been cleared.
+        if (santa::KernelWillKillForInvalidSignature(csFlags, imageCPUType)) {
+          return SNTSigningStatusInvalid;
+        } else if ((csFlags & CS_SIGNED) == 0) {
```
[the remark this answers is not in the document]
russellhancox: Critical system binaries cannot change once the system is up due to SIP or tamper protection.

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -781,7 +782,11 @@ - (nonnull SNTCachedDecision*)decisionForFileInfo:(nonnull SNTFileInfo*)fileInfo
       platformBinaryState:pbs
       signingStatusCallback:^SNTSigningStatus {
         uint32_t csFlags = targetProc->codesigning_flags;
-        if ((csFlags & CS_SIGNED) == 0) {
+        // CS_KILLED, and CS_KILL without CS_VALID on native arm64, identify
+        // images the kernel rejected even if CS_SIGNED has already been cleared.
+        if (santa::KernelWillKillForInvalidSignature(csFlags, imageCPUType)) {
+          return SNTSigningStatusInvalid;
```
mlw, line 808: nit: If you run an unsigned arm64 binary, this will report a status of `SNTSigningStatusInvalid` but we should report `SNTSigningStatusUnsigned` (obviously it won't run, but the reason would be inconsistent).

Source/common/CodeSigningIdentifierUtils.h
```diff
@@ -32,6 +33,12 @@ static inline bool CdhashStrictlyEnforced(uint32_t csFlags) {
   return (csFlags & CS_VALID) && (csFlags & (CS_HARD | CS_KILL));
 }
 
+// Returns true when the kernel will kill the process for code-signing
+// invalidity even if we allow the execution to proceed. Native
+// arm64 code must be signed, while unsigned x86_64 code may execute on Intel or
+// under Rosetta, so CS_KILL without CS_VALID is only terminal for arm64.
+bool KernelWillKillForInvalidSignature(uint32_t csFlags, cpu_type_t imageCPUType);
```
mlw: nit: I think we should consider a name change because "invalid" isn't the only reason. Maybe simply `KernelWillKillForCodeSigning`?

Source/santad/SNTPolicyProcessor.mm
mlw: I think we only want to test `CS_KILLED` here - otherwise we lose the fidelity of the other `SNTSigningStatus*` enum values. This would also mean we wouldn't need to thread thru the CPU type info.

Source/santad/SNTPolicyProcessor.mm
mlw: The review bot called out that this line is no longer correct with the change to the `singingStatusCallback` block. Assuming the block is updated to handle `CS_KILLED` - that would report "invalid", and here, that would carry thru, even though the binary was truly unsigned.

This is a pretty niche case - executing unsigned binary on arm64 isn't going to work regardless, but the status would be inconsistent.

Source/santad/SNTExecutionController.mm
```diff
@@ -449,6 +451,25 @@ - (void)validateExecEvent:(const Message&)esMsg
...
+    cd.silentBlockTTY = YES;
+    cd.holdAndAsk = NO;
+    NSString* extra = @"Kernel will kill the process for code signature invalidity; "
+                      @"suppressing block UI";
+    cd.decisionExtra =
+        cd.decisionExtra ? [NSString stringWithFormat:@"%@; %@", cd.decisionExtra, extra] : extra;
+    LOGW(@"Denying %@ but suppressing block UI: the kernel will kill this process for code "
+         @"signature invalidity (codesigning_flags=0x%x). The exec would have failed regardless of "
+         @"Santa's decision.",
+         @(targetProc->executable->path.data), targetProc->codesigning_flags);
```
mlw: nit: Use `santa::StringTokenToNSString(targetProc->executable->path)`

mlw (changes requested):

Source/common/CodeSigningIdentifierUtils.h
russellhancox: Done

Source/santad/SNTExecutionController.mm
russellhancox: Done

Source/santad/SNTPolicyProcessor.mm
russellhancox: See below

Source/santad/SNTPolicyProcessor.mm
russellhancox: As discussed - yes this is technically incorrect but is closer to matching user expectations. I've expanded the comment above KernelWillKillForCodeSigning to better explain the trade-off

mlw (dismissed):

mlw (approved):

PR #1139, opened 2026-08-12
santad: Reuse completed codesign validation across non-cacheable decisions

mlw:
`SNTPolicyProcessor` decided whether to run static code signature validation by checking for an empty `certSHA256`, which is always empty for ad-hoc and linker-signed binaries. Non-cacheable CEL decisions therefore repeated full validation on every execution of the same binary — measured at ~250ms per exec for a 54MB ad-hoc binary, ~13ms for a 6MB one, with no per-file memoization below us to soften it. That cost lands on the exec authorization path.

`SNTCachedDecision` now records the `OSStatus` validation returned; the guard keys on that, and failure effects are replayed from it so a reused failure behaves identically to a fresh one. Skipping validation on reuse loses nothing: every field `UpdateCachedDecisionSigningInfo` writes is already carried by `initWithCachedIdentity:`.

Only outcomes that cannot change while the identity is valid are recorded — a success, or an unsigned verdict that static validation and the kernel's `codesigning_flags` agree describes this executable. Any other failure is re-derived on the next execution, so a failure that may depend on state outside this vnode cannot pin the identity.

### DIFF
--- a/Source/common/SNTCachedDecision.h
+++ b/Source/common/SNTCachedDecision.h
@@ -59,6 +59,10 @@
 @property NSDate* secureSigningTime;
 @property NSDate* signingTime;
 
+/// The OSStatus returned by static code signature validation for this identity,
+/// or nil if validation has not run yet.
+@property NSNumber* codesignValidationStatus;
+
 @property NSString* quarantineURL;
 
 @property NSString* customMsg;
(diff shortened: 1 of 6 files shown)


pmarkowsky (dismissed):
LGTM. 

Only thought would be should we add a benchmark to prevent regression in any future refactoring? I'd do that in a future PR thought.

pmarkowsky (approved):
LGTM.

russellhancox (approved):

PR #1152, opened 2026-08-24
santad/gui: add sigterm-then-sigkill to KillingMachine and a pre-quit UI notification

sharvilshah:
Plumbing for timed rule kills, on top of #1149. Nothing calls this yet;
the component that uses it comes in the next PR, so this is inert.

- KillingMachine: KillingMachineTermThenKill() sends SIGTERM to matching
  process groups, waits a grace period, re-matches, and SIGKILLs
  survivors. KillingMachineAnyMatch() reports whether any process
  matches a request.
- Notifier XPC protocol gains postTimedRuleKillNotificationForApplication:deadline:.
  The GUI shows a "will quit at 5:00 PM" style banner for it, honoring
  silent mode. Strings added to all eight catalogs.
- Tests for both sides; no behavior change for anything existing.

Part of SNT-532

### DIFF
--- a/Source/gui/SNTNotificationManager.mm
+++ b/Source/gui/SNTNotificationManager.mm
@@ -408,6 +408,36 @@ static NSString* const silencedNotificationsKey = @"SilencedNotifications";
   [un addNotificationRequest:req withCompletionHandler:nil];
 }
 
+- (void)postTimedRuleKillNotificationForApplication:(NSString*)app deadline:(NSDate*)deadline {
+  if ([SNTConfigurator configurator].enableSilentMode) return;
+  if (!app || !deadline) return;
+
+  UNUserNotificationCenter* un = [UNUserNotificationCenter currentNotificationCenter];
+
+  NSDateFormatter* formatter = [[NSDateFormatter alloc] init];
+  formatter.timeStyle = NSDateFormatterShortStyle;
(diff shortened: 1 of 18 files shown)


Source/gui/SNTNotificationManager.mm
```diff
@@ -408,6 +408,36 @@ - (void)postRuleSyncNotificationForApplication:(NSString*)app {
   [un addNotificationRequest:req withCompletionHandler:nil];
 }
 
+- (void)postTimedRuleKillNotificationForApplication:(NSString*)app deadline:(NSDate*)deadline {
+  if ([SNTConfigurator configurator].enableSilentMode) return;
+  if (!app || !deadline) return;
+
+  UNUserNotificationCenter* un = [UNUserNotificationCenter currentNotificationCenter];
```
russellhancox: As discussed, probably best to move this to a real Santa dialog as UNUserNotificationCenter notifications are not configured in some environments (some customers have disabled them) and even in those that do they're easily missed (partially because of DnD). Given that these notifications are a warning before an action that could lose data, we should err on the side of caution

sharvilshah: Yep, implemented

russellhancox (approved):

PR #1153, opened 2026-08-24
santad: add SNTTimedRuleKills which the kill-at-window-end component

sharvilshah:
Stacked on #1152 

Note to reviewers: over half the diff is a single test suite, the new change is `SNTTimedRuleKills` plus small wiring (~660 lines)

New component that quits processes when
their policy_for_range window closes. Nothing records entries yet; that isthe next PR, so this still is inert. 

- SNTTimedRuleKills: one persisted entry per rule (state.plist, same
  per-key accessors as the temporary-mode keys), a wall-clock timer, a
  pre-quit banner at the warning time, and at the deadline a rule
  re-check followed by term-then-kill of matching process groups.
  BINARY and CERTIFICATE rules log and skip.
- SNTConfigurator gains the TimedRuleKills state key; SantadDeps wires
  the component up.
- Most of the diff is tests: 1172 of 1942 added lines, 28 tests covering
  persistence, restart, rule edits, and kill ordering.
  
Part of SNT-532

### DIFF
--- a/Source/santad/SNTTimedRuleKills.mm
+++ b/Source/santad/SNTTimedRuleKills.mm
@@ -1,0 +1,564 @@
+/// Copyright 2026 North Pole Security, Inc.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+
+#import "Source/santad/SNTTimedRuleKills.h"
+
+#import <CommonCrypto/CommonDigest.h>
+#import <Foundation/Foundation.h>
+#include <libproc.h>
+#include <signal.h>
+#include <sys/param.h>
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <memory>
+#include <optional>
+
+#import "Source/common/MOLXPCConnection.h"
+#import "Source/common/SNTConfigurator.h"
+#import "Source/common/SNTKillCommand.h"
+#import "Source/common/SNTLogging.h"
+#import "Source/common/SNTRule.h"
+#import "Source/common/SNTRuleIdentifiers.h"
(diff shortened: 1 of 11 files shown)


sharvilshah:
@coderabbitai full review

Source/santad/SNTTimedRuleKills.mm
```diff
@@ -0,0 +1,564 @@
...
+
+  SNTKillRequest* request = KillRequestForEntry(entry);
+  if (!request) {
+    LOGW(@"Unable to build a kill request for %@; nothing killed", entry.identifier);
+    return;
+  }
+
+  SNTKillResponse* response = self.killBlock
+                                  ? self.killBlock(request, kTermGrace)
+                                  : santa::KillingMachineTermThenKill(request, kTermGrace);
```
mlw: When this method is called via `processDueEntriesSerialized`, we're iterating over the processes that need to be terminated. The call to `KillingMachineTermThenKill` though will wait 5s for each due process. This means we will sleep for `kTermGrace * X` (where X is number of due processes). We should send `SIGTERM` to everything at once so that the grace is applied concurrently.

Source/santad/SNTTimedRuleKills.mm
```diff
@@ -0,0 +1,564 @@
...
+@property SNTNotificationQueue* notifierQueue;
+@property SNTRuleTable* ruleTable;
+@property SNTConfigurator* configurator;
+@property dispatch_queue_t queue;
+/// Entries keyed by rule type, identifier and CEL hash, so repeated execs under
+/// the same rule share one entry. Only ever touched on `queue`.
+@property NSMutableDictionary<NSString*, SNTTimedRuleKillEntry*>* entries;
+/// Test seam for the one call with no safe form to exercise against real
+/// processes. Never set in production, where the real kill runs instead.
+@property(copy) SNTKillResponse* (^killBlock)(SNTKillRequest* request, NSTimeInterval grace);
```
mlw: These two seams could be one `santa::KillEnv`, which #1149 already established for exactly this and which this test file already builds (`MakeEnv`/`FakeEnv`), routed into the real `KillingMachineTermThenKill` in `testSIGTERMThenSIGKILLWithFiveSecondGrace`.

Taking a `KillEnv` in the initializer would drop both properties from production and the two `self.killBlock ? ... :` branches, and let every kill/warning test exercise the production path with faked syscalls instead of replacing it.

mlw (dismissed):

Source/santad/SNTTimedRuleKills.mm
sharvilshah: Done

mlw (approved):

PR #1157, opened 2026-08-26
MOLCodesignChecker: Detect platform binaries without a platform identifier

tburgin:
Use the `anchor apple` requirement to identify platform binaries statically.


tburgin:
I can't confidently move forward with this as-is. This change would influence FormatSigningID which can lead to different enforcement decisions if we overreport platform status. A larger refactor would be needed to ensure we only ever use ES for platform binary status.

PR #1160, opened 2026-08-27
santad: Evaluate ProcessesWithOptions FAA rules

russellhancox:
- Apply a matched process's Action, bypassing the rule's RuleType and AuditOnly options, and use its option overrides for the resulting notification and TTY message
- Remember the overrides of the process matched at exec time for the process-centric rule types, so they apply to every event it produces
- Skip the process match when answering a read-only access for rules with no per-process AllowReadAccess override, preserving the previous ordering
- Convert the sync server's ProcessWithOptions rules into the config form

### DIFF
--- a/Source/santad/EventProviders/FAAPolicyProcessor.mm
+++ b/Source/santad/EventProviders/FAAPolicyProcessor.mm
@@ -16,6 +16,7 @@
 
 #include <bsm/libbsm.h>
 
+#include <optional>
 #include <vector>
 
 #include "Source/common/AccountLookup.h"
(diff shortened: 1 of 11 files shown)


Source/santad/EventProviders/FAAPolicyProcessor.mm
```diff
@@ -370,13 +393,37 @@ es_auth_result_t CombinePolicyResults(es_auth_result_t result1, es_auth_result_t
   // layer. If the policy would generally allow access to the resource,
   // producing the full kAllow result would potentially result in better
   // system performance.
-  if (PolicyAllowsReadsForTarget(msg, target, policy->options.allow_read_access)) {
-    return FileAccessPolicyDecision::kAllowedReadAccess;
+  if (PolicyAllowsReadsForTarget(msg, target, options.allow_read_access)) {
```
mlw: `PolicyAllowsReadsForTarget` was just checked about (line 380). We should try and remove the duplciate call.

mlw (changes requested):

mlw (approved):

PR #1161, opened 2026-08-27
santad: record and enforce timed rule kills from policy evaluation

sharvilshah:
Stacked on #1159 
Part of SNT-532

This is the PR that turns the time-based-rule feature on. 
Santa now records a pending quit when a rule asks for one, and carries it out.

- When a launch is allowed by a policy_for_range() rule with the kill
  flag set, santad saves a pending quit for that rule: the deadline,
  when to warn, and the window it came from. The component that fires
  these are in the two PRs below this one; this PR adds its caller.
- Setting the clock backwards no longer buys more time. santad keeps a
  monotonic reading alongside each deadline and a floor of believable
  time across restarts, so a pending quit fires on schedule even if the
  wall clock is rolled back, and windows cannot be reopened by rewinding
  the clock.
- KillingMachine can now handle several pending quits in one pass:
  one SIGTERM sweep, one shared grace period, one SIGKILL sweep, with
  each process group signaled once even when two rules cover it.

Note for reviewers: the bare today() overload ID is renamed to
today_timestamp for consistency (dispatch is by function name, so no
behavior change). And 2/3rd of this PR is tests.


sharvilshah:
@coderabbitai full review

sharvilshah:
@coderabbitai full review

Source/santad/SNTTimedRuleKills.mm
```diff
@@ -638,6 +740,7 @@ - (BOOL)rescheduleForOpenWindowSerialized:(SNTTimedRuleKillEntry*)entry now:(NSD
   entry.deadline = deadline;
   entry.notifyAt = [deadline dateByAddingTimeInterval:-lead];
```
mlw: The lead here is re-derived by subtraction, but RecordPendingKill() clamps the value being subtracted from:

```
.notify_at = std::max(now, window.window_end - lead),
```
so an exec allowed near the window’s close records `notifyAt == now`, and the effective lead becomes however long happened to be left rather than the formula’s `min(5m, len/10)`. The clamp itself is right — there’s no point scheduling a warning in the past — but line 775 then treats that artifact as the entry’s intended lead and carries it into every later occurrence.

Concretely: exec allowed at 16:59:30 under a 09:00–17:00 window, machine sleeps through 17:00, wakes inside tomorrow’s occurrence. The deadline correctly moves to tomorrow 17:00, but the warning is now 30 seconds out on a full 8-hour window — and stays 30 seconds for every reschedule after that.

mlw (dismissed):

Source/common/cel/Activation.h
```diff
@@ -95,11 +102,18 @@ class Activation : public ::google::api::expr::runtime::BaseActivation {
...
+  // calendar truth from the system clock.
+  std::function<absl::Time()> now_;
 
   // Set during evaluation when a relative-time function (today(), now(),
   // policy_for_range()) is used, which makes the result non-cacheable. Mutable
   // so it can be updated from the const evaluation path.
   mutable bool usedRelativeTime_ = false;
+  // Set during evaluation when policy_for_range() matched an open window with
+  // should_kill set. The earlier deadline wins if it is set more than once.
+  mutable std::optional<PendingKill> pendingKill_;
```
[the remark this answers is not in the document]
sharvilshah: @coderabbitai this is fixed in 6bff9dd560167bc05d21db31e3fa92a810b6c13d

Source/santad/SNTTimedRuleKills.mm
sharvilshah: Fixed in a64b42d4 (base branch PR #1159, merged in here). Reschedule now recomputes the lead via min(5m, len/10) instead of subtracting, formula lives in NotificationLead() and RecordPendingKill uses it too. The 16:59:30 scenario is now a unit test

mlw (dismissed):

mlw (approved):

PR #1163, opened 2026-08-27
santad: don't cache execution rule misses for file-hash-only lookups

tburgin:
Follow-up to #1162. The miss cache is keyed on the binary SHA-256. A caller could poison the cache with `santactl rule --check --sha256`.

Only use the cache on the exec path when code signing status is valid.

### DIFF
--- a/Source/santad/DataLayer/SNTRuleTable.mm
+++ b/Source/santad/DataLayer/SNTRuleTable.mm
@@ -36,6 +36,8 @@
 #include "Source/common/String.h"
 #include "Source/common/cel/Evaluator.h"
 
+#include <array>
+#include <cstring>
 #include <memory>
 #include <string>
 
@@ -46,9 +48,13 @@ static const int64_t kTransitiveRuleCullingThreshold = 500000;
 // Consider transitive rules out of date if they haven't been used in six months.
 static const NSUInteger kTransitiveRuleExpirationSeconds = 6 * 30 * 24 * 3600;
 
-// Maximum number of file hashes remembered by the execution rule miss cache. Measured at ~1.3MB
+// Maximum number of file hashes remembered by the execution rule miss cache. Measured at ~0.8MB
 // of live heap when full.
 static const uint64_t kExecutionRuleMissCacheSize = 10000;
+
+// Miss cache key: the binary SHA-256 as its 64 hex characters. Fixed size so the key is trivially
+// copyable into the db block and stored inline by the cache, leaving a lookup allocation-free.
+using ExecutionRuleMissCacheKey = std::array<char, 64>;
 
 static void addPathsFromDefaultMuteSet(NSMutableSet* criticalPaths) {
   // Create a temporary ES client in order to grab the default set of muted paths.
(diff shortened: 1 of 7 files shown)


mlw (dismissed):

pmarkowsky (dismissed):
Going forward with another strategy.

Source/santad/DataLayer/SNTRuleTableTest.mm
```diff
@@ -509,18 +509,19 @@ - (void)testExecutionRuleMissCacheIgnoresLookupsWithoutAFileHash {
...
+  SNTRule* r = [self.sut executionRuleForIdentifiers:ids useCache:YES];
   XCTAssertNotNil(r, @"a miss with no file hash must not be cached");
   XCTAssertEqual(r.type, SNTRuleTypeCDHash);
 }
 
-// A cached miss must only suppress lookups for the file it was recorded for. Note there is no
-// rule write between the two lookups here -- a write would clear the cache and hide the bug.
+// A cached miss must only suppress lookups for the file it was recorded for. Both lookups carry
+// the same Team ID, so only the file hash distinguishes them. Note there is no rule write between
+// them -- a write would clear the cache and hide the bug.
```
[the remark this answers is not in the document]
tburgin: A hash-only lookup can seed the cache

russellhancox (dismissed):

Source/santad/SNTPolicyProcessor.mm
```diff
@@ -703,7 +703,9 @@ static void ApplyCodesignValidationFailure(SNTCachedDecision* cd, OSStatus statu
     }
   }
 
-  SNTRule* rule = [self.ruleTable executionRuleForIdentifiers:CreateRuleIDs(cd)];
+  BOOL signatureVerdictIsStable = (cd.codesignValidationStatus != nil);
```
mlw: We might also want to check `cd.signingStatus != SNTSigningStatusInvalid` in case `cd.signingStatus` and `cd.codesignValidationStatus` disagree

Source/santad/DataLayer/SNTRuleTable.mm
```diff
@@ -603,8 +608,8 @@ - (SNTRule*)executionRuleForIdentifiers:(struct RuleIdentifiers)identifiers {
   // If we already looked in the db during another exec and there was no rule,
   // return early. The cache will be cleared when new rules are added. Keyed by
   // file sha256. Using a cdhash is appealing, but a cdhash is not guaranteed
-  // unique across files.
-  std::string missCacheKey = identifiers.binarySHA256.length
+  // unique across files. An empty key means this lookup must not be cached.
+  std::string missCacheKey = (useCache && identifiers.binarySHA256.length)
```
mlw: nit: if we add `__block` we can eliminate the copy into the block. I think we could also consider using `std::array` here instead of string to get rid of a couple of allocations, a bit larger change though.

mlw (changes requested):

Source/santad/DataLayer/SNTRuleTable.mm
tburgin: I moved to std::array on the stack.

Source/santad/SNTPolicyProcessor.mm
tburgin: Done

mlw (approved):

